Issue #22, opened 2014-12-09
We need support for 'Today 01:56 AM'


horva:
Currently it returns something like datetime.datetime(2014, 12, 9, 15, 17, 21, 562654) - date is correct here but time corresponds to import time because of last line in freshness_date_parser.py - we should consider changing this maybe by reinitializing freshness parser each time.

Allactaga:
Resolved in https://github.com/scrapinghub/dateparser/pull/26

Issue #12, opened 2014-11-27
Extend language redetection to every subparser used.


Allactaga:
For now language detection works only with subparser extending `dateutil` behavior. We should move it one level up to the main parser, so we can use same detected language for all approaches, including freshness subparser and formats. This way we can set default formats specific to some languages.

Allactaga:
Resolved in https://github.com/scrapinghub/dateparser/pull/26

Issue #9, opened 2014-11-26
Provide a way to check supported languages


eliasdorneles:
Suggested on #6 

The idea is to have something like ``dateparser.languages`` that would allow one to check support for a given language.

Right now, there are language specific code in both ``date_parser.py`` and ``freshness_date_parser.py``.
Any thoughts on how we should do that?

Allactaga:
Resolved in https://github.com/scrapinghub/dateparser/pull/26

PR #26, opened 2014-12-18
Feature descriptive languages

Allactaga:
So, the main idea is instead of detecting languages separately for different parsers and make parsers translate dates on their own, we detect and translate dates outside of parsers and make parsers operate strictly on English dates. Parsers code became very simple because of this and it would also make much more easier to add new parsers if we would ever need to.

Branch in its current state lacks majority of tests, but all functionality is in place. Also as you can see in `languages.yaml` file, adding new languages is very easy.

This request targets issues #9 and #12.


asadurski:
Hello @Allactaga,
Oops, sorry for interfering then. Since we needed both Thai and Vietnamese in some sites, I decided to get involved. I hope that's more of help than trouble.
Changes merged in #32. I added failing Vietnamese tests too.
I think we don't have a way to check for weekdays, right? So temporarily I added weekday test entries for Vietnamese in freshness date parser tests but they would need to be updated manually depending on time passed from last weekday.

Allactaga:
Hi, @asadurski, thanks for getting involved. More help is always better!
I fixed tests for the Vietnamese (btw weekdays are processed by dateutil parser, not by freshness. Think about it as freshness only parse forms like "... ago" and you can not say "Thursday ago").
Although there are couple of cases i did not fix and i think it is best to go without such support:
* "19 Jan 2010 lúc 2:19pm", this date is mixing two languages, so it is not a valid date to consider. You can simply fix it with `re.sub(ur"\blúc\b", "", date_string")` in spider and it would be valid english dates/
* "Thg 11 25", while in this particular case it is obvious that it is November, it is not quite clear in the string like "1 năm 1 tháng 1 tuần". It is most likely "1 year 1 month 1 day" but could also be "1 year 1 january day". I think it is better for now to substitute such dates with "Tháng mười một" in spider.

data/languages.yaml
```diff
@@ -0,0 +1,1514 @@
...
+        - hoje: 0 dias
+        - um: 1
+        - uma: 1
+        - alguns segundos: 44 segundos
+
+
+ru:
+    name: Russian
+
+    skip: ["около", "в", "и", "г."]
```
Allactaga: Why is "г." in skip?

data/languages.yaml
```diff
@@ -0,0 +1,1514 @@
...
+        - dziś: 0 dnia
+        - dzisiaj: 0 dnia
+        - onegdaj: 2 dnia
+        - przedwczoraj: 2 dnia
+        - wczoraj: 1 dnia
+
+fa:
+    name: Persian
+
+    skip: ['\u064e', '\u064f']
```
Allactaga: Please avoid unicode escaping here. We want native speakers easily to read through the document

data/languages.yaml
waqasshabbir: I noticed some dates like 'Сентября 2015 г. в 10:23'. And г. won't translate to anything. Hence, I put it in the skipped list.

Allactaga: That is short for year in Russian. :)

If you added more words like that, could you point them out please, so we can ask native speakers about them.

waqasshabbir: If we'd move г. to 'hour' array it would translate the above string like 'september 2015 year 10:23' and hence freshness_parser will parse it for '2015 year' returning an incorrect date.

Allactaga: This date is not processed by the current code in master, right?
Let's create an issue for this date string and go for faster merge.

waqasshabbir: And at the same time, do you want me to move г. to 'hour' array in yaml? I'll create an issue right now.

Allactaga: Yes please. But it is 'year', not 'hour'.

data/languages.yaml
waqasshabbir: These are diacritics that I added proactively. I was reading a guide on how dates are written in Persian and above characters were added to some month names. Generally, on forums/websites, these diacritics were not found with dates. I'll just remove them for now and we can add them later if needed.

Allactaga: I don't mean you need to remove them. I mean do not use escaping, use ` َ` and ` ُ`

dateparser/date.py
```diff
@@ -87,55 +90,145 @@ def get_last_day_of_month(year, month):
...
+        return instance._parse()
+
+    def _parse(self):
+        for parser in (
+            self._try_timestamp,
+            self._try_freshness_parser,
+            self._try_given_formats,
+            self._try_dateutil_parser,
+            self._try_hardcoded_formats,
+        ):
```
eliasdorneles: This is really lovely, folks!
Awesome to have this clear in the code the ordering of the attempts.
+sys.maxint =)

dateparser/date.py
```diff
@@ -87,55 +90,145 @@ def get_last_day_of_month(year, month):
...
+            date_obj = parser()
+            if self._is_valid_date_obj(date_obj):
+                return date_obj
+        else:
+            return None
```
eliasdorneles: A little nitpicking: I'd remove the else and just return None in the top level (maybe it's just me, I find Python's for/else construct a bit weird).

dateparser/date.py
```diff
@@ -87,55 +90,145 @@ def get_last_day_of_month(year, month):
...
+        if self._translated_date is None:
+            self._translated_date = self.language.translate(self.date_string, keep_formatting=False)
+        return self._translated_date
+
+    def _get_translated_date_with_formatting(self):
+        if self._translated_date_with_formatting is None:
+            self._translated_date_with_formatting = self.language.translate(self.date_string, keep_formatting=True)
+        return self._translated_date_with_formatting
+
+    def _is_valid_date_obj(self, date_obj):
```
eliasdorneles: Sorry, a bit more bikeshedding... :D
This would look better to me as:

    return all(isinstance(date_obj, dict),
               date_obj.get('date_obj'),
               date_obj.get('period') in ('day', 'week', 'month', 'year'))

dateparser/freshness_date_parser.py
```diff
@@ -5,235 +5,51 @@
...
+                r'about|ago|\d+',
+                r':|[ap]m'] 
+
+        date_string = re.sub(r'\s+', ' ', date_string.strip())
+
+        words = filter(lambda x: x if x else False, re.split('\W', date_string))
+        words = filter(lambda x: not re.match(r'%s' % '|'.join(skip), x), words)
+        return not bool(words)
+
+    def _parse_time(self, date_string):
```
eliasdorneles: Could we have a docstring here explaining why this is needed?
If I understood correctly, this is for dates like "yesterday 2pm", with yesterday replaced by "2 days", right?

Something like:

    """Attempts to parse the time part for date strings like '1 day ago at 2pm'"""

dateparser/languages/validation.py
```diff
@@ -0,0 +1,310 @@
+# -*- coding: utf-8 -*-
+import re
+from logging import getLogger
+
+
+class LanguageValidator(object):
```
eliasdorneles: This class feels to me overly-complicated and with lots of duplication.
Do we really need all this?

This is like re-parsing all the file again and checking types and values.
If we really want to go this route, maybe a XML file with a schema would be a better format -- although personally, I don't see the point.

I get that we want to show an informative message if there is something wrong in the YAML file, but  maintaining this will be a real burden.

I propose keeping the validation code minimal and use assertions in the code to prevent wrong types and values.

Makes sense?

eliasdorneles:
Hey folks -- I've done a bit of reviewing, and put some comments above (mostly some nitpicking).
Anyway, I just wanted to say that this is truly awesome work -- thoroughly tested and well refactored!

Sometimes I'd find some lines of code that were confusing, so I tried commenting them out and running the tests -- and, as expected, some of them failed (proving a good coverage) and then I could debug and figure out the purpose of that code.

Well done, folks, I'm blown away! :wave: :wave: 

I look forward to the next release! =)

dateparser/date.py
Allactaga: Let me explain why i do like for-else clause. It is explicitly stating "Hey, we expected something to happen in the loop, but as it did not happen, here is our plan 'B'". While it does the same job, it is a great way to organize code, because without the else, there are no expectations set.

dateparser/date.py
Allactaga: A agree, your code is more pythonic. Thanks!
Although it would require adding `sorted(date_obj.keys()) == ['date_obj', 'period']` to do the same job.

dateparser/languages/validation.py
Allactaga: I agree that validation is large and complicated but let me disagree with some other points.

1. Sometimes duplication is necessary. I once work on a project where we had same permission set for posts and comments. So one guy from our team once decided "Hey, it is duplication, lets use one function for both", he refactor the code and half a year later when all code points to new function we had to add an additional permission just for posts and not for comments. Guess what happened :grinning: . Same thing is here: while most of the validation is for lists, it all could change at some point. We can require different validation for weekdays for example.

2. It is actually not re-parsing the file, but validate already extracted data. File is loaded and transformed to yaml by `LanguageDataLoader`

3. Initial purpose of this change was to provide easy readable way to define languages. XML is not that. While it is great for machine-communication, it is rather hard to read for a human. Also i am not sure if it is possible to define facet that would check that groups in replacement are properly correlated to the matching regex.

4. Informative message is just a bonus here, main purpose is to keep all language information consistent. I mean if changes in structure of languages are made it is better to make a proper changes in validator to be sure that all languages received this change and nothing was left out. But validator is actually the one part of code that changes least frequently.

5. Assertions in the code instead of aside validation is really bad idea: parsing code is complicated enough by itself, and validation code does not needed for work. It is best to have validation in separate place.

Someday we can refactor this class to tests suite, but it should be done much-much later on the roadmap. For now we should be considered very young project that does not know what waits around the corner. And to be prepared for the rapid changes we sometimes need messy code. At least until we support the majority of languages around the world and know more about how they treat dates and do not make unnecessary assumptions (like word tokens does not contain spaces or only consists of word-characters) and then there would be less surprises on the way and we can relax a bit :smiley: .

I hope i have managed to explain now why this approach was chosen.

Allactaga:
Hey Elias, thanks for the input. I know it is quite a large PR and really appreciated your comments. :smiley:

dateparser/date.py
eliasdorneles: I'm not sure if it needs checking if it _only_ has these two keys -- if it answers to the protocol is good enough to me (a la duck typing).
If it's really needed, I'd go with comparing sets.

dateparser/languages/validation.py
eliasdorneles: I know that it's not actually re-parsing the file, it's just quite like it in the sense that it also reflects the whole structure.

Thanks for the detailed explanation, I understand that the main purpose of this is to increase confidence in consistence configuration across all the languages.

I still think there's a lot of room for improvement here, but I rather have this PR merged already, since this is such a minor concern. :)

dateparser/date.py
Allactaga: Protocol kinda required all valid responses to have exact same format or indicate 'not parsable' with `None`. Users would expect that all objects have same set of fields, don't you agree?

eliasdorneles: I agree that is the expected, it just seemed to me it wouldn't need enforcing. :)

Allactaga: I just hope someday to see a class instance here instead of the dict. :)

dateparser/freshness_date_parser.py
waqasshabbir: @eliasdorneles included the docstring.

waqasshabbir:
Fixes #22 and #38

PR #40, opened 2015-01-12
Miscellaneous fixes.

waqasshabbir:
This fixes dates like:
* Today at 10:56 AM issue #22.
* Correct parsing for dates previously failing tests.
* Correct parsing for dates where wrong time was picked up.


Allactaga:
Base was changed to the https://github.com/scrapinghub/dateparser/pull/26
View https://github.com/scrapinghub/dateparser/pull/41 instead.

PR #41, opened 2015-01-12
Miscellaneous fixes

Allactaga:
This is https://github.com/scrapinghub/dateparser/pull/40 with changed base.

This fixes dates like:
* Today at 10:56 AM issue #22.
* Correct parsing for dates previously failing tests.
* Correct parsing for dates where wrong time was picked up.

### DIFF
--- a/data/languages.yaml
+++ b/data/languages.yaml
@@ -108,6 +108,7 @@ en:
         - (\d+)\s*sec(s?): \1 second\2
         - noon: '12:00'
         - midnight: '00:00'
+        - (\d+)h(\d+)m?: \1:\2
 
 
 de:
--- a/dateparser/freshness_date_parser.py
+++ b/dateparser/freshness_date_parser.py
@@ -13,7 +13,22 @@ class FreshnessDateDataParser(object):
     def __init__(self, now=None):
         self.now = now or datetime.utcnow()
 
+    def _are_all_words_units(self, date_string):
+        skip = [
+            r'year|month|week|day|hour|minute|second',  # units
+            r'about|ago|\d+',  # others
+        ]
+
+        date_string = re.sub(r'\s+', ' ', date_string.strip())
+
+        words = filter(lambda x: x if x else False, re.split('\W', date_string))
+        words = filter(lambda x: not re.match('|'.join(skip), x), words)
+        return not bool(words)
+
     def parse(self, date_string):
+        if not self._are_all_words_units(date_string):
+            return None, None
+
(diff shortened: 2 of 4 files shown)


data/languages.yaml
```diff
@@ -108,6 +108,7 @@ en:
         - (\d+)\s*sec(s?): \1 second\2
         - noon: '12:00'
         - midnight: '00:00'
+        - (\d+)h(\d+)m?: \1:\2
```
Allactaga: Is it still needed, considering freshness would only parse appliable dates?

dateparser/freshness_date_parser.py
```diff
@@ -13,7 +16,40 @@ class FreshnessDateDataParser(object):
...
+                r':|[ap]m'] 
+
+        date_string = re.sub(r'\s+', ' ', date_string.strip())
+
+        words = filter(lambda x: x if x else False, re.split('\W', date_string))
+        words = filter(lambda x: not re.match(r'%s' % '|'.join(skip), x), words)
+        return not bool(words)
+
+    def _parse_time(self, date_string):
+        date_string = re.sub(r'\d+\s*(%s)' % _UNITS, '', date_string)
```
Allactaga: May be we should use regex and flags identical to the one in `get_kwargs`? How about creating a compiled one and use in both places?

dateparser/freshness_date_parser.py
```diff
@@ -13,7 +16,40 @@ class FreshnessDateDataParser(object):
...
+        date_string = re.sub(r'[\.,]', '', date_string)
+        try:
+            return parse(date_string).time()
+        except:
+            pass
+
     def parse(self, date_string):
+        time = self._parse_time(date_string)
```
Allactaga: Better to be put inside `if date:` body, so extra work can be avoided if there is no date.

data/languages.yaml
waqasshabbir: It is. Because translated string is passed onto dateutil parser, where time component (20h30m) is changed to `20 hour 30 minute` and is not supported by it.

dateparser/freshness_date_parser.py
waqasshabbir: I supposed are_all_words_units would exit the date parser if no parseable information is found. Sure let me still put it inside `if date:`.

dateparser/freshness_date_parser.py
waqasshabbir: Made and pushed the changes accordingly.

PR #43, opened 2015-01-22
adding regex for h between time

eLRuLL:
I could also add a `date_cleaner` when getting this so it can replace that `h`, but I think this is a common date, and could appear on various cases.

### DIFF
--- a/data/languages.yaml
+++ b/data/languages.yaml
@@ -292,6 +292,7 @@ fr:
         - aujourd'hui: 0 jours
         - un: 1
         - une: 1
+        - (\d+)h(\d+)m?: \1:\2
 
 
 es:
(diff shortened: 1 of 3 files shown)


Allactaga:
We already have a solution for this [here](https://github.com/scrapinghub/dateparser/blob/master/data/languages.yaml#L111)
Could you please provide a date string, where it is not working for the existing code?

waqasshabbir:
@Allactaga it's a french date. Example: `Jeudi 22 janvier 2015 à 18h00`. The solution doesn't seem to apply to it. The translated string looks like `thursday 22 january 2015  18 h 00`.

Allactaga:
It can be done in simplification for the language then. I mean not every language have h/m standing for hour/minute.
Also would be nice to bring '21h34m15s' into the picture

PR #47, opened 2015-01-29
Feature limit subset of languages

waqasshabbir:
Refactored ExactLanguage to ExactLanguages to accept a list of languages.
Added relevant tests.

### DIFF
--- a/dateparser/date.py
+++ b/dateparser/date.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import calendar
+import collections
 import re
 from datetime import datetime, timedelta
 
@@ -8,7 +9,7 @@ from dateutil.relativedelta import relativedelta
 from dateparser.date_parser import date_parser
 from dateparser.freshness_date_parser import freshness_date_parser
 from dateparser.languages import LanguageDataLoader
-from dateparser.languages.detection import AutoDetectLanguage, ExactLanguage
+from dateparser.languages.detection import AutoDetectLanguage, ExactLanguages
 
 
 def sanitize_spaces(html_string):
@@ -71,18 +72,18 @@ def get_intersecting_periods(low, high, period='day'):
 
 def sanitize_date(date_string):
     date_string = re.sub(
-        u'\t|\n|\r|\u00bb|,\s\u0432|\u0433\.|\u200e|\xb7|\u200f|\u064e|\u064f',
+        r'\t|\n|\r|\u00bb|,\s\u0432|\u0433\.|\u200e|\xb7|\u200f|\u064e|\u064f',
         ' ', date_string, flags=re.M
     )
     date_string = sanitize_spaces(date_string)
     date_string = re.sub(r'\b([ap])(\.)?m(\.)?\b', r'\1m', date_string, flags=re.DOTALL | re.I)
-    date_string = re.sub('^.*?on:\s+(.*)', r'\1', date_string)
+    date_string = re.sub(r'^.*?on:\s+(.*)', r'\1', date_string)
 
     return date_string
 
 
 def get_date_from_timestamp(date_string):
-    if re.search('^\d{10}', date_string):
+    if re.search(r'^\d{10}', date_string):
         return datetime.fromtimestamp(int(date_string[:10]))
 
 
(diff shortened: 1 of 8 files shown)


tests/test_date.py
```diff
@@ -284,6 +284,16 @@ def test_should_parse_date_with_timezones_using_format(self):
...
 
+    @parameterized.expand([
+        param(['ur', 'li']),
+        param(['pk',]),
+    ])
+    def test_should_raise_error_when_unknown_language_given(self, shortnames):
+        with self.assertRaises(ValueError) as context:
+            date.DateDataParser(languages=shortnames)
+
+        self.assertTrue(context.exception)
```
Allactaga: Why do we need this assert? Isn't `assertRaises` enough?

tests/test_languages.py
```diff
@@ -324,13 +324,35 @@ def then_no_language_was_detected(self):
...
+
+    def given_known_languages(self, shortnames):
+        self.known_languages = [self.language_loader.get_language(shortname) for shortname in shortnames]
+
     def given_detector(self):
-        self.assertIsInstance(self.known_languages, list, "Require a language to initialize")
-        self.assertEqual(1, len(self.known_languages), "Could be initialized with exactly one language")
-        self.detector = ExactLanguage(language=self.known_languages[0])
+        self.assertIsInstance(self.known_languages, list, "Require a list of languages to initialize")
+        self.assertGreaterEqual(len(self.known_languages), 1, "Could be initialized with one or more languages")
```
Allactaga: Should be "Could *only* be initialized", otherwise doesn't looks like error description.

dateparser/date.py
```diff
@@ -214,19 +217,21 @@ def _is_valid_date_obj(self, date_obj):
 
 class DateDataParser(object):
 
-    def __init__(self, language=None, allow_redetect_language=False):
-        if isinstance(language, basestring):
+    def __init__(self, languages=None, allow_redetect_language=False):
+        if isinstance(languages, collections.Iterable):
```
Allactaga: What if it is not `Iterable`?

waqasshabbir: It should check for `list` instance instead. I missed the fact that even strings are `Iterable`. Fixed it and also added else clause to raise TypeError in case invalid type is supplied.

tests/test_date.py
waqasshabbir: Thank you for pointing this out. Fixed it and modified `assertRaise` to `assertRaiseRegexp` to be better.

tests/test_languages.py
waqasshabbir: Fixed.

dateparser/date.py
Allactaga: What about `isinstance(languages, (list, tuple, collections.Set))` ?

dateparser/date.py
```diff
@@ -214,19 +216,23 @@ def _is_valid_date_obj(self, date_obj):
...
-            if language in available_language_map:
-                language = available_language_map[language]
+
+            if all([language in available_language_map for language in languages]):
+                languages = [available_language_map[language] for language in languages]
             else:
-                raise ValueError("Unknown language %r" % language)
+                unsupported_languages = set(languages) - set(available_language_map.keys())
+                raise ValueError("Unknown language(s) %r" % ', '.join(unsupported_languages))
+        elif languages:
```
Allactaga: What if we pass `languages=""` ?

waqasshabbir: bool("") evaluate to False, it seems to work just fine.

Allactaga: So you say it is a valid input ?

waqasshabbir: Oh, now I got your point. What do you suggest?

Allactaga: So, if we check for valid input in first if clause, and here we want to check if anything else was passed (and we have None as a default valid value), then i would say `elif languages is not None` would be appropriate.

waqasshabbir: Thank you @Allactaga.

dateparser/date.py
waqasshabbir: Just adopted this change and the one advised below. Please review.

Issue #36, opened 2015-01-06
Missing date parts


Allactaga:
Sometimes there are dates that don't have all the information to get the exact date. Like "December 21" or "Friday". We can assume either current year and week, or the one that is latest in the past.

To achieve that, instead of calling `parse` method in `dateutil_parse` function we would need to call for the `_parse` method of dateutil parser and then, when we have information on what parts are parsed (and depending on configuration), we either chose date that is current week/month/year or the last one seen.

Allactaga:
Require https://github.com/scrapinghub/dateparser/issues/35

Allactaga:
This should also affect `period` of the date, so "December 2014" return `period: month` and not `period: day`

Issue #35, opened 2015-01-06
Configuration


Allactaga:
It looks that we need to parametrize parsing behavior. Instead of keep adding parameters to the parse function I suggest to create a Registry changeable with `configure(key=value)` function. 
Example settings could be `NO_DATES_FROM_FUTURE` (to parse dates from web, where some times pieces from the created date is missing and we are assuming past) or `SUPPORT_BEFORE_COMMON_ERA` (to use custom datetime class inherited from datetime but supporting Astronomical year numbering)

PR #15, opened 2014-11-30
Porting code to be compatible with py3 and py2

eliasdorneles:
Hey, folks!

I ported the code to work also for Python 3 (specifically, 3.3 and 3.4).

Summary of the changes:

* using absolute imports instead of relative
* adjusted tox.ini and .travis.yml to test also with py3
* adapted new_relativedelta overriding ``__init__`` instead of ``__new__``
* using dict.items and dict.values instead of dict.iteritems and dict.itervalues
* using ``next(some_iter)`` instead of ``some_iter.next()``
* protected code dependant on specific versions (ideally, this would go away)

I'm also commenting below some stuff I'm not sure if it's handled the best way.

@csalazar @Allactaga Any concerns?

PR #50, opened 2015-02-03
Implements #35 and #36

markbaas:


dateparser/date_parser.py
```diff
@@ -32,10 +34,65 @@ def __add__(self, other):
...
 
+class new_parser(parser.parser):
+    """
+    Implements an alternate parse method which supports preference to dates in future and past.
+    For more see issue #36
+    """
+
+    def parse(self, timestr, default=None, ignoretz=False, prefer_future=None, **kwargs):
+        # timestr needs to be a buffer as required by _parse
+        timestr = timestr if not isinstance(timestr, str) else StringIO(timestr)
```
Allactaga: This line is not present in original `parse` method from `dateutil`, why do we need it here?

dateparser/date_parser.py
```diff
@@ -32,10 +34,65 @@ def __add__(self, other):
...
+
+        # Set right prefer_future var
+        prefer_future = prefer_future if prefer_future is not None \
```
Allactaga: We expect to have many settings in the future. Let's not use method argument (or it will very long list some day) and only use `settings`.

dateparser/conf.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+
+
+class Settings(object):
+    PREFER_DATES_FROM_FUTURE = None
```
Allactaga: It is very confusing when option have both `None` and `False` as valid options. It is best to use here two different `True/False` options and add some validation, so they both could not be True.

dateparser/date_parser.py
```diff
@@ -32,10 +34,65 @@ def __add__(self, other):
...
+            new_date = new_date.replace(hour=0, minute=0, second=0, microsecond=0)
+
+        return new_date
+
+
 def dateutil_parse(date_string, **kwargs):
     """Wrapper function around dateutil.parser.parse
     """
-    today = datetime.utcnow().replace(hour=0, minute=0, second=0, microsecond=0)
+    today = datetime.utcnow()
```
Allactaga: Why?

dateparser/date_parser.py
```diff
@@ -32,10 +34,65 @@ def __add__(self, other):
...
+        for field in ['year', 'month', 'day', 'hour', 'minute', 'second', 'microsecond']:
+            value = getattr(res, field)
+            if value is not None:
+                new_date = new_date.replace(**{field: value})
+                given_fields.append(field + 's')
+        if res.weekday and not res.day:
+            new_date = new_date + new_relativedelta(weekday=res.weekday)
+
+        # Correct date to if prefer in future of past
+        if prefer_future is not None:
```
Allactaga: Lets extract this to another method.

dateparser/date_parser.py
```diff
@@ -32,10 +34,65 @@ def __add__(self, other):
...
+        res = self._parse(timestr, **kwargs)
+
+        if res is None:
+            raise ValueError("unknown string format")
+
+        # Fill in missing date
+        given_fields = []  # fields set by
+        new_date = default
+
+        for field in ['year', 'month', 'day', 'hour', 'minute', 'second', 'microsecond']:
```
Allactaga: Why not using `repl = {}` following block from original method?

tests/test_date_parser.py
```diff
@@ -215,6 +219,44 @@ def test_dates_not_parsed(self, date_string):
         self.when_date_is_parsed()
         self.then_date_was_not_parsed()
 
+    @parameterized.expand([
```
Allactaga: Behavior without any preference should also be tested.

dateparser/date_parser.py
markbaas: The original parse has done something so that the string gets accepted by the _parse function. This line makes sure that timestr is always StringIO

dateparser/date_parser.py
markbaas: OK

dateparser/conf.py
markbaas: Perhaps it'd be better to change the setting into: PREFER_DATES_FROM
And give it three options: DEFAULT, PAST, FUTURE.

dateparser/date_parser.py
markbaas: OK

dateparser/date_parser.py
markbaas: In this case I save myself an unnecessary variable.

tests/test_date_parser.py
markbaas: Behavior without any preference is tested by all the other tests ;)

dateparser/date_parser.py
markbaas: This is essential. This code sets the clock to midnite. However, what if default is 16:00 and I want the future 15:00, it would result the same day, however it should be the day before.

dateparser/date_parser.py
Allactaga: Can you please point out to this [in the code](https://github.com/dateutil/dateutil/blob/master/dateutil/parser.py#L291) ?

dateparser/conf.py
Allactaga: Ok. But let's make this "current_period", "past" and "future" strings (not objects) to make it easier to read from the file later.

dateparser/date_parser.py
Allactaga: Very good point. Thank you.

dateparser/date_parser.py
Allactaga: What about `given_fields` you've added instead?

dateparser/date_parser.py
```diff
@@ -32,10 +34,65 @@ def __add__(self, other):
...
+        if prefer_future is not None:
+            for field in ['microseconds', 'seconds', 'minutes', 'hours', 'days',
+                          'weeks', 'months', 'years']:
+                # Can't override given field
+                if field in given_fields:
+                    continue
+
+                delta = relativedelta(**{field: 1})
+
+                if (prefer_future is False and new_date >= default and new_date - delta < default) or\
```
Allactaga: Lets also split this into two short methods with behavior specific to the past and future. It would be much more clear to read through.

tests/test_date_parser.py
Allactaga: Tests should be explicit. We should have test with such configuration set and test for this behavior exactly. In general tests should only test one thing each.

dateparser/conf.py
markbaas: OK

dateparser/date_parser.py
markbaas: The given_fields is used for a different purpose. While, populating data it saves the fields there which should not be overridden. If you think the initial implementation with the repl = {} variable is more efficient, I'm also fine with that.

dateparser/date_parser.py
markbaas: https://github.com/dateutil/dateutil/blob/master/dateutil/parser.py#L908

I changed my StringIO line into that exact line.

dateparser/date_parser.py
```diff
@@ -32,10 +33,82 @@ def __add__(self, other):
...
+        given_fields = []  # fields that cannot be overriden
+        new_date = default
+
+        # Populate all fields
+        repl = {}
+        for field in ['year', 'month', 'day', 'hour', 'minute', 'second', 'microsecond']:
+            value = getattr(res, field)
+            if value is not None:
+                repl[field] = value
+                given_fields.append(field + 's')
```
Allactaga: Instead of defining `given_fields` you can use keys of `repl` to iterate over and later `relativedelta(**{field + 's': 1})`

dateparser/date_parser.py
```diff
@@ -32,10 +33,82 @@ def __add__(self, other):
...
+        if settings.PREFER_DATES_FROM is not 'current_period':
+            for field in ['microseconds', 'seconds', 'minutes', 'hours', 'days',
+                          'weeks', 'months', 'years']:
+                # Can't override a given field
+                if field in given_fields:
+                    continue
+
+                # Try if applying the delta for this field corrects the problem
+                delta = relativedelta(**{field: 1})
+                if (settings.PREFER_DATES_FROM == 'past'
```
Allactaga: I believe this way it would be more easy to read through:
```python
@staticmethod
def _correct_for_past(date, delta, default):
    if settings.PREFER_DATES_FROM != 'past':
        return date

    if date - delta < default <= delta:
        date = date - delta
    return date

@staticmethod
def _correct_for_future(date, delta, default):
    if settings.PREFER_DATES_FROM != 'future':
        return date

    if date <= default < date + delta:
        date = date + delta
    return date
```

dateparser/date_parser.py
```diff
@@ -2,14 +2,15 @@
 from __future__ import unicode_literals
 
 import re
-
 from datetime import datetime
+from six import binary_type
```
Allactaga: Can we please use `binary_type = bytes if sys.version_info[0] == 3 else str  # PY3 compatibility` here to not introduce new dependency? We would address compatibility issue in https://github.com/scrapinghub/dateparser/pull/15 some time later.

dateparser/date_parser.py
```diff
@@ -32,10 +33,103 @@ def __add__(self, other):
...
+
+        # Clean hour and minutes, etc in case not defined
+        if not res.hour and not res.minute and not res.second and not res.microsecond:
+            new_date = new_date.replace(hour=0, minute=0, second=0, microsecond=0)
+
+        return new_date
+
+    @staticmethod
+    def _populate(res, default):
+        given_fields = []  # fields that cannot be overriden
```
Allactaga: You don't need this definition now

dateparser/date_parser.py
```diff
@@ -32,10 +33,103 @@ def __add__(self, other):
...
+            value = getattr(res, field)
+            if value is not None:
+                repl[field] = value
+        new_date = new_date.replace(**repl)
+
+        # Fix weekday
+        if res.weekday and not res.day:
+            new_date = new_date + new_relativedelta(weekday=res.weekday)
+
+        return new_date, [key + 's' for key in repl.keys()]
```
Allactaga: Just use `return self._correct(new_date, repl)` here instead of returning a tuple.

dateparser/date_parser.py
```diff
@@ -32,10 +33,103 @@ def __add__(self, other):
...
+        if date < default < date + delta:
+            date += delta
+            return date
+
+        return None
+
+    @staticmethod
+    def _correct_for_past(date, delta, default):
+        if settings.PREFER_DATES_FROM != 'past':
+            return None
```
Allactaga: Returning unchanged date as i suggested here instead of `None` would go better with method name `correct_*` and would allow not to check the result of this function in base `_correct` method but simply use several `date = correc_*(date)`

tests/test_date_parser.py
```diff
@@ -245,6 +306,9 @@ def wrapped(date_string):
         self.add_patch(patch('dateparser.date.date_parser', new=self.date_parser))
         self.parser = DateDataParser()
 
+    def given_configuration(self, key, value):
+        settings.update(key, value)
```
Allactaga: It is better to patch settings object here. Then we would not rely on `reload_settings` later. Tests have tendency to be copy-pasted and with patch we would have more guaranteed way of rolling back.

dateparser/date_parser.py
```diff
@@ -32,10 +33,103 @@ def __add__(self, other):
...
+
+        for field in ['microseconds', 'seconds', 'minutes', 'hours', 'days',
+                      'weeks', 'months', 'years']:
+            # Can't override a given field
+            if field in given_fields:
+                continue
+
+            # Try if applying the delta for this field corrects the problem
+            delta = relativedelta(**{field: 1})
+            new_date = new_parser._correct_for_future(date, delta, default) or new_parser._correct_for_past(date, delta, default)
```
Allactaga: As we calling other class methods from this one, it is better to use `@classmethod` for `_correct()` and use `cls.` here instead of class name.

dateparser/date_parser.py
markbaas: Sure, however, dateutil will still be depending on that library either way.

Allactaga: I see, thank you. We had long ago quick discussion on the topic [here](https://github.com/scrapinghub/dateparser/pull/15#discussion-diff-21075948), but as python3 support is still pending, let's for now use line that i've suggested.

tests/test_date_parser.py
```diff
@@ -14,6 +14,7 @@
 from dateparser.date_parser import DateParser
 from dateparser.languages import LanguageDataLoader
 from dateparser.languages.detection import AutoDetectLanguage, ExactLanguage
+from dateparser.conf import settings, reload_settings
```
Allactaga: Unused import `reload_settings`

Allactaga:
Also, can you please merge `master` into this branch?

PR #51, opened 2015-02-03
Fix logging

horva:
Removed calling logging.basicConfig on import level, added get_logger function which should be used for logging inside whole dateparse module

### DIFF
--- a/dateparser/languages/validation.py
+++ b/dateparser/languages/validation.py
@@ -70,9 +70,8 @@ class LanguageValidator(object):
         value = info['no_word_spacing']
         if value not in [True, False]:
             cls.get_logger().error(
-                "Invalid 'no_word_spacing' value %(value)r for '%(id)s' "
-                "language: expected boolean",
-                {'value': value, 'id': language_id})
+                "Invalid 'no_word_spacing' value %(value)r for '%(id)s' language: "
+                "expected boolean", {'value': value, 'id': language_id})
             result = False
 
         return result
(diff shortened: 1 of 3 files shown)


dateparser/utils.py
```diff
@@ -26,3 +27,18 @@ def increase_regex_replacements_group_positions(replacement, increment):
...
+
+
+def get_logger():
+    logger = logging.getLogger('dateparser')
+
+    # if there already is a handler for this namespace, no need to set one
+    if len(logger.handlers):
+        return logger
+
+    formatter = logging.Formatter('%(asctime)s %(levelname)s: [%(name)s] %(message)s')
```
Allactaga: Can you use `logging.config.dictConfig` here?

dateparser/languages/validation.py
```diff
@@ -1,10 +1,11 @@
 # -*- coding: utf-8 -*-
 import re
-from logging import getLogger
+from dateparser.utils import get_logger
 
 
 class LanguageValidator(object):
-    logger = getLogger('dateparser')
+    logger = get_logger()
```
Allactaga: This still would be configured on import. Please move to the `__init__` method.

dateparser/languages/validation.py
```diff
@@ -63,9 +69,10 @@ def _validate_word_spacing(cls, language_id, info):
 
         value = info['no_word_spacing']
         if value not in [True, False]:
-            cls.logger.error("Invalid 'no_word_spacing' value %(value)r for '%(id)s' language:"
-                             " expected boolean",
-                             {'value': value, 'id': language_id})
+            cls.get_logger().error(
+                "Invalid 'no_word_spacing' value %(value)r for '%(id)s' "
+                "language: expected boolean",
```
Allactaga: Why did you move 'language' to another line splitting the phrase?

horva: You mean, why is it not like this? 

    cls.get_logger().error(
        "Invalid 'no_word_spacing' value %(value)r for '%(id)s' language: "
        "expected boolean",

I did it kind of automatically because there were lot of entries. Picked language word to go to next row, for no particular reason.

Allactaga: I know, i am nitpicking here, but don't you agree that older variant is easier to read in the code, because there is complete phrase on each line?

horva: Yeah, I agree it is better. Updated PR.

PR #57, opened 2015-02-17
Fixes issue#49

waqasshabbir:


tests/test_date_parser.py
```diff
@@ -276,6 +276,56 @@ def test_dates_without_preference(self, date_string, expected):
         self.then_period_is('day')
         self.then_date_obj_exactly_is(expected)
 
+    @parameterized.expand([
+        param('February 2015', datetime(2015, 1, 31), datetime(2015, 2, 28)),
+        param('February 2012', datetime(2015, 1, 31), datetime(2012, 2, 29)),
+        param('March 2015', datetime(2015, 1, 25), datetime(2015, 3, 25)),
+        param('April 2015', datetime(2015, 1, 31), datetime(2015, 4, 30)),
+        param('April 2015', datetime(2015, 2, 28), datetime(2015, 4, 28)),
+        param('December 2014', datetime(2015, 2, 15), datetime(2014, 12, 15)),
```
Allactaga: What do you say if we use keyword arguments for readability: `param('December 2014', today=datetime(2015, 2, 15), expected=datetime(2014, 12, 15))` ?

tests/test_date_parser.py
```diff
@@ -276,6 +276,56 @@ def test_dates_without_preference(self, date_string, expected):
...
+    ])
+    def test_dates_with_day_missing_having_default_preference(self, date_string, utcnow, expected):
+        self.given_utcnow(utcnow)
```
Allactaga: Let's add explicit `current` configuration here.

tests/test_date_parser.py
```diff
@@ -276,6 +276,56 @@ def test_dates_without_preference(self, date_string, expected):
...
+    ])
+    def test_dates_with_day_missing_prefering_first_day_of_month(self, date_string, utcnow, expected):
+        self.given_configuration('PREFER_DAY_OF_MONTH', 'first')
+        self.given_utcnow(utcnow)
+        self.given_parser()
+        self.given_date_string(date_string)
+        self.when_date_is_parsed()
+        self.then_date_was_parsed_by_date_parser()
+        self.then_date_obj_exactly_is(expected)
+
```
Allactaga: Let's add another test to check that only missing dates are substituted:
```@parameterized.expand([
    param('16 February 2015', prefer_day_of_month='first', expected datetime(2015, 2, 16)),
    ...
```

dateparser/date_parser.py
```diff
@@ -82,6 +88,25 @@ def _populate(cls, res, default):
...
         return cls._correct(new_date, [key + 's' for key in repl.keys()], default)
 
+    @staticmethod
+    def get_valid_day(res, new_date):
+        _, tail = calendar.monthrange(res.get('year', new_date.year),
+                                      res.get('month', new_date.month))
+
+        if 'day' in res:
+            if res['day'] > tail:
+                raise ValueError('Day not in range for month',)
```
Allactaga: There is no test for this behavior.

dateparser/date_parser.py
```diff
@@ -73,7 +75,11 @@ def _populate(cls, res, default):
             value = getattr(res, field)
             if value is not None:
                 repl[field] = value
-        new_date = new_date.replace(**repl)
+
+        # Fix day and populate date with result
+        repl_copy = repl.copy()
```
Allactaga: Why do we use a copy here? Don't we want later code use corrected `repl` ?

waqasshabbir: Corrected `repl` interferes later on with correcting dates for past/future. Hence, we are using copy here.

dateparser/date_parser.py
waqasshabbir: Added one now.

tests/test_date_parser.py
waqasshabbir: Fixed.

tests/test_date_parser.py
```diff
@@ -276,6 +276,81 @@ def test_dates_without_preference(self, date_string, expected):
...
+    @parameterized.expand([
+        param('14 February 2015', prefer_day_of_month='current', expected=datetime(2015, 2, 14)),
+        param('1 January 2015', prefer_day_of_month='last', expected=datetime(2015, 1, 1)),
+        param('30 April 2015', prefer_day_of_month='first', expected=datetime(2015, 4, 30)),
+    ])
```
Allactaga: Hm, it looks like i was hasty to suggest three attributes for the test. I think we can get same result only iterating via `prefer_day_of_month`, and fixing f.ex. `date_string='14 February 2015', today='22 February 2015'` inside the test.

waqasshabbir: makes sense. Just made some changes. Please a take look again.

waqasshabbir:
Renamed the test and removed day assertion changes. Please take a look now.

PR #58, opened 2015-02-18
Unicode \xa0 in utf-8 is represented as \xc2\xa0. Hence, the fix.

waqasshabbir:


dateparser/date.py
```diff
@@ -14,7 +14,7 @@
 
 
 def sanitize_spaces(html_string):
-    html_string = re.sub(u'\xa0', ' ', html_string)
+    html_string = re.sub(u'\xc2\xa0', ' ', html_string)
```
Allactaga: Should there still be `u''` prefix in this case? And anyway, wouldn't this just work with `re.UNICODE` flag instead of encoding?

tests/test_date.py
```diff
@@ -284,6 +284,12 @@ def test_should_parse_date_with_timezones_using_format(self):
         self.assertEqual('day', date_data['period'])
         self.assertEqual(expected, date_data['date_obj'])
 
+    def test_should_parse_with_no_break_space_in_dates(self):
+        date_string = "08-08-2014 18:29"
```
Allactaga: It is not clear on reading that it is no-break space. Let's use explicit representation here.

dateparser/date.py
waqasshabbir: Agreed, `u''` prefix won't be needed. Yep, using `re.UNICODE` makes it work just fine.

tests/test_date.py
waqasshabbir: Sure.

PR #60, opened 2015-03-04
Proper use of parsers not dependant on language

Allactaga:


waqasshabbir:
Hi @Allactaga, I'm not sure if this is the best approach. Why not we try omitting skip words from the date formats as well like we do for date strings? Though format with `at` in them works. But it stills fail to correctly parse for non-english prepositions like below:

    In [4]: DateDataParser().get_date_data(u'02/12/2014 \xe0 15:08', date_formats=[u"%d/%m/%Y \xe0 %H:%M"])
    Out[4]: {'date_obj': datetime.datetime(2014, 2, 12, 15, 8), 'period': 'day'}

waqasshabbir:
Replaced by #194

PR #64, opened 2015-03-11
Added 'a las' spanish word

csalazar:
I've added 'a las' spanish words to skip. The translation has to keep a trailing space because after 'a las' comes the time.


Allactaga:
Hi Claudio. Can you please add a test with time describing situation referenced in your comment?

csalazar:
Eugene, I've added a more descriptive example.

PR #68, opened 2015-03-29
Add 'Fev' month name in FR lang

seagatesoft:
This forum: http://ile-maurice.com/ uses "Fev" instead of "Fév".


waqasshabbir:
Here's what @Allactaga suggests: since by French rules, `Fev` isn't a valid short name so we should replace it appropriately treating it as a typo.

PR #69, opened 2015-03-30
Added shortname for french days names

waqasshabbir:
Please review and merge required for http://www.diagonaledesfous.com


Allactaga:
Hm, so how does French themselves solve this duality of `Mar = Mardi/Mars` ? Do they shorten "Mars" ?

waqasshabbir:
Days are normally written with `Le` in the beginning like `Le Mar` for month it is always `Mar`.

redapple:
As far as I know and read online:

- days of the weeks in `fr_FR` are [shortened like this](https://fr.wikipedia.org/wiki/Semaine#De_nos_jours):
  - `mardi` becomes `mar.`

But other `fr_*` locales don't have the `.`(dot) dot at the end. See for example http://lh.2xlibre.net/values/abday/, `fr_FR` vs. `fr_BE`

Same for [abbreviated month names](https://fr.wikipedia.org/wiki/Mois#Abr.C3.A9viations):
- for `fr_FR`, "Mars" is not abbreviated in "Mar", but left as is

but other locales do shorten "Mars": http://lh.2xlibre.net/values/abmon/

waqasshabbir:
Hi @redapple thanks for your input! TIL about glibc files. I think we might want to consider adding context of `locales` to dateparser at some point in future. But for now, we can assume `fr_FR` for our purpose. In the meantime, we'll make a note of it somewhere in our docs. What do you think @Allactaga?

asadurski:
My opinion: Locale should be introduced when the differences in word forms are big enough to split detection into two separate languages. We're probably going to get there with Chinese (zh_CN and zh_TW), but for minor differences, like this one, I would vote for a variant within one language. We could use negative lookahead expressions like `(?!\.)` to make sure that's not an abbreviation.

waqasshabbir:
Made the change to drop Month name abbreviation unsupported by `fr_FR` locale. Thank you all for the help!

PR #71, opened 2015-03-31
Added 'min' string to Portuguese minute variations

horva:


Allactaga:
Please add a test for the case you are adding this.

data/languages.yaml
```diff
@@ -557,6 +557,7 @@ pt:
         - hora
         - horas
     minute:
+        - min
```
Allactaga: Let's try to keep full singular form as the start of the list and add other forms after.

PR #72, opened 2015-03-31
Added simplification for 'less then a minute ago'

horva:


Allactaga:
Please add a test for the case you are adding this.

Issue #52, opened 2015-02-04
Day/month/year order for different locales


Allactaga:
We should specify what day/month/year order is prefferable for different languages.
We can use [this information](http://en.wikipedia.org/wiki/Date_format_by_country) as a starting point and define `date_order = DMY` for each language in our file (don't forget to add a specific validation for this). Then, based on this info we should pass [proper arguments](https://labix.org/python-dateutil#head-b95ce2094d189a89f80f5ae52a05b4ab7b41af47) to dateutil parser.

marven:
Will start working on this.

alexis-via:
I am a new user of dateparser, and I was very disappointed to see that it doesn't take into account the language to guess the date format day/month/year vs month/day/year. For example:

```
>>> import dateparser
>>> dateparser.parse(u'2/12/2015', languages=['fr'])
datetime.datetime(2015, 2, 12, 0, 0)
```

In French, the day is always before the month, so the result should be datetime.datetime(2015, 12, 2, 0, 0). It's quite disappointing ; it means you ALWAYS have to give the date_formats as argument, otherwise you will have some big troubles !

waqasshabbir:
Hi @alexis-via, we're working on adding preferred date order for different languages. See PR #73.

Also, can you please share more details on the error you reported [here](https://github.com/m3nu/invoice2data/issues/5). This can be a lot of help as on our end all tests are passing successfully.

Thank you.

alexis-via:
@waqasshabbir OK, thanks for your answer !

About the problem I have that I reported on invoice2data, I am still trying to understand the exact issue and I'll report a bug if I confirm that the issue is in dateparser. I am in fact able to import dateparser from a python console (which contradics what I said on the github issue on invoice2data), but it never works from Odoo (odoo imports invoice2data which itself imports dateparser). It crashes on dateparser/conf.py on this line: https://github.com/scrapinghub/dateparser/blob/master/dateparser/conf.py#L44 and I don't understand why... it's like if get_data() wasn't doing anything. I'll try to get some help from senior python devs to have a better understanding of what happens and, if I confirm that it's dateparser fault and not odoo fault, I'll open a new bug report here.

juliusdanek:
Has this been resolved?

waqasshabbir:
Hi @juliusdanek not yet, it's a WIP. Should be resolved soon. Thanks!

juliusdanek:
Awesome. Very excited! Love the package

[...]

PR #73, opened 2015-04-03
Add date_order parameter for each language

marven:
Based on https://github.com/scrapinghub/dateparser/issues/52

### DIFF
--- a/dateparser/date.py
+++ b/dateparser/date.py
@@ -172,7 +172,16 @@ class _DateLanguageParser(object):
 
     def _try_dateutil_parser(self):
         try:
-            date_obj = date_parser.parse(self._get_translated_date())
+            date_order = self.language.info.get('date_order', '')
+            yearfirst, dayfirst = (False, False)
+            if date_order.startswith('Y'):
+                yearfirst = True
+            if 'DM' in date_order:
+                dayfirst = True
+            date_obj = date_parser.parse(
+                self._get_translated_date(),
+                dayfirst=dayfirst, yearfirst=yearfirst
+            )
             return {
                 'date_obj': date_obj,
                 'period': 'day',
--- a/dateparser/date_parser.py
+++ b/dateparser/date_parser.py
@@ -170,7 +170,7 @@ def dateutil_parse(date_string, **kwargs):
 
 
 class DateParser(object):
-    def parse(self, date_string):
+    def parse(self, date_string, dayfirst=False, yearfirst=False):
         date_string = unicode(date_string)
 
         if not date_string.strip():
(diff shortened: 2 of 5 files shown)


dateparser/date.py
```diff
@@ -172,7 +172,16 @@ def _try_freshness_parser(self):
 
     def _try_dateutil_parser(self):
         try:
-            date_obj = date_parser.parse(self._get_translated_date())
+            date_order = self.language.info.get('date_order', '')
+            yearfirst, dayfirst = (False, False)
+            if date_order.startswith('Y'):
+                yearfirst = True
+            if 'DM' in date_order:
+                dayfirst = True
```
Allactaga: What would you say if we refactor this block to something like
```python
date_order_params = self._get_date_order_params()
date_obj = date_parser.parse(self._get_translated_date(), **date_order_params)
```

dateparser/date_parser.py
```diff
@@ -170,7 +170,7 @@ def dateutil_parse(date_string, **kwargs):
 
 
 class DateParser(object):
-    def parse(self, date_string):
+    def parse(self, date_string, dayfirst=False, yearfirst=False):
```
Allactaga: Hm, maybe it's better to use `**kwargs` here. I mean we just passing these arguments to `dateutil_parse` function.

dateparser/languages/validation.py
```diff
@@ -10,7 +10,7 @@ class LanguageValidator(object):
                   'monday', 'tuesday', 'wednesday', 'thursday', 'friday', 'saturday', 'sunday',
                   'january', 'february', 'march', 'april', 'may', 'june', 'july',
                   'august', 'september', 'october', 'november', 'december',
-                  'year', 'month', 'week', 'day', 'hour', 'minute', 'second']
+                  'year', 'month', 'week', 'day', 'hour', 'minute', 'second', 'date_order']
```
Allactaga: We also need to add validation method for it that would check existence of `date_order` key and correctness of its value.

Allactaga:
Hi Marven,
Sorry for the late response. You've done great work here. I see that you've chosen USA default order (MDY) for the English language. But Europe and Australia uses DMY order. This is a tricky part and we from our side need to work on some configuration issues to resolve this properly. We'll get back to you when it's done.

waqasshabbir:
Hi @marven we've introduced `settings` to dateparser. Can you please refresh this PR according to new changes? It'll be good to be able to specify a custom date order using settings in addition to default order in `languages.yaml`. Thank you!

Digenis:
Defining locale boarders using languages is bound to cause confusion and mistakes.
year/month/day order is not a language feature
and it shouldn't surprise anyone if it's not even a country feature.
One proper solution is to use locales instead of languages for configuration.
Is there such a plan?

waqasshabbir:
Fixed in PR #187

PR #75, opened 2015-04-07
skipping when expression on thai

eLRuLL:
appears on `thailandtrailblazerclub.com`


runicer156:
@eLRuLL Look good for me.

PR #78, opened 2015-04-09
Feature uniform unit tests (test_date.py)

Allactaga:
This is updated version with fixed bugs of the previous PR for the same issue.

### DIFF
--- a/tests/test_date.py
+++ b/tests/test_date.py
@@ -8,8 +8,6 @@ from collections import OrderedDict
 import unittest
 from datetime import datetime, timedelta
 
-from dateutil.relativedelta import relativedelta
-
 from mock import Mock, patch
 
 from nose_parameterized import parameterized, param
@@ -24,6 +22,10 @@ class DateRangeTest(BaseTestCase):
         super(DateRangeTest, self).setUp()
         self.period_start = NotImplemented
         self.period_end = NotImplemented
+        self.result = NotImplemented
+        self.period_end = NotImplemented
+        self.period_size = NotImplemented
+        self.range = NotImplemented
 
     @parameterized.expand([
         param(begin=datetime(2014, 6, 15), end=datetime(2014, 6, 25), expected_length=10)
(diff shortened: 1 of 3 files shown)


tests/test_date.py
```diff
@@ -84,224 +126,348 @@ def test_date_arguments_and_date_range_with_default_post_days(self):
               expected_results=[datetime(2014, 12, 1)]),
     ])
     def test_should_one_date_for_each_month(self, low, high, expected_results):
```
waqasshabbir: It might be more readable if rephrased to `test_day_should_be_one_for_each_date_in_the_range_when_period_is_month`

Allactaga: What about `test_day_should_be_only_one_date_representing_each_month_when_period_is_month` then?

In your variant it is not very clear what `one` should it be.

tests/test_date.py
```diff
@@ -84,224 +126,348 @@ def test_date_arguments_and_date_range_with_default_post_days(self):
...
+        param('microseconds'),
+        param('some_period'),
     ])
-    def test_should_reject_easily_mistaken_dateutil_arguments(self, invalid_period):
-        low = datetime(2014, 6, 15)
-        period = datetime(2014, 6, 25)
-
-        with self.assertRaisesRegexp(ValueError, "Invalid period: {}".format(invalid_period)):
-            date.get_intersecting_periods(low, period, period=invalid_period).next()
+    def test_should_reject_easily_mistaken_dateutil_arguments(self, period_size):
```
waqasshabbir: Perhaps this test can be distinguished from [one above](https://github.com/scrapinghub/dateparser/pull/78/files#diff-1899ed0097831b86662064a1e60ee782R60) having the same name. Also, arguments for the above test can be wrapped by param function to keep the practice consistent.

Allactaga: Good catch with wrapping arguments, i'll fix that. Although those two tests are belongs to different testcases (this is the difference), but they are practically checking the same. How do you propose to change names?

waqasshabbir: I missed the the test case hierarchy - looking at it makes sense. I don't think we need to change the name.

tests/test_date.py
waqasshabbir: I modified my original comment. Does that make sense?

PR #79, opened 2015-04-20
Added chinese translation for 'half an hour'

csalazar:
Added language entry and test for translation of 'half an hour'.


Allactaga:
Google translate says that `时` is also used for `hour`, not only `time`. Do you think is it safe to add this to `skip`?

asadurski:
I confirm @Allactaga's doubts. The sign `'時'` should be added as `'hour'`. For instance `"2時30分"` should be equal to `'2:30'`.
@csalazar, would you like me to take care of the necessary corrections?

csalazar:
@asadurski sure Artur, go ahead!

PR #80, opened 2015-04-21
Feature improve french language

csalazar:
I've added new words and tests for French cases. Maybe can we add regexp support for ``skip`` attribute?


Allactaga:
Why have you decided to handle "less than" for French differently than for English and other cases we already have?

As for using regexp for `skip` I don't think that it is an option at the moment. Our translation is token based, and regexp is only used with simplifications.

PR #81, opened 2015-04-26
Proposal for new fr skip char: 'ã'

horva:


Allactaga:
I don't see that `ã` is valid French. Could you please pinpoint me to some documents? Also PR is missing test.

Allactaga:
Declined.

Issue #44, opened 2015-01-23
Skipped tests should be converted


Allactaga:
During the migration to the declarative languages approach we marked some tests as skipped, because the nature of code in those places almost completely changed. Now, when the behavior of working with languages is more likely finalized, those skipped tests should be rewritten to test that new code is still working for input from old cases. Tests should be designed in a way described [here](https://github.com/scrapinghub/dateparser/issues/31#issuecomment-71035831).

tsrdatatech:
I will assign this to myself since I am already doing issue #31

waqasshabbir:
Fixed by #85

Issue #31, opened 2014-12-27
Uniform unit tests


shirk3y:
It seems that we have plenty of methodologies used in unit tests.
Some of them are:

1.

        date = DateParser(language='cz').parse('pon 16. čer 2014 10:07:43')
        self.assertEqual(date.year, 2014)
        self.assertEqual(date.month, 6)
        self.assertEqual(date.day, 16)
        self.assertEqual(date.hour, 10)
        self.assertEqual(date.minute, 07)
        self.assertEqual(date.second, 43)

2.

       parser = DateParser()
        date_fixtures = [
            ('13 iunie 2013', datetime(2013, 6, 13)),
            ('14 aprilie 2014', datetime(2014, 4, 14)),
            ('18 martie 2012', datetime(2012, 3, 18)),
        ]

        for dt_string, correct_date in date_fixtures:
            parsed = parser.parse(dt_string)
            self.assertEquals(correct_date.date(), parsed.date())

3.

        @parameterized.expand([
        param('Sep 03 2014 | 4:32 pm EDT', datetime(2014, 9, 3, 21, 32)),
        param('17th October, 2034 @ 01:08 am PDT', datetime(2034, 10, 17, 9, 8)),
        param('15 May 2004 23:24 EDT', datetime(2004, 5, 16, 4, 24)),
        param('15 May 2004', datetime(2004, 5, 15, 0, 0)),
        param('Nov 25 2014 10:17 pm EST', datetime(2014, 11, 26, 4, 17)),
        ])

4.

       date = DateParser(language='pl').parse('Środa, 26 listopada 2014 10:11:12')
       self.assertEqual(date.timetuple()[:6], (2014, 11, 26, 10, 11, 12))

Maybe we could create simple method/function for asserting correct date? Eg.
        
        date = DateParser(language='en').parse('Tue, 25 Dec, 2012 12:00')
        self.assertDate(date, 2012, 12, 25, 12, 0)

Depending of number of params given to `assertDate` we check date with given resolution

horva:
What about having python file having key: value entries, key representing arg of dateparser.parse and value the result. Then we can iterate through all the entries within single test, asserting with dateparser.parse, which would make adding new cases much easier.

Suggested format:

    from datetime import datetime
    TODAY = datetime(2015, 1, 1)  # for freshness and other relative calculations

    TEST_DATA = {
        'en': {
            'sep 19': datetime(2014, 9, 19),
            ...
        },
        'other': {
[...]

PR #85, opened 2015-05-08
Feature better tests

Allactaga:
Last changes required for Issues #31 and #44

### DIFF
--- a/tests/test_date_parser.py
+++ b/tests/test_date_parser.py
@@ -208,11 +208,24 @@ class TestDateParser(BaseTestCase):
         param('November 19, 2014 at noon', datetime(2014, 11, 19, 12, 0)),
         param('December 13, 2014 at midnight', datetime(2014, 12, 13, 0, 0)),
         param('Nov 25 2014 10:17 pm EST', datetime(2014, 11, 26, 3, 17)),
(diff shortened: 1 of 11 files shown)


tests/test_date_parser.py
```diff
@@ -394,106 +491,12 @@ def then_period_is(self, period):
...
-
-    def test_finding_no_language_should_work_for_numeric_dates(self):
-        dp = DateParser()
-        self.assertEqual(datetime(2014, 8, 13).date(),
-                         dp.parse('13 Ago, 2014').date())
-
-        parsed_date = dp.parse(u'13/08/2014')
-        self.assertEqual(datetime(2014, 8, 13).date(), parsed_date.date())
+    def then_error_was_raised(self, error_cls, error_message):
+        self.assertIsInstance(self.error, error_cls)
```
waqasshabbir: We have this definition at [here](https://github.com/scrapinghub/dateparser/pull/85/files#diff-36128d81467a740315c9e5f49860c291R188) too. Why not move it to `BaseTestCase?`

Allactaga: I was thinking about extracting common methods to domain-oriented mixins, but there were only few of them, so I left it as it were. I'll move this one to `BaseTestCase` and if there would be too many such methods in `BaseTestCase` at some point we can move them to separate mixin then.

PR #86, opened 2015-05-20
Feature iso date format with non english language

### LINKED ISSUE #45: Dates "08/17/14 17:00 PM (PDT)" not parsed.
Timezones recognition should be more generic

waqasshabbir:
Fixes iso datestamp format parsing issue.

`parse('2015-05-02T10:20:19+0000', languages=['fr'])`
 is not parsed, while
`parse('2015-05-02T10:20:19+0000', languages=['en'])` has good result.

Fixes #45


dateparser/languages/language.py
```diff
@@ -24,40 +26,51 @@ def __init__(self, shortname, language_info):
             if isinstance(value, int):
                 simplification[key] = str(value)
 
+        self._cached = self
```
Allactaga: Why do we aliasing `self` as `self._cached` ?

dateparser/languages/loader.py
```diff
@@ -35,7 +36,9 @@ def _load_data(self):
         else:
             data = self.file.read()
         data = load_yaml(data)
-        base_data = data.pop('base', {})
+        base_data = data.pop('base', {'skip': []})
+        # also add any skip tokens from custom settings
+        base_data['skip'] = settings.SKIP_TOKENS
```
Allactaga: Isn't it replacing and not appending?

dateparser/timezone_parser.py
```diff
@@ -4,14 +4,13 @@
...
-TIMEZONE_REGEX_PATTERN = r'(\b|\d)%s$'
-
 
 def pop_tz_offset_from_string(date_string, as_offset=True):
-    for name, info in _tz_offsets.iteritems():
+    for name, info in _tz_offsets:
         timezone_re = info['regex']
         if timezone_re.search(date_string):
-            date_string = timezone_re.sub(r'\1', date_string)  # \1 = (\b|\d) in TIMEZONE_REGEX_PATTERN
+            # \1 = (\b|\d) in TIMEZONE_REGEX_PATTERN
```
Allactaga: Is this comment make sense now, after we deleted `TIMEZONE_REGEX_PATTERN` ?

dateparser/languages/language.py
waqasshabbir: I did that because we generate a new instance of Language whenever a new skip token is added and identified like [here](https://github.com/scrapinghub/dateparser/pull/86/files#diff-70e80316a9eb316ceb7f7f1c8f63d152R42). Is there a better practice?

dateparser/languages/loader.py
waqasshabbir: Thank you! good catch. Let me fix it.

dateparser/timezone_parser.py
waqasshabbir: Missed it. Let me fix it now.

dateparser/languages/language.py
Allactaga: But now original methods are never called, they all proxying to `self._cached`. Is there a point?
How do you feel about instead of implementing this on a `Language` class level go level above and try to implement this on `LanguageDataLoader` level, or `BaseLanguageDetector` level or even `DateDataParser` level?

waqasshabbir: I wasn't sure if direct assignments to `self` like `self = SomeInstance()` a good practice or not. Hence, used `_cached` attrib. I tried implementing it on almost all levels. Higher levels slowed down the speed and had other issues but due to speed imperative I didn't look into fixing those issues. Tests took about 16 seconds to finish when I tried implementing it on DateDataParser level.

dateparser/timezone_parser.py
```diff
@@ -22,18 +20,32 @@ def convert_to_local_tz(datetime_obj, datetime_tz_offset):
...
-            'offset': timedelta(seconds=tz_info[1]),
-        }
-        for tz_info in timezone_info_list
-    }
+
+    def get_offset(tz_obj, repl='', replw=''):
+        return (
+            tz_obj[0],
+            {
+                'regex': re.compile(re.sub(repl, replw, regex % tz_obj[0]), re.IGNORECASE),
```
Allactaga: wouldn't it be clearer to pass `regex` with other function parameters?

dateparser/timezones.py
```diff
@@ -1,393 +1,448 @@
...
-    ('YEKT', 18000),
-    ('YERST', 14400),
-    ('YERT', 10800),
-    ('YPT', -28800),
-    ('YST', -32400),
-    ('YWT', -28800),
-    ('zzz', 0)
+    {
+        'regex_patterns':
+            [r'(.)\(?%s\)?$',],
```
Allactaga: About using `(.)`:
```python
>>> re.sub(r'(.)\(?UTC\)?$', r'\1', "19:00(MUTC)")
'19:00(M'```

File: dateparser/timezones.py [hunk 6]
```diff
@@ -1,393 +1,448 @@
...
+             ('YERST', 14400),
+             ('YERT', 10800),
+             ('YPT', -28800),
+             ('YST', -32400),
+             ('YWT', -28800),
+             ('zzz', 0)]
+    },
+    {
+        'regex_patterns': 
+            [r'(.)%s',],
```
Allactaga: It does not have to be end of the string?

dateparser/languages/language.py
Allactaga: What have caused those speed issues? I mean we are generating updated languages once and then just reusing them, right?

waqasshabbir: If we would customize settings before importing `parse`, then it would work alright. But if we'd customize settings after importing `parse`, it won't work hence requiring generation of languages again.

Allactaga: We can address this isue when we'd have more configuration changes to work with.
What about using `LanguageDataLoader` for now then?

dateparser/timezones.py
waqasshabbir: Not essentially.e.g. the example string given in #45

Allactaga: But we don't use any parentheses in this group of regexps.

waqasshabbir: Not really sure, why I omitted it. Now I looked into it carefully, may be we should add end of the string check. Thank you!

waqasshabbir:
@Allactaga please take a look again.

dateparser/timezones.py
waqasshabbir: I agree that using `.` is somewhat greedy. But is the date string in question representing a valid date? `MUTC` is not a valid timezone abbreviation, hence, should not be cleaned out of the string. What do you think?

dateparser/date.py
```diff
@@ -304,6 +314,11 @@ def get_date_data(self, date_string, date_formats=None):
         TODO: Timezone issues
 
         """
+        global default_language_loader
+        if settings.SKIP_TOKENS != self._skip_tokens:
+            default_language_loader = LanguageDataLoader()
+            self = DateDataParser(*self._default_args)
```
Allactaga: What do you think is happening in this block?

dateparser/languages/loader.py
```diff
@@ -34,10 +36,14 @@ def _load_data(self):
             data = get_data('data', 'languages.yaml')
         else:
             data = self.file.read()
-        data = load_yaml(data)
-        base_data = data.pop('base', {})
+        
+        if not self._raw_data:
```
Allactaga: Does it need to be an attribute? Also, how is it different from data?

dateparser/timezones.py
Allactaga: `MUTC` of course should not be cleaned out of the string. But should `UTC` be cleaned out of string with `MUTC` ?

dateparser/date.py
waqasshabbir: If `DateDataParser` instance's `_skip_tokens` differ from `settings.SKIP_TOKENS` it creates a new instance of `LanguageDataLoader` and assigns it to `default_language_loader` which is then used in `DataDataParser.__init__` while creating a new instance for `DateDataParser` with originally provided arguments and updating reference to self. This ensures new skip tokens are accounted for while detecting language and subsequently parsing date string.

dateparser/languages/loader.py
waqasshabbir: I wanted to use it as a static variable so we don't have to load data from yaml every time we create a new `LanguageDataLoader` instance. Fixed it for that.

dateparser/timezones.py
waqasshabbir: I see. What do you suggest we should do to address this?

dateparser/date.py
Allactaga: You understand that this is happening every time you call `get_date_data` because you updating local variable `self` and not the instance itself, right?
Also, as for now we don't expose our settings interface to user, let's assume it's not changing and move this check to `__init__` method, getting rid of `global` in the same time. Instead, create a class method `get_language_loader` and use class attribute to lazy instantiate default language loader.

dateparser/languages/loader.py
Allactaga: We would still load data from yaml with `get_data` or `self.file.read()`. Though we don't call entire `_load_data` method if `_data` attribute is not None. I am really not sure what are you trying to achieve with `_raw_data`

dateparser/timezones.py
Allactaga: How about deleted `(\b|\d)` ?

dateparser/languages/loader.py
waqasshabbir: Does the following make sense? I am trying to force parsing of yaml file only once. Without this implementation, it will be parsed every time a new `LanguageDataLoader` instance is created.

```
   def _load_data(self):

        if not LanguageDataLoader._raw_data:
            if self.file is None:
                data = get_data('data', 'languages.yaml')
            else:
                data = self.file.read()
            LanguageDataLoader._raw_data = load_yaml(data)
```

Allactaga: *New* `LanguageDataLoader` is created for a good reason only. For now, while we not yet expose settings for the user, do you expect this to happen more that once? Also, we don't want loader to be able to load *only one data*. What if new instance was defined with different `file` ?

waqasshabbir: Removed `_raw_data` attribute.

dateparser/date.py
waqasshabbir: Made changes accordingly, please take a look again.

dateparser/timezones.py
```diff
@@ -1,393 +1,448 @@
...
-    ('YEKT', 18000),
-    ('YERST', 14400),
-    ('YERT', 10800),
-    ('YPT', -28800),
-    ('YST', -32400),
-    ('YWT', -28800),
-    ('zzz', 0)
+    {
+        'regex_patterns':
+            [r'(^\w|\w$|\s|\d)\(?%s\)?$'],
```
waqasshabbir: `\b` wasn't suitable because it would also match parenthesis and hence not omit them failing case likes #45

Allactaga: Hm. Can you please give me a code example. I can't think of the valid case. It is working like this though:
 
```python
>>> re.sub(r'(\b|\d)\(?UTC\)?$', r'\1', "19:00(UTC)")
'19:00'
```

waqasshabbir: Here:
```
In [14]: re.sub(r'(\b|\d)\(?UTC\)?$', r'\1', '19:00 (UTC)')
Out[14]: '19:00 ('
```

Allactaga: Okay, i see. So what are `^\w` and `\w$` there for?

Allactaga: I am also thinking in the direction of `(\W|\d|_)`

tests/test_date.py
```diff
@@ -413,18 +416,7 @@ def given_now(self, year, month, day, **time):
...
-            language_loader = LanguageDataLoader()
-            language_map = date.default_language_loader.get_language_map()
-
-            ordered_languages = OrderedDict([
-                (shortname, language_map[shortname])
-                for shortname in restrict_to_languages
-            ])
-            language_loader._data = ordered_languages
-            self.add_patch(patch('dateparser.date.default_language_loader', new=language_loader))
+        self.parser = date.DateDataParser(languages=restrict_to_languages, **params)
```
Allactaga: This change is not restricting languages in loader to a limited subset anymore

tests/test_date.py
```diff
@@ -449,6 +441,9 @@ def when_multiple_dates_are_parsed(self, date_strings):
     def then_date_was_parsed(self):
         self.assertIsNotNone(self.result['date_obj'])
 
+    def then_date_was_not_parsed(self):
```
Allactaga: This method is not used

waqasshabbir: Removed.

tests/test_date.py
waqasshabbir: Fixed.

PR #90, opened 2015-06-17
using strict markup in README.rst to avoid breaking description in PyPI

eliasdorneles:
Hey fellows!

So, the [description page in pypi](https://pypi.python.org/pypi/dateparser) is basically the concatenation of `README.rst` and `HISTORY.rst`.
The ReST markup supported in PyPI is more strict (e.g.: you can't use autodocs, it also does some more strict validation), you can check if it will work in PyPI using `restview --strict README.rst`.

This PR removes the unsupported markup from `README.rst` and adjust the titles hierarchy to match `HISTORY.rst`.

Looks good?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -19,6 +19,12 @@ dateparser -- python parser for human readable dates
 any string formats commonly found on web pages.
 
 
+Documentation
+=============
+
+Documentation can be found `here <https://dateparser.readthedocs.org/en/latest/>`_.
+
+
 Features
 ========
 
--- a/README.rst
+++ b/README.rst
@@ -20,7 +20,8 @@ any string formats commonly found on web pages.
 
 
 Features
---------
+========
+
 * Generic parsing of dates in English, Spanish, Dutch, Russian and several other langauges and formats.
 * Generic parsing of relative dates like: ``'1 min ago'``, ``'2 weeks ago'``, ``'3 months, 1 week and 1 day ago'``.
 * Generic parsing of dates with time zones abbreviations like: ``'August 14, 2015 EST'``, ``'July 4, 2013 PST'``.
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -28,16 +29,14 @@ Features
...
 
 Usage
------
-The most straightforward way is to use the :func:`dateparser.parse` function,
-that wraps around most of the functionality in the module.
+=====
 
-.. automodule:: dateparser
-   :members: parse
+The most straightforward way is to use the `dateparser.parse` function,
```
waqasshabbir: @eliasdorneles @Allactaga will it be okay to leave this part? I don't get any red flags for this when I execute the above command: `restview --strict ../README.rst`

Allactaga: Yeah, this PR doesn't have any issues with `restview`, though It would be cool if we find a way to not use `:func:` in the `README.rst` (so pypi won't break markdown) but still keep it for `docs/` so it would appear in `readthedocs`

waqasshabbir: How about re-writing `dateparser.parse ` as  \`dateparser.parse <#dateparser.parse>\`_

waqasshabbir:
Hi folks! I took the liberty to make some changes in this PR. Please take a look. According to the command below, it seems to be passing fine now.

`python setup.py --long-description | rst2html.py > output.html`

eliasdorneles:
+1, nice work, Waqas, LGTM! :)

Just a note, I found easier to check this with `restview --pypi-strict <(python setup.py --long-description)` - might be worthy to note this somewhere (I just found out about --pypi-strict, in latest restview it's supposed to use additional strict rules for PyPI).

It seems the PR needs syncing -- @waqasshabbir can you take care of that?

waqasshabbir:
Sure, we will note that under the release guidelines which @Allactaga intends to compile as soon as he gets a chance. Thank you!

PR #99, opened 2015-07-24
fixes to work under Python3 and PyPy

eliasdorneles:
Hey, folks,

@Allactaga, @csalazar, @kmike and I teamed up to port dateparser to Python 3 -- we got excited and also made sure the tests are running in PyPy as well. :)

Some interesting changes:

* We got rid of the new_relativedelta class because after realizing we don't need it anymore
* We also got rid of the decoding to bytes that was being done in parse_with_formats, as discussed in previous PR: https://github.com/scrapinghub/dateparser/pull/15/files#r21062682

That's it, folks -- thanks everyone for the help! =)

### DIFF
--- a/dateparser/date_parser.py
+++ b/dateparser/date_parser.py
@@ -2,40 +2,20 @@
 from __future__ import unicode_literals
 
 import calendar
-import re, sys
+import re
+import sys
 from datetime import datetime
 from collections import OrderedDict
 
+import six
 from dateutil import parser
 from dateutil.relativedelta import relativedelta
 from dateparser.timezone_parser import pop_tz_offset_from_string, convert_to_local_tz
 
-from conf import settings
+from .conf import settings
 
 
 binary_type = bytes if sys.version_info[0] == 3 else str
-
-
-class new_relativedelta(relativedelta):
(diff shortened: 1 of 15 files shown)


eliasdorneles:
@Allactaga @asadurski @waqasshabbir So what do you think, fellows? Who can merge this? :)

I'd love to see a release with the new support for Python 3! :)

dateparser/date_parser.py
```diff
@@ -72,7 +52,7 @@ def get_period(res):
             ('month', ['month']),
             ('year', ['year']),
         ])
-        for period, markers in periods.iteritems():
+        for period, markers in six.iteritems(periods):
             for marker in markers:
                 if getattr(res, marker) is not None:
```
kmike: hasattr?

asadurski:
@eliasdorneles Already taking care of it. It's definitely enough for a new release.

dateparser/date_parser.py
waqasshabbir: result object has attributes with values set to None, hence, using getattr and check for value is needed here. Since, `weekday` can be `0`, `is not None` predicate is used.

eliasdorneles:
@asadurski yay! :)

PR #103, opened 2015-08-06
Added Belarussian language.

wistful:


data/languages.yaml
```diff
@@ -1811,3 +1811,124 @@ ph:
         - kahapon: 1 araw
         - ngayon: 0 segundo
         - isang araw: 2 araw
+
+
+be:
```
waqasshabbir: Isn't the code for Belarus is `by`? AFAIK `be` is for `Belgium`.

data/languages.yaml
```diff
@@ -1811,3 +1811,124 @@ ph:
...
+    name: Belarussian
```
waqasshabbir: Looks like the correct spellings are `Belarusian`.

data/languages.yaml
```diff
@@ -1811,3 +1811,124 @@ ph:
...
+        - Субота
+        - Суб
+        - Сб
+    sunday:
+        - Нядзеля
+        - Няд
+        - Нд
+
+    january:
+        - Студзеня
```
waqasshabbir: Does `Belarusian` have short names for months? If so, it will be great to have them specified over here.

waqasshabbir:
Good job @wistful! Please take care of the comments above and we'll be ready to merge.

data/languages.yaml
immerrr: It does, but short names are not widely used.  Anyway, according to linux locale, they are
```
"Стд"
"Лют"
"Сак"
"Крс"
"Тра"
"Чэр"
"Ліп"
"Жнв"
"Врс"
"Кст"
"Ліс"
"Снж"
```

data/languages.yaml
immerrr: +1

data/languages.yaml
immerrr: Also +1

data/languages.yaml
```diff
@@ -1811,3 +1811,124 @@ ph:
...
+        - ўчора: 1 дзень
+        - сення: 0 дзён
+        - сёння: 0 дзён
+        - хв: хвіліну
+        - сек: секунду
+        - ^гадзіна: 1 гадзіна
+        - ^гадзіну: 1 гадзіну
+        - ^хвіліну: 1 хвіліну
+        - ^секунду: 1 секунду
+        - некалькі секунд: 44 секунды
```
immerrr: *сякýнд

immerrr: "у першым складзе перад нацiскам" applies to some previous lines, too.

immerrr:
Great stuff!

data/languages.yaml
wistful: Done.

data/languages.yaml
wistful: Keep both cases.

data/languages.yaml
wistful: Done.

data/languages.yaml
wistful: Done.

data/languages.yaml
immerrr: I'm not sure about leaving in grammatically incorrect forms. It will increase recall, but I can't remember any precedents in other languages, e.g. there's no "сикунд" in Russian. @Allactaga what do you think?

Allactaga: We do not provide support for incorrect forms and typos. Sure some sites can have a misspelling in their date representation code and such support would be handy in those cases, however some of our user can rely on the fact that incorrect dates should not parse. We have plans to add support in the future for configuring parser manually so it can handle user provided forms. In the mean time typos should be handled on user side (f.ex. with `re.sub`)

wistful: I guess the rule was changed few years ago.
I kept both cases as all online translators translate 'second' as 'секунда' not 'сякунда'.
I assume the both versions are correct.

immerrr: @wistful i guess you're right, shame on me and my rusty Belarusian, borrowed words are excluded from that rule.

wistful: incorrect words were reverted.

wistful:
Looks like ready to merge.

waqasshabbir:
Thanks @wistful @immerrr! Merging now.

PR #104, opened 2015-08-06
Removed check to assert position of abbreviated timezone (e.g. EDT) at the end of the string.

waqasshabbir:
I ran into this scenario while working with dates on usnews.com.

Previously,

```
In [1]: from dateparser import parse
pars
In [2]: parse('Wed Aug 05 12:00:00 EDT 2015')

In [3]: parse('Wed Aug 05 12:00:00 2015 EDT')
Out[3]: datetime.datetime(2015, 8, 5, 21, 0)
```

After the changes,

```
In [1]: from dateparser import parse

In [2]: parse('Wed Aug 05 12:00:00 2015 EDT')
Out[2]: datetime.datetime(2015, 8, 5, 21, 0)

In [3]: parse('Wed Aug 05 12:00:00 EDT 2015')
Out[3]: datetime.datetime(2015, 8, 5, 21, 0)
```

### DIFF
--- a/dateparser/timezones.py
+++ b/dateparser/timezones.py
@@ -6,7 +6,7 @@
 timezone_info_list = [
     {
         'regex_patterns':
-            [r'(\W|\d|_)\(?%s\)?'],
+            [r'(\W|\d|_)\(?%s(\)|$|[^\w+])'],
         'timezones':
             [('ACDT', 37800),
              ('ACST', 34200),
--- a/dateparser/timezones.py
+++ b/dateparser/timezones.py
@@ -6,7 +6,7 @@
 timezone_info_list = [
     {
         'regex_patterns':
-            [r'(\W|\d|_)\(?%s(\)|$|[^\w+])'],
+            [r'(\W|\d|_)%s($|\W)'],
         'timezones':
             [('ACDT', 37800),
              ('ACST', 34200),
--- a/dateparser/timezones.py
+++ b/dateparser/timezones.py
@@ -6,7 +6,7 @@
 timezone_info_list = [
     {
         'regex_patterns':
-            [r'(\W|\d|_)\(?%s\)?$'],
+            [r'(\W|\d|_)\(?%s\)?'],
         'timezones':
             [('ACDT', 37800),
              ('ACST', 34200),
--- a/dateparser/date_parser.py
+++ b/dateparser/date_parser.py
@@ -10,8 +10,9 @@ from collections import OrderedDict
 import six
 from dateutil import parser
 from dateutil.relativedelta import relativedelta
-from dateparser.timezone_parser import pop_tz_offset_from_string, convert_to_local_tz
 
+from .timezone_parser import pop_tz_offset_from_string, convert_to_local_tz
+from .utils import strip_braces
 from .conf import settings
 
 
@@ -171,6 +172,7 @@ class DateParser(object):
         if not date_string.strip():
             raise ValueError("Empty string")
 
+        date_string = strip_braces(date_string)
         date_string, tz_offset = pop_tz_offset_from_string(date_string)
 
         date_obj, period = dateutil_parse(date_string)
--- a/dateparser/timezone_parser.py
+++ b/dateparser/timezone_parser.py
@@ -2,10 +2,12 @@
 import re
 from datetime import datetime, timedelta
 
-from dateparser.timezones import timezone_info_list
+from .timezones import timezone_info_list
+from .utils import strip_braces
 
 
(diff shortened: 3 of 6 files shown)


Allactaga:
First, you should also add tests to `test_timezone_parser.py` as you are making changes in timezones.
Second: `Wed Aug 05 12:00:00 EDTERR 2015`.

waqasshabbir:
Thank you @Allactaga! Please review changes again.

dateparser/timezones.py
```diff
@@ -6,7 +6,7 @@
 timezone_info_list = [
     {
         'regex_patterns':
-            [r'(\W|\d|_)\(?%s\)?$'],
+            [r'(\W|\d|_)\(?%s(\)|$|[^\w+])'],
```
Allactaga: Could you explain this regex please?

waqasshabbir: it checks for three possibilities: Is the character next to timezone abbreviation e.g. `EDT` a closing bracket? or is end of the string? or is *not* any word character [a-zA-Z0-9_]

Allactaga: 1. There is `\W` for `not \w`
2. `+` inside `[]` is just a plus sign
3. Isn't `\)` is part of _not_ any word characters? Also may be for clarity it's better to use another regex in the list (would be slower), that handles matching "(...)" instead of "(..." or "...)".

waqasshabbir: Sure, that gives us expression `(\W|\d|_)\(?%s($|\W)`. The revised expression only checks for either the end of string or not any word character next to timezone abbreviation. It might look repetitive to have another regex just to bifurcate for this binary difference. What do you think?

Allactaga: Hmmm... Also, wouldn't `\(` be matched as part of first `\W`?

waqasshabbir: If we don't match it separately, the resultant date string (stripped of timezone info) looks like below. 
```
In [3]: regex  = re.compile(r'(\W|\d|_)PDT($|\W)', re.IGNORECASE)

In [4]: regex.sub(r'\1', '08/17/14 17:00 (PDT)')
Out[4]: '08/17/14 17:00 ('
```

For which, dateutil parser returns `Unknown string format` error and hence no date is parsed. So, either we can exclude `(` from the match group or replace it from the string. Either at [dateparser](https://github.com/scrapinghub/dateparser/blob/master/dateparser/date_parser.py#L174) or [timezone parser](https://github.com/scrapinghub/dateparser/blob/master/dateparser/timezone_parser.py#L12).

Allactaga: Hm. Why did it become our issue in the first place?
```ipython
In [1]: from dateparser.languages.loader import default_language_loader

In [2]: en = default_language_loader.get_language('en')

In [3]: en.translate("08/17/14 17:00 (")
Out[3]: u'08/17/14 17:00 '
```

waqasshabbir: Like visible from the `date_parser.py` code in the comment above, we don't translate date at this level. Firstly parenthesis wrapped timezone issue was reported [here](https://github.com/scrapinghub/dateparser/issues/45)

Allactaga: Okay, so timezones is *internal* mechanics of our parser working with `dateutil` and applied only there. I see...

I really don't like existing solution of just adding `(` to regex because of `17:00 (EDT)`, `17:00 [EDT]`, `17:00 {EDT}`, `17:00 <EDT>`, `17:00 | EDT`, `17:00 ^EDT` and so on... Any ideas?

waqasshabbir: how about this expression `(\W|\d|_)\W?%s($|\W)`?

Allactaga: I think we should not rush this and think for a while. May be move timezone extraction to a level higher.

waqasshabbir: Since, braces don't provide any contextual meaning I believe we can strip them off the date string at the time of extracting timezone. What benefits do you imagine we will have by moving timezone extraction to a level higher?

waqasshabbir:
@Allactaga any comments?

Allactaga:
@waqasshabbir on what in particular?

waqasshabbir:
@Allactaga on using a function to strip braces off the date string.

Allactaga:
I don't have any counter-examples in my head in the moment. So we can use it and open an issue if this would intervene with some languages or date form. One thing is I'd put call to it in `DateParser.parse` itself and not `pop_tz_offset_from_string` because AFAIU it's dateutil itself not working very good with brackets and not our timezone module.

PR #105, opened 2015-08-25
Introducing regex module

asadurski:
Introduction of `regex` module (used instead of standard `re`) will allow for more flexible approach to regular expressions.

### DIFF
--- a/dateparser/calendars/hijri.py
+++ b/dateparser/calendars/hijri.py
@@ -1,7 +1,7 @@
-import re
 from datetime import datetime
 
 from umalqurra.hijri_date import HijriDate
+import regex as re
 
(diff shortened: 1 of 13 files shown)


sardok:
It would be better to put it in ```requirements.txt``` as well.

asadurski:
Yup, it's already there: https://github.com/scrapinghub/dateparser/commit/560756f044ace493a85b387fa04c4d0eefdb56db#diff-b4ef698db8ca845e5845c4618278f29aR3

sardok:
Sorry, artur, missed that.

waqasshabbir:
This looks good. Can we raise a warning when `re` module is used? Urging to use `regex` instead.

PR #123, opened 2015-10-28
Handle time information with punctuation symbol look-alike e.g. '\u2019'

seagatesoft:


Allactaga:
I'd expect to have general solution for all look-alike punctuation symbols that would be applied to any language, not just one particular case for French language.

seagatesoft:
@Allactaga Can we replace all of those punctuation symbols look-alike with `'` in `dateparser.dateparser.date.sanitize_date()` (I have updated the PR using this method)?

Allactaga:
`sanitize_date` would indeed be a good place for this, as it would be done just once for all languages.

However, due to unique agenda of this transformation I think it should be placed in a separate function which we call later from `sanitize_date`.

Also, may be it would be good idea to put those translations in separate file to load and apply, rather to hardcode them.

@asadurski, @waqasshabbir what do you guys think?

waqasshabbir:
How about adding universal simplifications here under `skip`. https://github.com/scrapinghub/dateparser/blob/master/data/languages.yaml#L2

asadurski:
My thoughts exactly, @waqasshabbir. I would rather keep it all in one file.

Allactaga:
Then it would be applied once **per each** language.

asadurski:
Are you referring to performance drop?
As for functionality, I think we're fine with universal replacements as long as we always replace with something, never delete.

Allactaga:
Yes, I was talking about performance decrease. Also storing such technical data separately would help to keep our language data file clean. I leave it up to you to decide.

waqasshabbir:
Since, this transformation is related to normalization and because we are just correcting here the real  `’` to typewrite apostrophe `'`, `sanitize_date` looks like a good place for it. Also, having a separate function for this one substitution looks like an overkill. May be in future if we come across similar examples, we can have a `normalize_date` function. And I agree with @Allactaga  `languages.yaml` might not be a good place for it.

seagatesoft:
I have collected apostrophe and aposthrope-alike characters from:

* https://en.wikipedia.org/wiki/Apostrophe#Unicode
* https://en.wikipedia.org/wiki/Prime_(symbol)#Representations
* https://en.wikipedia.org/wiki/Quotation_mark#Unicode_code_point_table
* https://www.cs.sfu.ca/~ggbaker/reference/characters/

dateparser/date.py
```diff
@@ -14,6 +14,12 @@
 from dateparser.languages.detection import AutoDetectLanguage, ExactLanguages
 
 
+APOSTROPHE_LOOK_ALIKE_CHARS = [
+    u'\u2019', u'\u02bc', u'\u02bb', u'\u055a', u'\ua78c',
+    u'\uff07', u'\u2032', u'\u2035', u'\u02b9', u'\uff07',
```
waqasshabbir: Hi Sigit! it looks great. How about using the literal character here instead of the utf code for better readability's sake?

seagatesoft: @waqasshabbir  Hmm, isn't it more readable to use the UTF code since those characters are looks alike? I named the array `APOSTROPHE_LOOK_ALIKE_CHARS` so developer reading this will know that those characters are similar to apostrophe (in fact u"\u2019" is the correct apostrophe character). But maybe I am being subjective here, what do you think @asadurski ?

asadurski: My position is - take some of both worlds, so either use literals and leave their UTF codes in the comment, or the other way around.
Literals are useful if one is searching in rendered page, UTF code is useful if one is checking code output, HTML etc.
In the future we would move those to separate YML file.

redapple: What about using `\N{name}`?

```
>>> u"\N{RIGHT SINGLE QUOTATION MARK}"
u'\u2019'
>>> u"\N{MODIFIER LETTER APOSTROPHE}"
u'\u02bc'
```

Youwotma: TIL about `\N{name}`

redapple: `\uxxxx` is also handy in comments.
I only learned about `\N{name}` recently myself: https://docs.python.org/2/reference/lexical_analysis.html#string-literals

asadurski: Neat! But now I am puzzled... In this case the descriptive name is enough for representation for human beings, but it's still good to have the UTF code nearby (well, actually the `\N{name}` is sufficient in most cases, so even the comment is not too necessary). And the rendered form is easy to find through printing the list... So how about:
```
u'\N{RIGHT SINGLE QUOTATION MARK}',   # u'\u2019'
u'\N{MODIFIER LETTER APOSTROPHE}',    # u'\u02bc'
u'\N{MODIFIER LETTER TURNED COMMA}',  # u'\u02bb'
u'\N{ARMENIAN APOSTROPHE}',           # u'\u055a'
u'\N{LATIN SMALL LETTER SALTILLO}',   # u'\ua78c'
u'\N{PRIME}',                         # u'\u2032'
u'\N{REVERSED PRIME}',                # u'\u2035'
u'\N{MODIFIER LETTER PRIME}',         # u'\u02b9'
u'\N{FULLWIDTH APOSTROPHE}',          # u'\uff07'
```
Doesn't it look pretty? :smile:

waqasshabbir: Agreed with @redapple `\uxxxx` is handy in comments. Also agrees with @seagatesoft's argument that using unicode strings are more distinct than *look alike* literal characters. Unicode names look great -- they might appear a little verbose yet suits our purpose. +1 for @asadurski's above representation.

seagatesoft: @waqasshabbir @redapple 
Thanks guys, I have applied @asadurski suggestion, please check again.

waqasshabbir:
@seagatesoft looks great! merging it now.

PR #124, opened 2015-11-02
Added hijri calendar

sardok:
Introducing new calendar class; **HijriCalendar** (more info about Hijri calendar could be found [here](https://en.wikipedia.org/wiki/Islamic_calendar).
It's build on top of **CalendarBase** class, however return type of ```get_date``` is inconsistent with **JalaliCalendar**'s.
**JalaliCalendar** returns ```datetime``` object where **HijriCalendar** returns ```{'period': '__period__', 'date_obj': '__datetime__'}``` as this library's generic ```parse``` function would return.
**HijriCalendar** uses internal functions and classes, so it knows about the period, after parsing & converting the given datetime string.
Since **CalendarBase**'s ```get_date``` expects raw datetime string, i think, it is better to be compatible with generic ```parse``` return type.


dateparser/calendars/hijri.py
```diff
@@ -0,0 +1,94 @@
...
+                return '{:02d}'.format(digit - 10)
+            fixed_date_string, nreplaced = re.subn(
+                r'(?<!\d)(29|30)', _sub_fn, translated, 1)
+            if not nreplaced:
+                continue
+
+            date_data = self._parser_get_date(fixed_date_string, date_formats, languages)
+            date_obj = date_data.get('date_obj')
+            if date_obj:
+                print 'date_obj: {}'.format(date_obj)
```
waqasshabbir: Hi Sinan, great work! Is this `print` statement left from debugging?

sardok: Yes, removed thanks.

asadurski:
@sardok: Approved. Thank you!

PR #127, opened 2015-11-13
Addition to Chinese days of the week

cassac:
Adds two more versions for saying days of the week in Chinese.

### DIFF
--- a/data/languages.yaml
+++ b/data/languages.yaml
@@ -990,18 +990,33 @@ cn:
 
     monday:
         - 星期一
+        - 周一
+        - 礼拜一
     tuesday:
         - 星期二
+        - 周二
+        - 礼拜二
     wednesday:
         - 星期三
+        - 周三
+        - 礼拜三
     thursday:
         - 星期四
+        - 周四
+        - 礼拜四
     friday:
         - 星期五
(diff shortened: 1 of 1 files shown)


asadurski:
Thank you for the new forms. Could you tell a bit more about them? Are they part of specific dialect?
Adding an example test would be very welcome here.

cassac:
They are standard Mandarin Chinese and not a dialect. There are three forms
for each day and four for Sunday, and they are all used interchangeably, so
it's quite important to include them.

If you don't speak Chinese you can verify with this online dictionary
http://ce.linedict.com/

How would I go about including a test?

asadurski:
Thank you for replying and apologies for late reply on my side. The addition looks fine to my eye.
The tests are parameterized, and the natural path would be to take a look at the examples and add your own. For days of the week, the correct place would be here:
https://github.com/scrapinghub/dateparser/blob/master/tests/test_languages.py#L26

waqasshabbir:
Hi @cassac thank you for your contribution. I resolved the conflict and merged the changes [here](https://github.com/scrapinghub/dateparser/commit/7b05e6473791936283fa4619f76345f767469cb7)

PR #129, opened 2015-11-18
Feature improved settings

waqasshabbir:
* Added interface to customize settings with addition of `settings=` keyword arg to `dateparser.parse` and `dateparser.date.DateDataParser`s constructor that accepts dict like `{'PREFER_DATES_FROM': 'past'}`
* added relevant tests
* Added registry to manager instances of `Settings` and `Language` classes to enable maximum reuse.

### DIFF
--- a/data/settings.yaml
+++ b/data/settings.yaml
@@ -2,5 +2,4 @@ settings:
     PREFER_DATES_FROM: 'current_period'
     SUPPORT_BEFORE_COMMON_ERA: False
     PREFER_DAY_OF_MONTH: 'current'
-    SKIP_TOKENS:
-        - t
+    SKIP_TOKENS: ["t"]
--- a/dateparser/conf.py
+++ b/dateparser/conf.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+import hashlib
 from pkgutil import get_data
 
 from itertools import chain
@@ -16,7 +17,7 @@ class SettingsRegistry(Registry):
             return 'default'
 
         keys= sorted(['%s-%s' % (key, str(kwargs[key])) for key in kwargs])
-        return ''.join(keys)
+        return hashlib.md5(''.join(keys)).hexdigest()
 
 
 class Settings(object):
--- a/dateparser/conf.py
+++ b/dateparser/conf.py
@@ -2,6 +2,7 @@
 from pkgutil import get_data
 
 from itertools import chain
+from functools import wraps
 from yaml import load as load_yaml
 
 """
@@ -72,3 +73,19 @@ class Settings(object):
 
 
 settings = Settings()
+
+
+def apply_settings(f):
+    @wraps(f)
+    def wrapper(*args, **kwargs):
+        if 'settings' in kwargs:
+            if isinstance(kwargs['settings'], dict):
+                kwargs['settings'] = settings.replace(**kwargs['settings'])
+            elif isinstance(kwargs['settings'], Settings):
+                kwargs['settings'] = kwargs['settings']
+            else:
+                raise TypeError("settings can only be either dict or instance of Settings class")
+        else:
+            kwargs['settings'] = settings
+        return f(*args, **kwargs)
+    return wrapper
--- a/dateparser/conf.py
+++ b/dateparser/conf.py
@@ -45,6 +45,7 @@ This only works with :mod:`DateDataParser` like below:
 class Settings(object):
 
     _attributes = []
+    _default = True
 
     def __init__(self, **kwargs):
         """
@@ -68,6 +69,7 @@ class Settings(object):
     def replace(self, **kwds):
         for x in self._attributes:
             kwds.setdefault(x, getattr(self, x))
+        kwds['_default'] = False
 
         return self.__class__(**kwds)
 
--- a/dateparser/conf.py
+++ b/dateparser/conf.py
@@ -21,7 +21,7 @@ class Settings(object):
     """
 
     _default = True
-    _yaml_data = {}
+    _yaml_data = None
 
     def __init__(self, **kwargs):
         self._updateall(
--- a/dateparser/conf.py
+++ b/dateparser/conf.py
@@ -1,9 +1,10 @@
 # -*- coding: utf-8 -*-
 import hashlib
 from pkgutil import get_data
-
 from itertools import chain
 from functools import wraps
+import six
+
 from yaml import load as load_yaml
 
 from .utils import registry
@@ -11,17 +12,21 @@ from .utils import registry
 
 @registry
 class Settings(object):
+    """Control and configure default parsing behavior of dateparser.
+    Currently, supported settings are:
+    - `PREFER_DATES_FROM`: defaults to `current_period`. Options are `future` or `past`.
+    - `SUPPORT_BEFORE_COMMON_ERA`: defaults to `False`.
+    - `PREFER_DAY_OF_MONTH`: defaults to `current`. Could be `first` and `last` day of month.
+    - `SKIP_TOKENS`: defaults to `['t']`. Can be any string.
+    """
 
-    _attributes = []
     _default = True
+    _yaml_data = {}
 
     def __init__(self, **kwargs):
-        """
-        Settings are now loaded using the data/settings.yaml file.
-        """
         self._updateall(
             chain(self._get_settings_from_yaml().items(),
-            kwargs.items())
+                  kwargs.items())
         )
 
     @classmethod
@@ -29,22 +34,24 @@ class Settings(object):
         if not args and not kwargs:
             return 'default'
 
-        keys= sorted(['%s-%s' % (key, str(kwargs[key])) for key in kwargs])
+        keys = sorted(['%s-%s' % (key, str(kwargs[key])) for key in kwargs])
         return hashlib.md5(''.join(keys)).hexdigest()
 
-    def _get_settings_from_yaml(self):
-        data = get_data('data', 'settings.yaml')
-        data = load_yaml(data)
-        return data.pop('settings', {})
+    @classmethod
+    def _get_settings_from_yaml(cls):
+        if not cls._yaml_data:
+            data = get_data('data', 'settings.yaml')
+            cls._yaml_data = load_yaml(data).pop('settings', {})
+        return cls._yaml_data
 
     def _updateall(self, iterable):
         for key, value in iterable:
-            self._attributes.append(key)
             setattr(self, key, value)
 
     def replace(self, **kwds):
-        for x in self._attributes:
+        for x in six.iterkeys(self._get_settings_from_yaml()):
             kwds.setdefault(x, getattr(self, x))
+
         kwds['_default'] = False
 
         return self.__class__(**kwds)
--- a/dateparser/conf.py
+++ b/dateparser/conf.py
@@ -6,7 +6,7 @@ from itertools import chain
 from functools import wraps
 from yaml import load as load_yaml
 
-from .utils import Registry, skip_init_if_instance_from_registry
+from .utils import Registry
 
 
 class SettingsRegistry(Registry):
@@ -25,7 +25,7 @@ class Settings(SettingsRegistry):
     _attributes = []
     _default = True
 
-    @skip_init_if_instance_from_registry
+    @Registry.skip_init_if_instance_from_registry
     def __init__(self, **kwargs):
         """
         Settings are now loaded using the data/settings.yaml file.
--- a/dateparser/conf.py
+++ b/dateparser/conf.py
@@ -30,17 +30,17 @@ class Settings(object):
         """
         Settings are now loaded using the data/settings.yaml file.
         """
-        self.updateall(
-            chain(self.get_settings_from_yaml().items(),
+        self._updateall(
+            chain(self._get_settings_from_yaml().items(),
             kwargs.items())
         )
 
-    def get_settings_from_yaml(self):
+    def _get_settings_from_yaml(self):
         data = get_data('data', 'settings.yaml')
         data = load_yaml(data)
         return data.pop('settings', {})
 
-    def updateall(self, iterable):
+    def _updateall(self, iterable):
         for key, value in iterable:
             self._attributes.append(key)
             setattr(self, key, value)
--- a/dateparser/conf.py
+++ b/dateparser/conf.py
@@ -6,27 +6,15 @@ from itertools import chain
 from functools import wraps
 from yaml import load as load_yaml
 
-from .utils import Registry
+from .utils import registry
 
 
-def _get_key(*args, **kwargs):
-    if not args and not kwargs:
-        return 'default'
-
-    keys= sorted(['%s-%s' % (key, str(kwargs[key])) for key in kwargs])
-    return hashlib.md5(''.join(keys)).hexdigest()
-
-
-_global_dict = {}
-
-
-@Registry(_global_dict, _get_key)
+@registry
 class Settings(object):
 
     _attributes = []
     _default = True
 
-    @Registry.skip_init_if_instance_from_registry
     def __init__(self, **kwargs):
         """
         Settings are now loaded using the data/settings.yaml file.
@@ -35,6 +23,14 @@ class Settings(object):
             chain(self._get_settings_from_yaml().items(),
             kwargs.items())
         )
+
+    @classmethod
+    def get_key(cls, *args, **kwargs):
+        if not args and not kwargs:
+            return 'default'
+
+        keys= sorted(['%s-%s' % (key, str(kwargs[key])) for key in kwargs])
+        return hashlib.md5(''.join(keys)).hexdigest()
 
     def _get_settings_from_yaml(self):
         data = get_data('data', 'settings.yaml')
--- a/dateparser/conf.py
+++ b/dateparser/conf.py
@@ -56,14 +56,17 @@ settings = Settings()
 def apply_settings(f):
     @wraps(f)
     def wrapper(*args, **kwargs):
-        if 'settings' in kwargs:
-            if isinstance(kwargs['settings'], dict):
-                kwargs['settings'] = settings.replace(**kwargs['settings'])
-            elif isinstance(kwargs['settings'], Settings):
-                kwargs['settings'] = kwargs['settings']
-            else:
-                raise TypeError("settings can only be either dict or instance of Settings class")
-        else:
+        kwargs['settings'] = kwargs.get('settings', settings)
+
+        if kwargs['settings'] is None:
             kwargs['settings'] = settings
+
+        if isinstance(kwargs['settings'], dict):
+            kwargs['settings'] = settings.replace(**kwargs['settings'])
+
+        if not isinstance(kwargs['settings'], Settings):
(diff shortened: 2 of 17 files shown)


data/settings.yaml
```diff
@@ -2,5 +2,4 @@ settings:
     PREFER_DATES_FROM: 'current_period'
     SUPPORT_BEFORE_COMMON_ERA: False
     PREFER_DAY_OF_MONTH: 'current'
-    SKIP_TOKENS:
-        - t
\ No newline at end of file
+    SKIP_TOKENS: ["t"]
```
asadurski: This change should not be necessary (anything that stands for it?).

dateparser/conf.py
```diff
@@ -1,67 +1,72 @@
...
         """
+        self.updateall(
+            chain(self.get_settings_from_yaml().items(),
+            kwargs.items())
+        )
 
+    def get_settings_from_yaml(self):
         data = get_data('data', 'settings.yaml')
         data = load_yaml(data)
-        settings_data = data.pop('settings', {})
```
asadurski: Just thinking... Should `get_settings_from_yaml()` also update `self._attributes`?

dateparser/date.py
```diff
@@ -12,6 +12,8 @@
 from dateparser.freshness_date_parser import freshness_date_parser
 from dateparser.languages.loader import LanguageDataLoader
 from dateparser.languages.detection import AutoDetectLanguage, ExactLanguages
+from dateparser.conf import apply_settings
+from dateparser.conf import settings as default_settings
```
asadurski: Unnecessary import.

dateparser/conf.py
waqasshabbir: `updateall` unifies settings attributes both from `settings.yaml` and `kwargs` at one place. `get_settings_from_yaml` is written more like a utility function to fetch and return settings mapping from yaml.

data/settings.yaml
waqasshabbir: Following `languages.yaml` styling, wherever we are dealing with characters which would be skipped we use *flow style sequences*. Those dealing with translation, use *block style sequences*. Made this change to be consistent with that styling.

dateparser/date.py
waqasshabbir: oh thanks! will fix it.

dateparser/conf.py
```diff
@@ -1,67 +1,72 @@
...
+            return 'default'
 
-    >>> from dateparser.conf import settings
-    >>> from dateparser import parse
-    >>> parse(u'March')
-    datetime.datetime(2015, 3, 16, 0, 0)
-    >>> settings.update('PREFER_DATES_FROM', 'future')
-    >>> parse(u'March')
-    datetime.datetime(2016, 3, 16, 0, 0)
+        keys= sorted(['%s-%s' % (key, str(kwargs[key])) for key in kwargs])
```
Allactaga: I think it would be safe to hash this string.

dateparser/conf.py
Allactaga: We have a name convention for this case:
* If method or attribute is supposed or safe to use as is on instantiated variables or classes outside of class code, we have no initial underscores (`public_method`)
* If it's not safe or some consideration need to be made, or it just internal method (like in this case) and doesn't make sense outside the class, then we use single underscore as a prefix (`_protected_method`)
* If we want to alert not only those who use our class, but also those who extends is, we use double underscore as a prefix (`__private_method`) which enables name mangling
* and of course there is also `__special_methods__` :)

Of course it's no Java encapsulation and we still can access all the methods and attributes, but with this convention we clarify our intent about scope of methods and attributes.

dateparser/conf.py
```diff
@@ -1,67 +1,72 @@
...
-
 settings = Settings()
+
+
+def apply_settings(f):
+    @wraps(f)
+    def wrapper(*args, **kwargs):
+        if 'settings' in kwargs:
+            if isinstance(kwargs['settings'], dict):
+                kwargs['settings'] = settings.replace(**kwargs['settings'])
```
Allactaga: As a nitpick, i'd write it a bit differently:
```python
if not isinstance(kwargs['settings'], (Settings, dict)):
    raise ...
if isinstance(kwargs['settings'], dict):
    ...
```

dateparser/date.py
```diff
@@ -326,14 +335,15 @@ def get_date_data(self, date_string, date_formats=None):
...
         else:
             return {'date_obj': None, 'period': 'day'}
 
     @classmethod
-    def _get_language_loader(cls):
-        if not cls.language_loader:
-            cls.language_loader = LanguageDataLoader()
-        return cls.language_loader
+    def _get_language_loader(cls, settings):
+        if any([not cls.language_loaders, settings not in cls.language_loaders]):
```
Allactaga: Since `language_loaders` is `{}` by default, I think we can just use second condition here :)

dateparser/date_parser.py
```diff
@@ -25,7 +25,8 @@ class new_parser(parser.parser):
     For more see issue #36
     """
 
-    def parse(self, timestr, default=None, ignoretz=False, **kwargs):
+    @apply_settings
```
Allactaga: Do we expect someone to call this method directly?

dateparser/utils.py
```diff
@@ -86,3 +86,24 @@ def find_date_separator(format):
     m = re.search(r'(?:(?:%[dbBmaA])(\W))+', format)
     if m:
         return m.group(1)
+
+
+class Registry(type):
```
Allactaga: It's not very encouraged to use metaclasses on libraries as we can never know what our users can do with it. They may want to subclass it with their own metaclass.

dateparser/conf.py
waqasshabbir: Thanks @Allactaga! much help. Will make necessary changes.

dateparser/conf.py
waqasshabbir: Just curious, how hashing here would be of value for us? Given that `dict` automatically hashes the keys.

Allactaga: Just for not storing long lines. No other reason. :)

dateparser/utils.py
waqasshabbir: @Allactaga I tried to achieve the same using factory pattern and decorators but this looks like a lot cleaner solution. Can you please suggest the alternate implementation?

On a side note, why do you think subclassing it with one's own metaclass will be tempting?

Allactaga: One of alternate implementations would be subclassing and using private attribute in `__new__`. (don't forget here, that `__init__` would still be called).

As for a side note, why do you thing there is **for sure** would be no such temptation? Deprive users of ways to write their own code (however bad it is) is not the best approach in my opinion.

waqasshabbir: Thank you for the tip! made the changes accordingly. Please let me know how you like the changes.

Regarding users wanting to use their own metaclasses, I am just curious why would they want to do this, of course, we can't be sure of their wants/needs. But provided the scope of our library, we expose certain methods and classes to users and quite often make assumptions that a component won't be required to be used individually. Likewise, our guarantee is only limited and do not cover not-prescribed usage of unexposed/internal components of the library.

dateparser/utils.py
```diff
@@ -86,3 +87,37 @@ def find_date_separator(format):
...
+
+    @classmethod
+    def get_key(cls, *args, **kwargs):
+        raise NotImplemented
+
+    @classmethod
+    def skip_init_if_instance_from_registry(cls, f):
+        @wraps(f)
+        def wrapper(self, *args, **kwargs):
+            if not getattr(self, cls._tag, False):
```
Allactaga: `bool('from_registry')` ;)

dateparser/utils.py
```diff
@@ -86,3 +87,37 @@ def find_date_separator(format):
...
+    def __new__(cls, *args, **kwargs):
+        key = cls.get_key(*args, **kwargs)
+
+        if key in cls._global_dict:
+            return cls._tag_instance(cls._global_dict[key])
+
+        return cls._global_dict.setdefault(
+            key,
+            object.__new__(cls, *args, **kwargs)
+        )
```
Allactaga: Why not use structure like
```python
if key not in cls._global_dict:
    cls._global_dict[key] = ...
return cls._global_dict[key]
``` ?

File: dateparser/utils.py [hunk 11]
```diff
@@ -86,3 +87,37 @@ def find_date_separator(format):
     m = re.search(r'(?:(?:%[dbBmaA])(\W))+', format)
     if m:
         return m.group(1)
+
+
+class Registry(object):
+
+    _global_dict = {}
```
Allactaga: This attribute would be shared across all child classes.

dateparser/utils.py
waqasshabbir: whenever we return an instance from _global_dict, we tag it with `from_register` attribute so that `__init__` doesn't get applied to it, The suggested structure would interfere with this condition.

dateparser/utils.py
waqasshabbir: Do you mean `hasattr(self, cls._tag)`?

Allactaga: No, I misunderstood the code. Just ignore my comment here :)

dateparser/utils.py
Allactaga: Wouldn't it be more convenient to tag them after method run, not on getting them from dictionary?

dateparser/utils.py
```diff
@@ -86,3 +87,37 @@ def find_date_separator(format):
...
+    def _tag_instance(cls, instance):
+        setattr(instance, cls._tag, True)
+        return instance
+
+    @classmethod
+    def get_key(cls, *args, **kwargs):
+        raise NotImplemented
+
+    @classmethod
+    def skip_init_if_instance_from_registry(cls, f):
```
Allactaga: I'm starting to think if instead of base class it would be more convenient to use class decorator :)

dateparser/utils.py
waqasshabbir: We want to be able to identify when an instance is returned from dictionary and when it is created fresh.  In case,  instance comes from dictionary we prevent `__init__` from being called on it. We manage this by tagging instances.  If we'd tag the instance after it is created and not at the time of returning from dictionary,  it would always appear to be coming  from dictionary and hence `__init__` won't be applied to it.

Allactaga: ... after method (in this case `__init__`) passed to `skip_init_if_instance_from_registry` run ...

waqasshabbir: it's `skip_init_if_instance_from_registry` responsibility to *identify* whether an instance is from registry or a fresh instance. It does so by checking for tag (.i.e. attribute `from_registry`'s presence). That's why, tagging doesn't seem to be a good fit here.

Allactaga: If you want to, you can go with this approach, but I am still not convinced about its flow being clear :)

I mentioned class decorator earlier, so let me share this approach with you for you to compare. Word of warning though: I didn't spent much time on naming, so this is only proof of concept.
```python
def registry(cls):
    def choose(creator):
        def constructor(cls, *args, **kwargs):
            key = cls.get_key(*args, **kwargs)

            if not hasattr(cls, "__registry_dict"):
                setattr(cls, "__registry_dict", {})
            registry_dict = getattr(cls, "__registry_dict")

            if key not in registry_dict:
                registry_dict[key] = creator(cls, *args, **kwargs)
            return registry_dict[key]
        return staticmethod(constructor)

    def self_destruct(initializer):
        def destructor(self, *args, **kwargs):
            if getattr(self, "__initialized", False):
                return
            else:
                initializer(self, *args, **kwargs)
                setattr(self, "__initialized", True)
        return destructor

    if not (hasattr(cls, "get_key")
            and isinstance(cls.get_key, MethodType)
            and cls.get_key.__self__ is cls):
        raise NotImplementedError("Registry classes require to implement class method get_key")

    setattr(cls, '__new__', choose(cls.__new__))
    setattr(cls, '__init__', self_destruct(cls.__init__.__func__))
    return cls
```

dateparser/utils.py
waqasshabbir: Sure, take a look again please.

dateparser/utils.py
waqasshabbir: Thanks Eugene! applied your changes. Your naming conventions are on fleek! :+1:

dateparser/languages/language.py
```diff
@@ -1,29 +1,46 @@
...
 class Language(object):
+
     _dictionary = None
     _splitters = None
     _wordchars = None
 
-    def __init__(self, shortname, language_info):
+    @classmethod
+    def get_key(cls, *args, **kwargs):
+        key_template = '%(shortname)s-%(skip_tokens)s'
```
Allactaga: I think we either need to describe how to build this method in docs, or better yet, to find a way to refactor this, so `Language` class was not registry itself, but better yet pass settings object to `translate` to `_join` to `_get_splitters` and store cache of splitters in language instead. What do you think.

dateparser/utils.py
```diff
@@ -86,3 +87,36 @@ def find_date_separator(format):
...
+    def choose(creator):
+        def constructor(cls, *args, **kwargs):
+            key = cls.get_key(*args, **kwargs)
+
+            if not hasattr(cls, "__registry_dict"):
+                setattr(cls, "__registry_dict", {})
+            registry_dict = getattr(cls, "__registry_dict")
+
+            if key not in registry_dict:
+                registry_dict[key] = creator(cls, *args, **kwargs)
```
Allactaga: Consider having `setattr(registry_dict[key], 'registry_key', key)` here, in case you need this information for some refactoring.

waqasshabbir: Thanks! done.

dateparser/languages/language.py
waqasshabbir: As you suggested, with new changes, `Language` class isn't a registry - `translate`/ `is_applicable` are called with `settings` instance, which we attach to `Dictionary` instance. Now we cache `sorted_words` (a list of words) and `split_regex` (complied regexes) only in `Dictionary` class for subsequent reuse.

dateparser/conf.py
waqasshabbir: Thanks! done.

dateparser/conf.py
```diff
@@ -1,67 +1,72 @@
...
+
+    @classmethod
+    def get_key(cls, *args, **kwargs):
+        if not args and not kwargs:
+            return 'default'
+
+        keys= sorted(['%s-%s' % (key, str(kwargs[key])) for key in kwargs])
+        return hashlib.md5(''.join(keys)).hexdigest()
 
+    def _get_settings_from_yaml(self):
```
Allactaga: Should we cache this, so we don't access file on each replace?

dateparser/conf.py
```diff
@@ -1,67 +1,72 @@
...
-
-    >>> settings.update('SKIP_TOKENS', ['de'])  # Turkish word for 'at'
-    >>> from dateparser.date import DateDataParser
-    >>> DateDataParser().get_date_data(u'27 Haziran 1981 de')  # Turkish (at 27 June 1981)
-    {'date_obj': datetime.datetime(1981, 6, 27, 0, 0), 'period': 'day'}
-"""
+@registry
+class Settings(object):
 
+    _attributes = []
```
Allactaga: Would be sharing one list across all class instances.

dateparser/conf.py
```diff
@@ -1,67 +1,72 @@
...
         data = get_data('data', 'settings.yaml')
         data = load_yaml(data)
-        settings_data = data.pop('settings', {})
+        return data.pop('settings', {})
 
-        for datum in settings_data:
-            setattr(self, datum, settings_data[datum])
+    def _updateall(self, iterable):
+        for key, value in iterable:
+            self._attributes.append(key)
```
Allactaga: Wouldn't this create duplicates in some cases?

waqasshabbir: Cached yaml data and removed altogether `_attributes` attrib. Thanks!

dateparser/conf.py
```diff
@@ -1,67 +1,79 @@
...
-    {'date_obj': datetime.datetime(1981, 6, 27, 0, 0), 'period': 'day'}
-"""
+        keys = sorted(['%s-%s' % (key, str(kwargs[key])) for key in kwargs])
+        return hashlib.md5(''.join(keys)).hexdigest()
 
+    @classmethod
+    def _get_settings_from_yaml(cls):
+        if not cls._yaml_data:
+            data = get_data('data', 'settings.yaml')
+            cls._yaml_data = load_yaml(data).pop('settings', {})
```
Allactaga: Just a **nitpick**: as initial `_yaml_data` equals `{}`, I'd expect `update` here. Or, another way to look at it, as we have assignment here, I'd expect `_yaml_data` be not initialized (`None`) :)

dateparser/date.py
```diff
@@ -94,7 +95,7 @@ def sanitize_date(date_string):
     date_string = re.sub(r'\b([ap])(\.)?m(\.)?\b', r'\1m', date_string, flags=re.DOTALL | re.I)
     date_string = re.sub(r'^.*?on:\s+(.*)', r'\1', date_string)
 
-    date_string = re.sub(ur'|'.join(APOSTROPHE_LOOK_ALIKE_CHARS), u"'", date_string)
+    date_string = re.sub(r'|'.join(APOSTROPHE_LOOK_ALIKE_CHARS), u"'", date_string)
```
Allactaga: Wouldn't it be converted back to unicode anyway?

dateparser/date.py
```diff
@@ -154,17 +156,17 @@ def __init__(self, language, date_string, date_formats):
...
-            self._try_timestamp,
-            self._try_freshness_parser,
-            self._try_given_formats,
-            self._try_dateutil_parser,
-            self._try_hardcoded_formats,
+                self._try_timestamp,
+                self._try_freshness_parser,
+                self._try_given_formats,
+                self._try_dateutil_parser,
+                self._try_hardcoded_formats,
```
Allactaga: Why making indent 8 spaces? `):` placed the way it is clearly separates condition part from body.

dateparser/languages/dictionary.py
```diff
@@ -70,22 +77,36 @@ def split(self, string, keep_formatting):
...
+        return self._sorted_words_cache[self._settings][self.info['name']]
+
+    def _get_split_regex_cache(self):
+        if (
+                self._settings not in self._split_regex_cache or
+                self.info['name'] not in self._split_regex_cache[self._settings]
+            ):
             self._construct_split_regex()
-        return self._split_regex
+        return self._split_regex_cache[self._settings][self.info['name']]
```
Allactaga: What if we use `self._settings.registry_key` here? :)

dateparser/date.py
waqasshabbir: Python 3 has dropped support for `ur` prefix. That's why changed it to raw string. But since, there aren't any back slashes, I'd change the prefix to `u`.i.e. unicode.

dateparser/date.py
waqasshabbir: Because the pylint said so :D. Reverted the change. Thanks!

Issue #100, opened 2015-07-29
parsed time zone different for fixed and relative dates


user500:
It appears that `dateparser.parse` converts a fixed date and time to the local time zone and a relative date and time to UTC? I was wondering if it might be possible to add an option to make them both return the same, or to set the time zone in the returned datetime object? I would like to convert the parsed date and time to a Unix time stamp, but I don't know whether the user will enter a fixed or relative date and time. This is with version 0.2.1 on Ubuntu 14.04.2 LTS. Thank you!

user500:
Thinking about it some more, it might be nice if the relative date and times continued to be parsed to UTC and there was an option to have the fixed date and times remain in the time zone parsed from the string (if any, UTC otherwise). In both cases would it be possible to have the time zone information stored in the returned datetime object?

waqasshabbir:
Hi @user500! thank you for taking out time and report this. We are planning to introduce configurations in release 0.5 which should also address this issue. Thanks once again!

waqasshabbir:
Hi @user500, we've introduced `TIMEZONE` setting to dateparser. You can use it as follows. By default, all dates are considered in `UTC`. If timezone setting is specified, dates are converted to the given timezone. 

```
# relative date
In [10]: dateparser.parse('2 hours ago')  # UTC
Out[10]: datetime.datetime(2016, 1, 20, 7, 54, 40, 195695)

In [9]: dateparser.parse('2 hours ago', settings={'TIMEZONE': 'US/Eastern'})
Out[9]: datetime.datetime(2016, 1, 20, 2, 54, 26, 450596)
```

```
# full date
In [17]: dateparser.parse('21 december 2015, 22:10') # assumes UTC
Out[17]: datetime.datetime(2015, 12, 21, 22, 10)

In [18]: dateparser.parse('21 december 2015, 22:10', settings={'TIMEZONE': 'Asia/Karachi'})
Out[18]: datetime.datetime(2015, 12, 22, 3, 10)
```

waqasshabbir:
Fixed in PR #132

PR #132, opened 2015-12-06
Feature timezone freshness parser

waqasshabbir:
Proposed fix for #100.

While we assume no timezone for complete dates unless timezone is specified in the given date string, for relative dates, we used to return dates in UTC time. This PR adds flexibility to convert that time to desirable timezone using `settings`. 

```
>>> from dateparser import parse
>>> parse('2 hours ago')
datetime.datetime(2015, 12, 6, 21, 39, 21, 720162)
>>> parse('2 hours ago', settings={'TIMEZONE': 'US/Eastern'})  # five hours behind UTC.
datetime.datetime(2015, 12, 6, 16, 39, 21, 720162)
```


tests/test_date_parser.py
```diff
@@ -304,11 +304,11 @@ def test_dates_parsing(self, date_string, expected):
         self.then_date_obj_exactly_is(expected)
 
     @parameterized.expand([
-        param('Sep 03 2014 | 4:32 pm EDT', datetime(2014, 9, 3, 21, 32)),
-        param('17th October, 2034 @ 01:08 am PDT', datetime(2034, 10, 17, 9, 8)),
-        param('15 May 2004 23:24 EDT', datetime(2004, 5, 16, 4, 24)),
+        param('Sep 03 2014 | 4:32 pm EDT', datetime(2014, 9, 3, 20, 32)),
```
asadurski: Wait... Those tests were wrong?

waqasshabbir: Yes @asadurski, `EDT` is 4 hours behind `UTC` because of day light savings instead of 5 hours (EST). The same applies to `PDT` case. Prior to [this change](https://github.com/scrapinghub/dateparser/pull/132/files#diff-620d77b1f1bb42b71cf2181c8b7ce193R183), these tests would pass successfully. 

Also note, EDT (-0400), instead of EST (-0500), is usually observed from last Sunday of April to Last Sunday of October.

asadurski: Right. It made sense after I've written it out. Thanks!

dateparser/utils.py
```diff
@@ -99,25 +130,15 @@ def constructor(cls, *args, **kwargs):
...
-    def self_destruct(initializer):
-        def destructor(self, *args, **kwargs):
-            if getattr(self, "__initialized", False):
-                return
-            else:
-                initializer(self, *args, **kwargs)
-                setattr(self, "__initialized", True)
-        return destructor
-
     if not (hasattr(cls, "get_key")
```
waqasshabbir: this wasn't needed once `Language` classes stopped to be returning from registry. Hence, removed.

dateparser/utils.py
```diff
@@ -99,25 +130,15 @@ def constructor(cls, *args, **kwargs):
             registry_dict = getattr(cls, "__registry_dict")
 
             if key not in registry_dict:
-                registry_dict[key] = creator(cls, *args, **kwargs)
+                registry_dict[key] = creator(cls, *args)
```
waqasshabbir: this was necessitated by [this change in Python 3](https://hg.python.org/cpython/file/44ed0cd3dc6d/Objects/typeobject.c#l2818) for excessive parameters.

asadurski:
Looks good!

waqasshabbir:
Hi @domenkozar, yes it does.

domenkozar:
Could we get a release with the fix? :)

waqasshabbir:
Sure, we've planned a release for it early coming week. Thanks!

PR #133, opened 2015-12-08
Added relative base date to count relative date from exact date

merito:
Hi,

It will be very useful to have an option to specify a date to count relative dates from it. It is my simple implementation to show you my idea. I'll be glad to see this feature.

Dawid


merito:
You've changed the way of testing, so now my changes work with the code from the master. Previous method of testing was definitely bad. I've added some tests for my feature. Please look at this and decide if it is good. This feature is very useful in my applications, I think it can be useful for the others

merito:
Make a new PR without this mess.

waqasshabbir:
Please make it configurable through settings instead of changing `DateDataParser.get_date_data()` or `parser()` signatures.

PR #138, opened 2016-01-21
Fix date formats for translated strings

waqasshabbir:
Earlier, it was a bit misleading how commas in date format would keep it from applying to the supplied date string. It was because `,` would get skipped by dateparser when translating string. This PR fixes that.

Secondly, translated date string is normalized for spaces.

Lastly, it improves tests for checking with date formats in the scope of `dateparser.parse` and `DateDataParser.get_date_data`. Prior to this change, `parse_with_formats` function was being tested by supplying actual date string.

```
In [1]: from dateparser import parse

In [2]: parse('10.01.2016, 20:35')
Out[2]: datetime.datetime(2016, 10, 1, 20, 35)

In [3]: parse('10.01.2016, 20:35', date_formats=['%d.%m.%Y, %H:%M'])
Out[3]: datetime.datetime(2016, 1, 10, 20, 35)
```


waqasshabbir:
Replaced by #194

Issue #136, opened 2016-01-19
unsupport chinese popular format : 2015年04月08日10:05


conanca:
It return None, but should be : datetime.datetime(2015, 4, 8, 10, 5)

waqasshabbir:
Fixed by PR #142

PR #142, opened 2016-02-01
Fixed issue with chinese dates with component numbers

waqasshabbir:
Fixes issue #136.


asadurski:
This is really cool. I thought that additional changes would be required to support this format and your solution is very simple (and the same format is also used in Japan, so that's cool for both languages). One thing to notice is that we need to include format with spaces in the regex (which is not popular, but it appears sometimes. Example: http://www.boj.or.jp/). Otherwise we get `ValueError: year is out of range` for such dates.

waqasshabbir:
Thanks @artur, done.

PR #145, opened 2016-02-15
Feature tz aware return type

### LINKED ISSUE #143: Returning ambiguous naive datetimes when tz info is available
Originally this was a bit more confusing because dateparser returned both a naive datetime object while also converting it using the local timezone offset. This was updated in 3.2, but I would still make the case that it makes more sense to return the datetime object with timezone data attached, when possible.

See this writeup that @thebigspoon put together: 

https://gist.github.com/thebigspoon/a5e38d029577fe9ddb94

waqasshabbir:
Fixes #143.

Example:

```
In [2]: parse('12 Feb 2015 10:56 PM EST', settings={'RETURN_AS_TIMEZONE_AWARE': True, 'TIMEZONE': 'EST'})
Out[2]: datetime.datetime(2015, 2, 12, 22, 56, tzinfo=<StaticTzInfo 'EST'>)

# if no `TIMEZONE` isn't specified.
In [3]: parse('12 Feb 2015 10:56 PM EST', settings={'RETURN_AS_TIMEZONE_AWARE': True})
Out[3]: datetime.datetime(2015, 2, 13, 3, 56, tzinfo=<StaticTzInfo 'UTC'>)
```

### DIFF
--- a/data/settings.yaml
+++ b/data/settings.yaml
@@ -4,3 +4,4 @@ settings:
     PREFER_DAY_OF_MONTH: 'current'
     SKIP_TOKENS: ["t"]
     TIMEZONE: 'UTC'
+    RETURN_AS_TIMEZONE_AWARE: False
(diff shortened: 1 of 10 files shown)


akoumjian:
Yes, this appears to address our concerns. Thank you.

thebigspoon:
works for me. thanks @waqasshabbir

Issue #147, opened 2016-02-16
Can't parse valid date


angru:
``` python
parse = lambda date: dateparser.parse(date, languages=['ru'], date_formats=['%d %b %Y %H:%M'])

print(parse('13 янв. 2005 19:13'))
print(parse('13 фев. 2005 19:13'))
print(parse('13 мар. 2005 19:13'))
print(parse('13 апр. 2005 19:13'))
print(parse('13 май. 2005 19:13'))
print(parse('13 июн. 2005 19:13'))
print(parse('13 июл. 2005 19:13'))
print(parse('13 авг. 2005 19:13'))
print(parse('13 сен. 2005 19:13'))
print(parse('13 окт. 2005 19:13'))
print(parse('13 ноя. 2005 19:13'))
print(parse('13 дек. 2005 19:13'))
```

> 2005-01-13 19:13:00
> 2005-02-13 19:13:00
> 2005-03-13 19:13:00
> 2005-04-13 19:13:00
> 2005-05-13 19:13:00
> 2005-06-13 19:13:00
> 2005-07-13 19:13:00
> None
> 2005-09-13 19:13:00
> 2005-10-13 19:13:00
> 2005-11-13 19:13:00
> 2005-12-13 19:13:00

lubuntu 14.04 x64 RU Python 3.5.1

PR #153, opened 2016-03-04
Fix sanitize Russian year

skrypka:
Hi,
I've fixed issue: https://github.com/scrapinghub/dateparser/issues/147
Problem was that sanitize_date remove 'г.' even if it's part of word like 'Авг.'


waqasshabbir:
Thanks @skrypka!

PR #156, opened 2016-03-16
Handle change in return type of private method in dateutil 2.5.0

### LINKED ISSUE #154: python-dateutil Upgrade to >2.5 breaks dateparser.parse
My team was recently successfully using dateparser.parse on our project.  Once the python-dateutil package upgraded from 2.4.2 to 2.5.0  dateparser.parse no longer worked.  We received the following error:
AttributeError: 'tuple' object has no attribute 'year'

To reproduce the error pin python-dateutil to 2.5.0 in your environment and run dateparser.parse with a reasonable datatime string.

We're pinning python-dateutil to 2.4.2 but thought you should know.

SpotlightKid:
Fixes #154.

The "fix" in #154 was really just a workaround and didn't address the cause of the problem: `dateparser` uses the private `_parse()` method of `dateutil.parser.parser`. This method changed its return type in `python-dateuti` 2.5.0 to a tuple. This PR handles this in a backwards-compatible way and removes the version restriction for python-dateutil in `setup.py` and `requirements.txt` again.

### DIFF
--- a/dateparser/date_parser.py
+++ b/dateparser/date_parser.py
@@ -33,6 +33,10 @@ class new_parser(parser.parser):
         # Parse timestr
         res = self._parse(timestr, **kwargs)
 
+        # python-dateutil >= 2.50 changed return type of parser._parse()
+        if isinstance(res, tuple):
+            res = res[0]
+
         if res is None:
             raise ValueError("unknown string format")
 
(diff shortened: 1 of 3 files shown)


waqasshabbir:
fixed by #165

PR #158, opened 2016-03-25
Add support for Japanese language

buhii:
Added support for Japanese language.


data/languages.yaml
```diff
@@ -2298,3 +2298,110 @@ fi:
     simplifications:
         - eilen: 1 päivä
         - tänään: 0 päivä
+
+
+ja:
```
waqasshabbir: two letter iso code for japan is `jp`.

waqasshabbir:
Thanks @buhii, great work! merging already.

buhii:
Thanks too, @waqasshabbir.
I'm sorry but I noticed that two iso code for Japanese is complicated.
Two letter iso country code (ISO 3166-1 alpha-2) for Japan is `jp`, but two letter language code for Japanese (ISO 639-1) is actually `ja`.
Shoud I revert the last commit?

waqasshabbir:
Good point! I had an impression that we've been using *country* codes and
not *language* codes. But I noticed `fa` being used for Persian language
instead of `ir` in languages.yaml. No need to revert it,  we will look into
it and adopt one clear standard and make relevant changes accordingly in a
separate PR soon. Thanks!

buhii:
All right, I understand. Thanks too!

zxq9:
This issue sneaks up everywhere i18n/l10n is confronted head-on. /(O.o)\

I recommend we stick with the current trend, which is to use the language, not country code:
https://github.com/scrapinghub/dateparser/blob/master/data/languages.yaml#L5

As an example, it would be a little weird to have a separate setting for Canada than the US, and within Canada represent Frence some alien way, etc.

waqasshabbir:
Yes,  Craig. Of course,  that's the way we gonna go.  I plan to run a
thorough check for the two letter codes mistakes if any tomorrow and make
sure that iso 639-1 is followed everywhere in our data file. Appreciate
your input,  thanks!

waqasshabbir:
Hi @buhii, @zxq9 fyi I did some cleaning [here](https://github.com/scrapinghub/dateparser/pull/160). Thank you!

Issue #155, opened 2016-03-08
Accents in French: how to ignore them?


YAmikep:
There should be a way to ignore accents.

For example, in french, the parser returns None when an accent is missing.
```
# No accent on the e
>>> dateparser.parse('04-decembre-2015')

# Accent on the e
>>> dateparser.parse('04-décembre-2015')
datetime.datetime(2015, 12, 4, 0, 0)
```

waqasshabbir:
Fixed by PR #159

PR #159, opened 2016-03-29
Diacritic character normalization

ammar-azif:
This PR is related to https://github.com/scrapinghub/dateparser/issues/155

Given date of missing diacritical marks, such as

```04-decembre-2015 # supposed to be 04-décembre-2015```

It can be parsed by setting normalize flag to true

```
>>> dateparser.parse(u'04-décembre-2015')
datetime.datetime(2015, 12, 4, 0, 0)
>>> dateparser.parse(u'04-decembre-2015')
>>> dateparser.parse(u'04-decembre-2015', settings={'NORMALIZE': True})
datetime.datetime(2015, 12, 4, 0, 0)
```

### CODE REVIEW

**Comment by waqasshabbir** [comment]
Great work! @ammar-azif.

I'd like to suggest a couple of improvements.

Instead of `normalize` argument, let's add a configuration to the `setting` and enables it by default. If a user wants the opposite behavior they can configure it using `settings=`.

Also, let's add more tests covering date strings with stripped accents words which in languages.yaml have accents.

**Comment by ammar-azif** [comment]
@waqasshabbir : I have made the changes based on your feedback. Please review again thanks.

File: dateparser/languages/dictionary.py [hunk 1]
```diff
@@ -19,6 +21,7 @@ class UnknownTokenError(Exception):
 class Dictionary(object):
     _split_regex_cache = {}
     _sorted_words_cache = {}
+    _denormalized_words_cache = {}
```
waqasshabbir: Where is this being used?

data/languages.yaml
```diff
@@ -1,7 +1,6 @@
 base:
     skip: [" ", ".", ",", ";", "-", "/", "'", "|", "@", "[", "]", "，"]
 
-
```
waqasshabbir: Can we restore this line? we are using two lines to separate blocks.

data/settings.yaml
```diff
@@ -5,3 +5,4 @@ settings:
     SKIP_TOKENS: ["t"]
     TIMEZONE: 'UTC'
     RETURN_AS_TIMEZONE_AWARE: False
+    NORMALIZE: False
```
waqasshabbir: Let's set it to `True`.

dateparser/languages/language.py
```diff
@@ -164,6 +168,43 @@ def _set_wordchars(self, settings=None):
...
+            key = normalize_unicode(key)
+            if isinstance(value, int):
+                c_simplification[key] = str(value)
+            else:
+                c_simplification[key] = normalize_unicode(value)
+            self._normalized_simplifications.append(c_simplification)
+
+    def _generate_simplifications(self):
+        self._simplifications = []
+        for simplification in self.info.get('simplifications', []):
```
waqasshabbir: we can enclose lines 186-193 in a separate function and add some sort of flag to normalize values. And reuse that function in both places.

tests/__init__.py
```diff
@@ -8,6 +9,7 @@ def setUp(self):
         self.__patches = []
 
         self.error = NotImplemented
+        settings.NORMALIZE = False
```
waqasshabbir: why do we need this?

dateparser/languages/dictionary.py
ammar-azif: Removed.

data/languages.yaml
ammar-azif: Done.

data/settings.yaml
ammar-azif: Done.

dateparser/languages/language.py
ammar-azif: Done. Thanks.

tests/__init__.py
ammar-azif: Initially I thought settings normalize flag should be set to False. Reverted the change. Thanks.

waqasshabbir:
Thanks @ammar-azif for the changes, looks good!

PR #161, opened 2016-04-01
adding features for more use-cases to date_parser

demelziraptor:
This PR is not ready yet but I want to get opinions first before finishing it.

### Specific issues I'm attempting to solve with this PR:
- Returning datetime with auto-detected timezone info; currently you have to manually enter the timezone as the TIMEZONE setting
- Returning datetime with no timezone info if none is detected in the string, else timezone info if it is detected; currently you either have timezone info or not depending on the code and not the date string
- Use the auto-detected timezone, or a default one if one is not detected

### Examples:

**String that I know is EST and I want a datetime with tz of EST**
Current code - possible

    DateDataParser(settings={'TIMEZONE': 'EST', 'RETURN_AS_TIMEZONE_AWARE': True}).get_date_data(u'23 March 2000, 1:21 PM')
    datetime.datetime(2000, 3, 23, 13, 21, tzinfo=<StaticTzInfo 'EST'>)

**String that I know is EST and I want a datetime converted to UTC**
Current code & PR code - only possible in hacky way

    DateDataParser(settings={'RETURN_AS_TIMEZONE_AWARE': True}).get_date_data(u'23 March 2000, 1:21 PM' + ' EST')
    datetime.datetime(2000, 3, 23, 18, 21, tzinfo=<StaticTzInfo 'UTC'>)

**String that I don't know is EST and I want a datetime converted to UTC**
Current code - possible

    DateDataParser(settings={'RETURN_AS_TIMEZONE_AWARE': True}).get_date_data(u'23 March 2000, 1:21 PM EST')
    datetime.datetime(2000, 3, 23, 18, 21, tzinfo=<StaticTzInfo 'UTC'>)

**String that I don't know is EST and I want a datetime converted to PST**
Current code - possible

    DateDataParser(settings={'TIMEZONE': 'PST', 'RETURN_AS_TIMEZONE_AWARE': True}).get_date_data(u'23 March 2000, 1:21 PM EST')
    datetime.datetime(2000, 3, 23, 10, 21, tzinfo=<StaticTzInfo 'PST'>)

**String that I don't know is EST and I want a datetime with tz of EST**
Current code - not possible

    DateDataParser(settings={'RETURN_AS_TIMEZONE_AWARE': True}).get_date_data(u'23 March 2000, 1:21 PM EST')
    datetime.datetime(2000, 3, 23, 18, 21, tzinfo=<StaticTzInfo 'UTC'>)
PR code - possible

[...]

PR #164, opened 2016-04-11
Allow client code to determine whether a timezone was explicitly specified

thomasst:
If ``RETURN_AS_TIMEZONE_AWARE`` is set to ``True``, and ``TIMEZONE`` is ``None``, a timezone aware date is only returned if the input string explicitly specifies the time zone, otherwise an unaware date is returned.

This change allows client code to determine whether or not a timezone was explicitly specified in the date string.

### DIFF
--- a/dateparser/date_parser.py
+++ b/dateparser/date_parser.py
@@ -182,7 +182,8 @@ class DateParser(object):
         if tz is not None:
             date_obj = tz.localize(date_obj)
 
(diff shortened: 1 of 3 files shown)


waqasshabbir:
Feel free to add to the discussion here: #161, thanks @thomasst!

waqasshabbir:
We will have to make relevant changes [here](https://github.com/scrapinghub/dateparser/blob/master/dateparser/freshness_date_parser.py#L57) as well before we can merge this.

PR #185, opened 2016-06-07
Add support Arabic Numeral

mrgrigorii:


mrgrigorii:
Added support Arabic Numeral

PR #186, opened 2016-06-07
Added support Greek, added "May - Toukokuuta" in fi.yaml

mrgrigorii:


waqasshabbir:
@mrgrigorii would you like to re-open this request? It seems fine though you need to add tests. Look into `tests/test_date_parser.py` and `tests/test_freshness_parser.py` files.

Issue #207, opened 2016-07-15
YMD dateformat is not functioning


solalatus:
## What I have tried?
I have tried to add language support for **Hungarian** - you can look at the result in:
https://github.com/solalatus/dateparser/commit/800f6ce76337bc42ae9126b9c89c4bffe644ac32
and
https://github.com/solalatus/dateparser/commit/2bc56830efd843788208b42259048f0aeff119e2

After adding the files, generally things functioned well, but failed in some cases. 

## Functioning case:
"2016 július 3."

## Non-functioning case:
"2016 július 15."

## Details
Dots are ommitted well, the only factor that breaks the code is when the day gets to double digits.
(Tested also above and below 12 to rule out confusion with months, but that is not the case. 9 works, 10 no longer...)

## Hope
If you have time to check it, maybe Hungarian language support can be added.

(Disclaimer: as an agglutinative language Hungarian is tricky, there are cases where parsing will fail, since enumerating all the suffixes is nearly impossible - they can be combined quite freely. Despite this, decent coverage is possible.)

# System:
Ubuntu 16.04, Anaconda 4.1.8, Python 3.5

waqasshabbir:
@solalatus this looks related to noramlization, it will be very convenient for me to debug the issue if you could create a PR. Can you do that?

solalatus:
@waqasshabbir pull request done. Please inform, if I can contribute in any other way!
Thx!

waqasshabbir:
Hi @solalatus, I looked into it today and found out the issue is related to new parser. Your changes should work fine now. Regarding the PR, can you please also add tests? Except for tests, PR looks great. Thanks.

solalatus:
I have added some test, but on my side it is still failing. I did not pull any changes from the master, so it might be all ok by merge. Could you please take a look? Thx

waqasshabbir:
Hi @solalatus, thank you for the contribution! I fixed the tests and merged the PR.

solalatus:
Thanks, it was a pleasure! :-)

PR #211, opened 2016-07-21
Adding Hungarian support

solalatus:
Created necessary files for Hungarian.
Coverage looks ok, but since Hungarian is a funny agglutinative language, some cases might be missig.

Creates some problems described here: https://github.com/scrapinghub/dateparser/issues/207
Merge with care!

Thx


waqasshabbir:
Replaced by #231

PR #214, opened 2016-07-23
tweak for timezone

brechmos:
The default was to assume the input date_time was UTC but the `tz_string` appears to hold the information.  So I created a timezone object based on the `tz_string`.  

One thing I did not check for was an empty `tz_string` as I did not follow it back through. So a little more work might have to be done.


waqasshabbir:
declining in favor of #235

Issue #230, opened 2016-09-07
default time zones of input data


kmike:
Hey! Why does dateparser convert date *to* timezone specified in settings, not *from* this timezone? 

For example, let's say there is a forum with UTC+5 timezone, so all dates are in UTC+5, and I want to have dates in UTC. Passing TIMEZONE='UTC+5' won't help - dateparser will assume dates are in UTC, and then convert them to UTC+5, i.e. it'll add five hours instead of subtracting five hours.

Once you have a date in a known timezone it is easy to convert it to any other timezone using other libraries, you don't need dateparser support for this. So currently 'TIMEZONE' option is merely a helper. 

The problem is that it is tricky to handle relative dates correctly with dateparser now. Let's say we're on that forum again, with UTC+5 dates. Old dates are written in full, e.g. "2016-09-07 12:30", recent dates are relative, e.g. "1 hour ago".

Without timezone information dateparser will think both dates are in UTC. It means it'll use utcnow() as a current time when computing a full date for "1 hour ago", which is not what they meant on a website.

Let's say current time (as used by website) is 2016-09-07 15:00 UTC+5, or 2016-09-07 10:00 UTC.

text  | date in UTC+5 | date in UTC | parse result
-----|---------------|------------- | ------------
1 hour ago | 2016-09-07 14:00 | 2016-09-07 09:00 | 2016-09-07 09:00 UTC
2016-09-07 12:30 | 2016-09-07 12:30 | 2016-09-07 07:30 | 2016-09-07 12:30 UTC

As you see, 'parse result' is incorrect. TIMEZONE option doesn't help here because both times are adjusted by the same amount.

`RELATIVE_BASE` can help, but the code is far from obvious:

```py
from datetime import datetime
import pytz

forum_tz = pytz.timezone("UTC+5")  # forum time zone

# current time as tz-aware UTC datetime
now_utc = pytz.utc.localize(datetime(2016, 9, 7, 10)) 

# here goes a hack: even though docs say 
# "Dateparser assumes all dates to be in UTC",
# we have to pass RELATIVE_BASE as a naive (not tz-aware) 
# datetime in UTC+5 timezone
forum_now = now_utc.astimezone(forum_tz)  # current time in forum timezone
settings = {
   'RELATIVE_BASE': forum_now.replace(tzinfo=None),
}
dt = dateparser.parse(date_string, settings=settings)

# the returned dt is not in UTC, it is in UTC+5. Convert it back to UTC:
[...]

PR #235, opened 2016-09-26
Timezone fixes

### LINKED ISSUE #212: Timezone confusion (mine)
The package looks great. I am very excited to use it. 

But I am confused by something, if I do:
```
>>> dt = dateparser.parse('2pm', settings={'TIMEZONE': 'US/Eastern', 'RETURN_AS_TIMEZONE_AWARE': True})
```
the output is 
```
datetime.datetime(2016, 7, 22, 10, 0, tzinfo=<DstTzInfo 'US/Eastern' EDT-1 day, 20:00:00 DST>)
```
but I would have thought it should be 14 for the hour as the timezone is the correct one. 

I am confused.

waqasshabbir:
fixes #212, #230 

There was a general confusion with `TIMEZONE` setting. To rectify that, we've added another setting, `TO_TIMEZONE` to make settings self explanatory. It's a major change and is not backward compatible.

Now, `dateparser` consider timezone to be system's default unless specified. e.g.

```
In [2]: parse('10:04 am EST')
Out[2]: datetime.datetime(2016, 9, 26, 10, 4)  # date isn't converted to UTC.
```

No conversion takes place, and specified timezone is attached to the resultant date
```
In [5]: parse('10:04 am', settings={'RETURN_AS_TIMEZONE_AWARE': True, 'TIMEZONE': 'EST'})
Out[5]: datetime.datetime(2016, 9, 26, 10, 4, tzinfo=<StaticTzInfo 'EST'>)
```

Conversion is explicit now and must be done using `TO_TIMEZONE` setting.
```
In [7]: parse('10:04 am', settings={'TIMEZONE': 'EST', 'TO_TIMEZONE': 'Utc'})
Out[7]: datetime.datetime(2016, 9, 26, 15, 4)
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -127,17 +127,22 @@ For more on date order, please look at `Settings`_.
 Timezone and UTC Offset 
 -----------------------
 
-Dateparser assumes all dates to be in UTC if no timezone is specified in the string. To convert the resultant `datetime` object to required timezone. You can do the following:
+`dateparser` automatically detects the timezone if given in the date string. If date has no timezone name/abbreviation or offset, you can still specify it using `TIMEZONE` setting.
 
     >>> parse('January 12, 2012 10:00 PM', settings={'TIMEZONE': 'US/Eastern'})
-    datetime.datetime(2012, 1, 12, 17, 0)
+    datetime.datetime(2012, 1, 12, 22, 0)
+
+You can also convert from one time zone to another using `TO_TIMEZONE` setting.
+
+    >>> parse('10:00 am', settings={'TO_TIMEZONE': 'EDT', 'TIMEZONE': 'EST'})
+    datetime.datetime(2016, 9, 25, 11, 0)
+
+    >>> parse('10:00 am EST', settings={'TO_TIMEZONE': 'EDT'})
+    datetime.datetime(2016, 9, 25, 11, 0)
 
 Support for tzaware objects:
 
-    >>> parse('12 Feb 2015 10:56 PM EST', settings={'RETURN_AS_TIMEZONE_AWARE': True})
-    datetime.datetime(2015, 2, 13, 3, 56, tzinfo=<StaticTzInfo 'UTC'>)
-
-    >>> parse('12 Feb 2015 10:56 PM EST', settings={'RETURN_AS_TIMEZONE_AWARE': True, 'TIMEZONE': None}) 
+    >>> parse('12 Feb 2015 10:56 PM EST', settings={'RETURN_AS_TIMEZONE_AWARE': True}) 
     datetime.datetime(2015, 2, 12, 22, 56, tzinfo=<StaticTzInfo 'EST'>)
 
 For more on timezones, please look at `Settings`_.
@@ -179,11 +184,13 @@ Dependencies
   * ruamel.yaml_ for reading language and configuration files.
   * jdatetime_ to convert *Jalali* dates to *Gregorian*.
   * umalqurra_ to convert *Hijri* dates to *Gregorian*.
+  * tzlocal_ to reliably get local timezone.
 
 .. _dateutil: https://pypi.python.org/pypi/python-dateutil
 .. _ruamel.yaml: https://pypi.python.org/pypi/ruamel.yaml
 .. _jdatetime: https://pypi.python.org/pypi/jdatetime
 .. _umalqurra: https://pypi.python.org/pypi/umalqurra/
+.. _tzlocal: https://pypi.python.org/pypi/tzlocal
 
 
 Supported languages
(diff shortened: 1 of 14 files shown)


kmike:
> Now, dateparser consider timezone to be system's default unless specified. e.g.

Hm, what's the reason for this change? Server timezone has nothing to do with a timezone of a site you're scraping. Using server timezone can produce hard-to-notice bugs - you can e.g. develop code locally, get it working, then deploy it to a server, and if a server happens to have a different timezone, you've got a wrong result in production. With UTC default there was no such a problem.

pganssle:
@kmike @waqasshabbir I honestly don't understand why, from a design perspective, you're adding in this needless complexity to begin with. It seems to me that the right way to do it is to return time zone aware datetimes *if* the string you parsed has a time zone in it, and otherwise return a naive time zone.  One option like, `IGNORE_TIME_ZONE` or something would allow people to specify that you want a naive object with time zone ignored, and then all other time zone logic is handled by the end user. Allowing user-specified lists of string <-> zone mappings is also probably a good idea.

If you do something where you are handling *all* time zone applications as part of the parsing step, people will be confused as to how the time zones are being treated, and you'll end up with a bunch of complicated options for how to deal with different cases using different zones - all of which can easily be handled by the user just manipulating the results using python's (superior) datetime functions.

kmike:
@pganssle there is one problem which can't be solved at post-processing: handling of relative dates like '2 hours ago'. On forums it is common to have relative and absolute dates mixed - older dates are absolute, recent posts/comments have relative dates. Because dateparser returns absolute dates, there must be a way to tell dateparser "this website uses this timezone" in order to parse such relative dates correctly. If you parse these dates as relative to server time or to UTC then you need to convert absolute dates to server time or UTC (otherwise returned dates will be incompatible) - and you need server timezone information for that. 

I agree that all other timezone processing is less critical because it can be handled by an end user as a post-processing step.

kmike:
Probably a better design would be the following:

* return absolute dates with time zone written in them as time-zone aware;
* return absolute dates without time zone as naive datetime objects;
* return relative dates as time-zone aware datetime objects, either in UTC or in server timezone;
* provide a function to normalize all dates to a single timezone; it should have 'default_timezone' argument with timezone for zone-less dates, and 'result_timezone' with a timezone date should be converted to.

waqasshabbir:
@kmike reason for this change is conversion to UTC in either relative dates or incomplete dates like `2 pm` was _implicit_ and caused confusions for many users. For example, the site you are parsing dates from is in Eastern timezone, you get correct result normalized to UTC when you parse `2 hrs ago`. But when you parse `today at 2 pm`, no conversion happens because dateparser by default assumed all dates to be in UTC, thus, a wrong recording.

You can also rephrase below as
> Now, dateparser consider timezone to be system's default unless specified. e.g.

dateparser no longer make implicit conversion to a **default** timezone.

Having to specify settings will help user be more aware of timezone conversions and implications. When you specify `TO_TIMEZONE` you know, you're converting from your environment's timezone to the specified one.

re better design, not really sure, how useful is the idea of mixing naive and tzaware objects when you can do general post-processing/normalizing using `settings` to either return all tzaware objects or naive objects converted to a desired timezone.

@pgnassle dateparser offers all of its primary functionality through one entry point `dateparser.parse`. That's why we introduced `settings` to allow customization of factors to _tailor_ the end result. For example, while parsing incomplete date like `Friday`, we can specify `PREFER_DATES_FROM` to either get Friday's date from the last week or next. If day is missing from a date string like `September 2016`, there's a setting to specify use either first of the month or last day to populate the missing day. Timezone conversion settings share the same idea and that's why are consistent with library's design.

For timezones with similar abbreviations, some work was done early on to adopt only the most commonly used ones and avoid duplicates. [See here](https://github.com/scrapinghub/dateparser/blob/master/dateparser/timezones.py)

pganssle:
@kmike I see how that is, but wouldn't it make more sense for relative dates to return `timedelta` objects, which completely obviates the need for this time zone thing?

@waqasshabbir The problem I have is that you are doing manipulations to the datetime that have no place in a parsing function. I think it's very reasonable to specify a *default* value for the time zone, which is replaced if a time zone is specified in the string. I also think it's very reasonable to say "even if a time zone is specified in the string, ignore it", but what you are doing is providing some facility for doing timezone math *within* the parsing function, which does not make sense, conceptually.

All of this makes the whole function quite confusing, because I would expect `parse('2014-04-21 12:00:00 PST', settings={'TIMEZONE': 'US/Eastern'})` to return `datetime(2014, 4, 21, 12, 0, tzinfo=PST)`, or, at a stretch, `datetime(2014, 4, 21, 16, 0, tzinfo=EST)`, but definitely not `datetime(2014, 4, 21, 16, 0)`. What setting do I use if I want to return something that's naive if the string is naive and tzaware if the string it tzaware? What settings would I use if I want to say, "If it's naive assume the timezone is EST, otherwise use whatever time zone setting is specified in the string?" Those would be the expected behavior of a parser, because they are actually settings about how a string is *interpreted*.

PR #245, opened 2016-11-03
Update zh.yaml

tkisme:
add 点


waqasshabbir:
@tkisme can you please also add a test?

PR #255, opened 2016-11-26
Add 'UT' timezone

andresp99999:
Adding support for 'UT' timezone, I have seen this being used in a few cases, i.e: http://feeds.businesswire.com/BW/Data_Management-rss

`pubDate>Thu, 24 Nov 2016 16:03:00 UT</pubDate>`


waqasshabbir:
Thanks @andresp99999!

PR #257, opened 2016-11-28
Fix timezone error for freshness dates

### LINKED ISSUE #256: Today and Yesterday with time/timezone not fully supported
I ran across a couple of cases using commonly encountered content from scraping that dateparser doesn't seem to handle well.  The settings information appears to have no bearing on the results though it is how I uncovered the issue.  

Note: all tests were run Nov. 27 at approximately 16:25 Mountain Standard Time

dateparser.parse('today 4:25 PM', settings={'TIMEZONE': 'US/Mountain', 'TO_TIMEZONE':'UTC'})
> datetime.datetime(2016, 11, 27, 16, 25)
** should be datetime.datetime(2016, 11, 27, 23, 25)

dateparser.parse('yesterday 4:25 PM', settings={'TIMEZONE': 'US/Mountain', 'TO_TIMEZONE':'UTC'})
> datetime.datetime(2016, 11, 26, 16, 25)
** should be datetime.datetime(2016, 11, 26, 23, 25)

The following work as I would expect using the current time:

dateparser.parse('today', settings={'TIMEZONE': 'US/Mountain', 'TO_TIMEZONE':'UTC'})
> datetime.datetime(2016, 11, 27, 23, 30, 7, 954625)

dateparser.parse('yesterday', settings={'TIMEZONE': 'US/Mountain', 'TO_TIMEZONE':'UTC'})
> datetime.datetime(2016, 11, 26, 23, 30, 30, 38122)

Given the above it appears that either the combination of adding a specific time or the timezone conversion is causing the timezone conversion to not be applied.

Thanks for the assistance!

Mark

waqasshabbir:
fixes #256

### DIFF
--- a/dateparser/freshness_date_parser.py
+++ b/dateparser/freshness_date_parser.py
@@ -7,7 +7,7 @@ from datetime import time
 
 from dateutil.relativedelta import relativedelta
 
-from dateparser.utils import apply_timezone
+from dateparser.utils import apply_timezone, localize_timezone
 from .parser import time_parser
 
 
@@ -44,7 +44,7 @@ class FreshnessDateDataParser(object):
 
         _time = self._parse_time(date_string, settings)
 
-        def  apply_time(dateobj, timeobj):
+        def apply_time(dateobj, timeobj):
             if not isinstance(_time, time):
                 return dateobj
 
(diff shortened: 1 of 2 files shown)


ammar-azif:
Hello @waqasshabbir ,

Since we only need date component from the object returned by `_parse`, can we rename it to `_parse_date` instead to avoid confusion? 

Other than that this looks good to me!

PR #277, opened 2017-02-05
[WIP] Parsing of dates independent of locale

### LINKED ISSUE #238: Result of dateparser.parse() should not depend on locale
```
Python 2.7.12 (default, Jul  1 2016, 15:12:24) 
[GCC 5.4.0 20160609] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import dateparser
>>> dateparser.__version__
'0.5.0'
>>> import locale
>>> locale.getlocale()
(None, None)
>>> dateparser.parse('21-FEB-2016', date_formats=['%d-%b-%Y'], languages=['en'])
datetime.datetime(2016, 2, 21, 0, 0)
>>> locale.setlocale(locale.LC_ALL, 'fr_FR.UTF-8')
'fr_FR.UTF-8'
>>> dateparser.parse('21-FEB-2016', date_formats=['%d-%b-%Y'], languages=['en'])
>>>
```

This is a problem for me: I want to be able to get a good result of dateparser.parse when I give languages=['en'] as argument even if my locale is not english. The users of my software (Odoo) may have any locale, but dateparser.parse() used with the "languages" argument should always give the same result !

waqasshabbir:
Fix #238

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,11 @@ install:
 # command to run tests, e.g. python setup.py test
 script: tox
 
+addons:
+  apt:
+    packages:
+      - language-pack-fr
+
 after_success:
     - codecov
 
--- a/dateparser/parser.py
+++ b/dateparser/parser.py
@@ -7,6 +7,8 @@ from collections import OrderedDict
 from datetime import datetime
 from datetime import timedelta
 
+from dateparser.utils.strptime import strptime
+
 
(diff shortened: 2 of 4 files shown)


waqasshabbir:
thanks @alexis-via for testing the changes. Still need to add tests. Will do it one of these days and will release a new version in a couple weeks as we might want to fix a couple other issues as well.

PR #296, opened 2017-03-23
Version 0.14 of ruamel.yaml causes an import error.

FlipperPA:
I'll include the text in a new issue. Using an earlier version fixes the issue, which the PR will make the setup do. I see there is a branch for pinned version in requirements, and that seems like a much better long term strategy for package stability. Thanks for your efforts! 👍


waqasshabbir:
Fixes #297

PR #305, opened 2017-04-10
Adding swedish language

eragnms:
I have added the Swedish language and tests for it.

/M


tests/test_languages.py
```diff
@@ -10,7 +10,6 @@
 
 from tests import BaseTestCase
 
-
```
waqasshabbir: double blank lines are recommended between imports and class definition. Please revert this change.

dateparser/__init__.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-__version__ = '0.6.0'
+__version__ = '0.7.0.dev1'
```
eragnms: Not sure about the procedures for pull requests, but I believe I have reverted as you proposed.
Thanks for the comment!

tests/test_languages.py
eragnms: Updated...
Thanks

PR #315, opened 2017-05-11
Language.simplify regex compilation and memoization

CJStuart:
Currently, regexes in the `Language.simplify` method are not compiled, and are recompiled
every time the `re.sub` method is called. As a result, the parsing takes a big performance hit, especially in the case where many languages have to be evaluated, for example when parsing dates from a data set with mixed languages or unparseable strings. 

This pull request adds compilation and memoization of the regexes to the Language.simplify method, which improves the performance of the tests by a factor 5.

Performance of tests before:
    
    $ make test
    (...)
    Ran 1628 tests in 68.507s

Performance of tests after:

    $ make test
    (...)
    Ran 1628 tests in 14.957s

### DIFF
--- a/dateparser/languages/language.py
+++ b/dateparser/languages/language.py
@@ -16,6 +16,7 @@ class Language(object):
     _dictionary = None
     _normalized_dictionary = None
     _simplifications = None
+    _simplification_patterns = None
     _normalized_simplifications = None
     _splitters = None
     _wordchars = None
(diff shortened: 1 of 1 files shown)


CJStuart:
A note on the 3.6 build failure: the failed tests are unrelated to this pull request. The error message on out-of-range dates in the datetime module was changed in Python 3.6.1, causing the tests to fail.

PR #318, opened 2017-05-28
tried to add tests for validation.py, utils/__init__.py, strptime.py

eszakharova:
Is it ok that function _validate_pertain_list [uses 'skip' field of language info](https://github.com/scrapinghub/dateparser/blob/master/dateparser/languages/validation.py#L114) and not 'pertain' field while making list of pertain tokens?


tests/test_languages.py
```diff
@@ -663,3 +674,145 @@ class TestAutoDetectLanguageDetectorWithoutRedetection(BaseAutoDetectLanguageDet
...
+    ])
+    def test_validate_skip_list_when_invalid(self, lang_id, lang_info):
+        result = self.validator._validate_skip_list(lang_id, lang_info)
+        self.assertFalse(result)
+
+    @parameterized.expand([
+        param('en', {}),
+
+    ])
+    def test_validate_skip_list_when_abscent(self, lang_id, lang_info):
```
waqasshabbir: only a nitpick small spelling mistake with `absent`.

tests/test_languages.py
```diff
@@ -663,3 +674,145 @@ class TestAutoDetectLanguageDetectorWithoutRedetection(BaseAutoDetectLanguageDet
...
+        result = self.validator._validate_skip_list(lang_id, lang_info)
+        self.assertTrue(result)
+
+    # @parameterized.expand([
+    #     param('en', {'pertain': 'string instead of list'}),
+    #     param('en', {'pertain': ['']}),
+    # ])
+    # def test_validate_pertain_list_when_invalid(self, lang_id, lang_info):
+    #     result = self.validator._validate_pertain_list(lang_id, lang_info)
+    #     self.assertFalse(result)
```
waqasshabbir: why is that commented out? perhaps you want to test for

```param('en', {'pertain': [''], 'skip': 'it is a string'})```

File: tests/test_utils_strptime.py [hunk 3]
```diff
@@ -73,7 +73,15 @@ def test_parsing_date_should_fail_using_datetime_strptime_if_locale_is_non_engli
...
         param('12 Dec 10 10:30:55.000111', '%d %b %y %H:%M:%S.%f', expected=datetime(2010, 12, 12, 10, 30, 55, 111)),
         param('12 Feb 2016 11:41:23', '%d %b %Y %I:%M:%S', expected=datetime(2016, 2, 12, 11, 41, 23)),
+
     ])
     def test_microseconds_are_parsed_correctly(self, date_string, fmt, expected):
         self.when_date_string_is_parsed(date_string, fmt)
         self.then_date_object_is(expected)
+
+    @parameterized.expand([
+        param('11 Dec 10 10:30:2011.999999', '%y %b %S %H:%M:%Y.%f', expected=datetime(2011, 12, 1, 10, 30, 10)),
```
waqasshabbir: can you please remove this test I think dateparser should be able to parse it? Let's raise a new issue for that.

tests/test_utils_strptime.py
```diff
@@ -73,7 +73,15 @@ def test_parsing_date_should_fail_using_datetime_strptime_if_locale_is_non_engli
...
+    ])
+    def test_not_parsing_microseconds_attribute_error(self, datestring, fmt, expected):
```
waqasshabbir: Don't you think it should parse microsecond part as well? Can you try to fix that?

PR #321, opened 2017-06-03
added scripts and data retrieved from unicode CLDR

sarthak5:
Added scripts use to retrieve and store data in desired format, and added the data retrieved. Still support for numbering systems and numerals need to be added and some more issues are to be dealt with.


scripts/get_cldr_data.py
```diff
@@ -0,0 +1,497 @@
...
+
+    try:
+        date_format_string = gregorian_dict["dateFormats"]["short"].upper()
+    except:
+        date_format_string = gregorian_dict["dateFormats"]["short"]["_value"].upper()
+
+    json_dict["date_order"] = DATE_ORDER_PATTERN.sub(
+            r'\1\2\3', DATE_ORDER_PATTERN.search(date_format_string).group())
+
+    json_dict["january"] = [gregorian_dict["months"]["stand-alone"]["wide"]["1"],
```
asadurski: This dict creation part could be made a lot shorter with loops.

scripts/get_cldr_data.py
```diff
@@ -0,0 +1,480 @@
+import requests
```
sarthak5: But these are imported in the scripts which won't be used by users, do we still need to add these to requirements?

scripts/get_cldr_data.py
```diff
@@ -0,0 +1,497 @@
...
+import time
+from collections import OrderedDict
+from orderedset import OrderedSet
+import base64
+import six
+
+from utils import get_dict_difference
+from order_languages import language_locale_dict
+
+OAuth_Access_Token = 'OAuth_Access_Token'       # Add OAuth_Access_Token here
```
sarthak5: I didn't know about that, will make changes soon.

sarthak5:
Thanks for the review @asadurski. I have modified the scripts and they work much faster now.

PR #324, opened 2017-06-13
[WIP] search with translate

eszakharova:


dateparser/search/search.py
```diff
@@ -0,0 +1,189 @@
...
+            self.language_detector = FullTextLanguageDetector(list(self.available_language_map.values()))
+
+        return self.language_detector._best_language(text)
+
+    @apply_settings
+    def search_dates(self, text, languages=None, settings=None):
+        language_shortname = self.detect_language(text=text, languages=languages)
+        if not language_shortname:
+            return {'Language': None, 'Dates': None}
+        return {'Language': language_shortname, 'Dates': self.search.search_parse(language_shortname, text, settings)}
```
asadurski: This should read: `self.search.search_parse(language_shortname, text, settings=settings)`, otherwise the `@apply_settings` wrapper will crash.

asadurski:
Apologies for holding that for so long.
I think we'll prepare a release with bug fixes as 0.7 and a separate release with your GSoC contribution as 0.8.

cc: @sarthak5

adsk2050:
Its not working properly. `dateparser.parse("of 31 Jan 2020, of 31 Jan 2020")` Doesn't return anything

PR #332, opened 2017-07-21
Fixed issue on ru.yaml

Dekat:
"now" isn't recognized as a simplification value !


Gallaecio:

PR #340, opened 2017-08-03
possible fix of 'ago' problem in Russian

eszakharova:
'_2000 год_' in Russian is '_year 2000_'
Before was parsed as '_2000 years ago_'
``` 
parse('2000 год')
datetime.datetime(17, 8, 3, 18, 44, 48, 222615)
```
After
``` 
parse('2000 год')
datetime.datetime(2000, 8, 3, 0, 0)
```

### DIFF
--- a/dateparser/freshness_date_parser.py
+++ b/dateparser/freshness_date_parser.py
@@ -22,6 +22,10 @@ class FreshnessDateDataParser(object):
     def __init__(self):
         self.now = None
 
+    def remove_wrong_kwargs_from_string(self, date_string):
+        date_string = re.sub('(?:^\d+\s+)(year)$', '', date_string)
+        return date_string
+
     def _are_all_words_units(self, date_string):
         skip = [_UNITS,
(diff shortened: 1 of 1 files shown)


Gallaecio:

noviluni:
The reason behind this is that he `relative-time` parser is applied before the `absolute-time` parser... so it could be fixed by using the `PARSERS` settings, either deleting the `relative-time` parser or putting it in the last position.
i.e.:
```
dateparser.parse('2000 year', settings={'PARSERS': ['timestamp', 'custom-formats', 'absolute-time', 'base-formats', 'relative-time']})
```

Your approach is not good because we can't hardcode that regex there and would have side effects for other languages. However, the issue is still valid.

We could probably add a regex into the ru.yaml file.

Maybe a simplification as  `- (\d+)\s*год\s+: \1` would work, however, I can't speak Russian and I don't know if we should add something more, as it could be that with other following or preceding words (as "ago" or "in") this simplification shouldn't be applied.

PR #348, opened 2017-09-11
Austrian German + dative forms added

adbar:
see https://en.wiktionary.org/wiki/J%C3%A4nner and https://en.wiktionary.org/wiki/Feber


Gallaecio:

PR #355, opened 2017-10-29
Add some more month reductions for Russian

bodqhrohro:


Gallaecio:

bodqhrohro:
These reductions already exist in CLDR data, support of which was in development while I was making this patch. So the PR is not needed anymore.

PR #309, opened 2017-04-26
Fixes mistakes and missing words in Czech datafile

janrygl:
Czech language is flective, therefore only lemma in dictionaries is enough.

Example for month June = červen

1st June = 1. června
June 1st = červen 1.

After the fix:
>> In [3]: dateparser.parse(u'6. června 2000', languages=['cs'])
>> Out[3]: datetime.datetime(2000, 6, 6, 0, 0)

Fixes cover:
 * correct weekday name abbrevs
 * all word forms for weekdays and month names
 * other small fixes

PR #367, opened 2017-12-18
Add Czech inflexions (cs.yaml)

KPilnacek:


KPilnacek:
Duplicitate of #309

asadurski:
But anyway, thank you for contributing!

PR #381, opened 2018-02-05
s/nose_parameterized/parameterized

dotlambda:
"The nose-parameterized package is deprecated and has been renamed to parameterized."
I simply followed the instructions from https://pypi.python.org/pypi/nose-parameterized.


asadurski:
Thank you for bringing the repo closer to the present.

PR #384, opened 2018-02-08
Fix History formatting, not well readable on Github

browniebroke:
While checking the changelog to see what changed recently, I found it hard to parse on Github. 
It's because RST is a bit picky regarding vertical spacing, and adding a few blank lines should fix it.


asadurski:
Thank you! That indeed looks a lot better.

PR #399, opened 2018-03-18
Fix: 'Week' and 'Sunday' problem in ID locale

### LINKED ISSUE #398: Problem in Bahasa Indonesia (ID)
I could not parse this string: "Minggu, 18 Mar 2018 07:30"
Which is basic Indonesia and linearly translate to English as "Sunday, 18 Mar 2018 07:30"

Here's my code:
```
import dateparser
print(dateparser.parse(u'Minggu, 18 Mar 2018 07:30'))
```

Interestinglly, "Senin, 18 Mar 2018 07:30" could be parsed correctly. (English: "Monday, 18 Mar 2018 07:30")

Further investigation leads to the incorrect translation from Indonesia to English,
`week` in Indonesia is `minggu`, same as `sunday` is `minggu`, so the `minggu` in the date translated into `week` which does not make sense in English thus returning invalid date.

feryandi:
Add new rule for 'week' and remove existing 'week' translation from Indonesia (id)
This should fix #398

Translation removed:
`week`: `minggu`
reason: in Indonesia language, `minggu` as standalone word is more likely to be translated into `sunday`

Rule added:
\1 week ago:
  - (\d+) minggu lalu
\1 week:
  - (\d+) minggu
reason: to mitigate the missing translation of week, I've hard-coded the most probable sentence which the word `minggu` translated as `week`

### DIFF
--- a/dateparser/data/date_translation_data/id.py
+++ b/dateparser/data/date_translation_data/id.py
@@ -95,7 +95,6 @@ info = {
         "bln"
     ],
     "week": [
-        "minggu",
         "mgg"
     ],
     "day": [
@@ -202,7 +201,8 @@ info = {
         ],
         "\\1 week ago": [
             "(\\d+) minggu yang lalu",
-            "(\\d+) mgg lalu"
+            "(\\d+) mgg lalu",
+            "(\\d+) minggu lalu"
         ],
         "in \\1 day": [
             "dalam (\\d+) hari",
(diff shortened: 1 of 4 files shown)


Gallaecio:
I see a problem with these changes: you are editing a JSON file that comes from Unicode CLDR data (see https://dateparser.readthedocs.io/en/latest/contributing.html#guidelines-for-editing-translation-data). Those files are not meant to be modified manually, only updated from Unicode CLDR data.

PR #414, opened 2018-05-14
[MRG+1] Check prefer_dates_from when parsing strings like "3 days"

Watchful1:
If prefer_dates_from is set to future, default to future for cases like "3 days"

### DIFF
--- a/dateparser/freshness_date_parser.py
+++ b/dateparser/freshness_date_parser.py
@@ -126,7 +126,11 @@ class FreshnessDateDataParser(object):
                     break
 
         td = relativedelta(**kwargs)
-        if re.search(r'\bin\b', date_string) or 'future' in prefer_dates_from:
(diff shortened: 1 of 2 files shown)


Gallaecio:
I’ve added `[WIP]` to the title of your pull request, please remove it or let us know when your code is ready for a review.

Watchful1:
I haven't made any changes since last August, it's been ready to merge since then. The latest two commits were just merging master in.

Gallaecio:
Sorry, I did not noticed the dates. Could you include an automated test covering your changes?

Watchful1:
Added a test.

I somewhat think that prefer future should be the default, instead of prefer past like it is currently. I don't think someone would say "3 days" and mean three days in the past.

Watchful1:
I'm re-submitting this pull request as #530 which is identical, but is from a separate branch so I keep making changes without polluting this pull request.

PR #428, opened 2018-06-26
Faster parsing of invalid inputs when no locale is given

### LINKED ISSUE #426: Very slow parsing of an invalid date
First several calls are above 1 second for me (python 3.6 on OS X):
```
In [1]: import dateparser

In [2]: %time dateparser.parse('not a date')
CPU times: user 1.94 s, sys: 99.2 ms, total: 2.04 s
Wall time: 2.13 s

In [3]: %time dateparser.parse('not a date')
CPU times: user 1.32 s, sys: 7.37 ms, total: 1.32 s
Wall time: 1.33 s
```
After some time it goes down to 150 -- 200 ms, which is still quite slow:
```
In [6]: %time dateparser.parse('not a date')
CPU times: user 145 ms, sys: 2.08 ms, total: 147 ms
Wall time: 152 ms

In [7]: %time dateparser.parse('not a date')
CPU times: user 147 ms, sys: 2 ms, total: 149 ms
Wall time: 153 ms
```

lopuhin:
Fixes #426 

3 commits are unrelated, but all give speed improvements (first two give the most). Overall speedup on ``not a date`` input is about 10x, from ~150ms to ~12-15ms, and good speed is achieved already at the second call, not after ~3-5th like before. 
be9b92d is a minimal fix, but this issue also affects ``BaseLanguageDetector``, so for a more extended fix I could try removing ``strip_timezone`` argument from ``Locale.is_applicable`` altogether and fixing all callers, but I'm not sure if it's worth it without knowing how language detectors are used.

### DIFF
--- a/dateparser/date.py
+++ b/dateparser/date.py
@@ -14,6 +14,7 @@ from dateparser.date_parser import date_parser
 from dateparser.freshness_date_parser import freshness_date_parser
 from dateparser.languages.loader import LocaleDataLoader
 from dateparser.conf import apply_settings
+from dateparser.timezone_parser import pop_tz_offset_from_string
 from dateparser.utils import apply_timezone_from_settings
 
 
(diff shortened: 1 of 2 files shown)


lopuhin:
I think py33 build failed because wheel no longer supports 3.3: https://travis-ci.org/scrapinghub/dateparser/jobs/397048568#L561

lopuhin:
I also checked how this affects "normal" parsing: on a test suite of about 230 dates, using parse or sometimes search_dates, always with given language, speed goes from 23 seconds to 6 seconds with this branch (so about 4x faster)

asadurski:
Thank you very much, @lopuhin!
I took the liberty of switching testing to Python 3.7 on the same branch.

lopuhin:
Thanks @asadurski !

PR #437, opened 2018-07-26
memoize `get_tz_offsets` to improve runtime performance of short-lived scripts

jhjaggars:
I am working on a small project that uses dateparser and noticed that the invocation of my script was taking around 500ms on my local machine to do a very small amount of work.  I took a look at what cProfile had to say and narrowed things down to the compilation of hundreds of regular expressions to be used for the calculation of timezone offsets.

This PR provides an alternative to calculating them up front by using the generator in every situation that `_tz_offsets` was used previously.   For the case where every offset regex needs to be compiled, the results are cached so that later iteration is fast.

### DIFF
--- a/dateparser/timezone_parser.py
+++ b/dateparser/timezone_parser.py
@@ -30,7 +30,7 @@ class StaticTzInfo(tzinfo):
 
 
 def pop_tz_offset_from_string(date_string, as_offset=True):
-    for name, info in _tz_offsets:
+    for name, info in get_tz_offsets():
         timezone_re = info['regex']
         timezone_match = timezone_re.search(date_string)
         if timezone_match:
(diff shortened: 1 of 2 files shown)


dateparser/timezone_parser.py
```diff
@@ -73,5 +76,4 @@ def get_local_tz_offset():
     return offset
 
 
-_tz_offsets = list(get_tz_offsets())
```
lopuhin: Why if this change needed? Instead of using already pre-computed ``_tz_offsets`` (which were computed on import time), now the function is called each time, right? Do you have an example benchmark that shows the performance issues?

jhjaggars: The change makes short-running scripts take less time to execute in many cases.  It accomplishes this by allowing the existing callers of `get_tz_offsets()` to exercise the `break` statement to stop processing early.  In order to make the long-running case fast as well I cached the calculations of each offset.  Here is an example that drove me to submitting this PR:

```
(dateparser) ➜  dateparser git:(master) ✗ cat bench.py
import dateparser
```

```
(dateparser) ➜  dateparser git:(master) ✗ time python bench.py
python bench.py  0.31s user 0.02s system 97% cpu 0.341 total
```

```
(dateparser) ➜  dateparser git:(memoize-get-timezone-offsets) ✗ time python bench.py
python bench.py  0.12s user 0.02s system 95% cpu 0.147 total
```

*However* after looking at the long-running case my changes are pretty bad, since the iteration over all the permutations is very nested.  I'm experimenting with some other techniques to preserve the performance of the (presumably more common) case of long-running applications.

Please close this PR.

lopuhin: @jhjaggars ah this makes sense now, thanks for an explanation! Indeed the time for the first import is quite big, thanks for finding the cause for this.
I'm not a package maintainer though :)

asadurski: Closing. Buuut if you manage to find a way that covers both cases (a setting maybe?), feel free to push it.

jhjaggars: @asadurski I've made a tweak to the way I cache the computed regular expressions that improves the performance of this PR.  In the worst case (When you need the item at the end of the list every time) it's probably ~4x slower than the current method in master.

Currently, I think the only way to really improve the performance in both cases would be to handle the regex lookup in a more clever manner, or as you mentioned some sort of switch to pick the behavior you want.

PR #438, opened 2018-07-26
Fix Portuguese month/day ordering issue

ottumm:
There is a bug with Portuguese dates where dateparser switches the month and day ordering. This regression occurred sometime between v0.6.0 and v0.7.0.

Example text is "segunda 10-02-2015 4:06am", which should parse to "2015-02-10" but currently parses to "2015-10-02".

### DIFF
--- a/tests/test_date_parser.py
+++ b/tests/test_date_parser.py
@@ -78,6 +78,7 @@ class TestDateParser(BaseTestCase):
         param('sexta-feira, 10 de junho de 2014 14:52', datetime(2014, 6, 10, 14, 52)),
         param('13 Setembro, 2014', datetime(2014, 9, 13)),
         param('Sab 3:03', datetime(2012, 11, 10, 3, 3)),
+        param('segunda 10-02-2015 4:06am', datetime(2015, 2, 10, 4, 6)),
         # Russian dates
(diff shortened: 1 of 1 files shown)


ottumm:
I've added a test case that exposes this regression. I'm happy (and motivated) to also fix the bug, but I'd appreciate a point in the right direction. Otherwise I may try using git-bisect to find the commit where this regressed.

ottumm:
Based on my `git-bisect` testing, it looks like https://github.com/scrapinghub/dateparser/commit/8c24e3aa62def8c46fe1b15408c8833bfa47465a is the first commit to break this.

ottumm:
Should I open an issue for this or is this PR sufficient?

Gallaecio:
I don’t think it makes sense to add a broken test to `master`. Instead, I think you should just create an issue, and whatever pull request is created to fix the issue, that should include the test.

ottumm:
I agree. I created this test in this PR, and then identified the commit that broke the test, in the hopes that someone who knows how to fix this bug could take it from there, since I do not.

Gallaecio:
Then please create an issue with those details (the example string that is not parsed as expected, the commit that you believe broke the parsing), also with a link to this pull request, and then close this pull request.

ottumm:
Done, thanks @Gallaecio!

PR #449, opened 2018-09-27
Fixed test scenario to avoid tests stop working in days bigger than 30

### LINKED ISSUE #440: Test failure on some days
The line

https://github.com/scrapinghub/dateparser/blob/a01a4d2071a8f1d4b368543e5e09cde5eb880799/tests/test_search.py#L423

can lead to `ValueError: day is out of range for month` if `datetime.datetime.today().day > 30`.

I think all occurences of `datetime.datetime.today().*` in the tests should be replaced by something deterministic or by a random number in a suitabe range.

rennerocha:
It should fix #440 .


asadurski:
Really good workaround!
:+1: :medal_sports:

PR #450, opened 2018-10-01
Included positional argument in tox.ini

rennerocha:
With this change it is possible to run specific tests using tox. For example:
`tox -e py36 -- tests/test_utils_strptime.py:TestStrptime`


asadurski:
:+1: :medal_sports:

PR #474, opened 2019-01-03
WIP: add test and changes for search_dates#469 

luisandia:


Gallaecio:
I don’t think “Up 1 second” refers to a point in the future, but a point in the past. It is a duration between a point in the past and the present (it has been up for 1 second, so it started being up 1 second in the past).

PR #476, opened 2019-01-04
Added "mediodía" and "medianoche" spanish words

christianfm10:
I've added 'mediodía' and 'medianoche' spanish words and a test for both cases


dateparser/data/date_translation_data/es.py
```diff
@@ -495,6 +495,12 @@
         },
         {
             "un": "1"
+        },
+        {
+            "(?:12\\s+)?mediodía": "12:00"
+        },
+        {
+            "(?:12\\s+)?medianoche": "00:00"
         }
```
eLRuLL: @christianfm10 please add a trailing comma here and also an extra newline at the end of the file. Thanks

Gallaecio:
@christianfm10 Do you plan to continue working on this? Otherwise, would you mind if I take over? I see this only needs dealing with @eLRuLL’s feedback, maybe adding “la” to the list of words (since there is “el”), and adding tests that cover the use of those words to be ignored.

dateparser/data/date_translation_data/es.py
noviluni: This file shouldn't be edited directly, instead he should run the `write_complete_data.py` script that generates the json file from the `.yaml` file.

Issue #411, opened 2018-05-07
Handle new relative date format


cyzanfar:
Are there any plans to add support for date format like this one: `'over a week ago'`? Facebook is using it so might be relevant in the near future.

I can take on this feature if you want and submit a PR.

Let me know Contributors!

asadurski:
Hello @cyzanfar!
You are more than welcome to take it on.

There are some similar feature requests to what I would call approximate parsing, e.g.: https://github.com/scrapinghub/dateparser/issues/390, https://github.com/scrapinghub/dateparser/issues/371.
Approximate, because I would understand "over a week ago" as a number of days between 8 and 13 (more than one week, less than two weeks). But as Dateparser returns a discrete value each time, I was thinking about treating the approximation as a setting - where one could specify the portion and the rounding of the time period that we return.

For instance, with approximation setting of 0.5 and rounding down: `less than a second` could be parsed into 500 ms, `over a week` would be 10 days.
With a setting of 0.1 and rounding down: `less than a second` - 100 ms, `over a week` - 8 days.

And what are your thoughts on this matter?

Issue #46, opened 2015-01-28
Simple arithmetic for the words


Allactaga:
We need to be able to transform token sequences like "seven hundred and sixty-five thousand, four hundred and thirty-two" to the "765432".  There could be different handling of such tokens in different languages (for example Roman numerals deals with subtractions). So let's for now only focus on how English tokens transforming to numbers. Let's call this approach "general" (later we would define which approach should be used in `languages.yaml` file)
Initial idea is to iterate through the list of tokens, skipping tokens that are in `skip`, or `[\W_]+`. Each token should be present in dictionary (`numbers` section of the language).

So if number represented by current token is less then previous, we use addition, if it is greater than several of previous nearby numbers, than those smaller number are describing this bigger one and use multiplication. Be sure to use multiplication only with those preceding number that are 1) less then current 2) directly chained with current.

This approach should of course be properly tested.

Gallaecio:
There is a great potential for code sharing between dateparser and price-parser here. I’ve recently proposed an English-only approach for price-parser (https://github.com/scrapinghub/price-parser/pull/11).

Time for **number-parser**? :grin:

Eveneko:
Hi, I'm interested in this idea. However, I have NLP course this semester. I'm not sure if it is a bit late, But I really want to participate, and I believe I have the ability to start work when I start work in the summer.

varunagarwal18:
There are many implementations for number parsers on Stack overflow. There is also a library called word2number in Python contributed by someone.

Gallaecio:
There are many implementations for English. But the end goal is to support different locales. And in unambiguous cases, without the knowledge of the locale of the input text. That can be hard.

ShantanuDube:
> There are many implementations for English. But the end goal is to support different locales. And in unambiguous cases, without the knowledge of the locale of the input text. That can be hard.

So what should be the main aim of the project... to solve this issue with respect to natural language processsing in english or in all different locales?

noviluni:
Hi @ShantanuDube @varunagarwal18 @Eveneko ! 

The idea is to support every supported language, however, if you check the code, most of the things are first translated to English and then processed to get the date. So this could be done as "X language" --> "English" --> "numbers". The first natural step would be "English" --> "numbers", but we also need to develop a "framework" to easily add support for the other languages. 

[...]

PR #477, opened 2019-01-05
[WIP] Parse natural language time

### LINKED ISSUE #239: Richer natural language parsing of times
It would be nice to be able to parse times that are described in natural language.
For example this works: 
- "today at noon"
- "today at 3 pm"

But these don't: 
- "today at three o'clock" 
- "today at three in the afternoon"

Without thinking about it too much, it doesn't seem to be very difficult to add the relevant words as simplifications. What is your position on this?

### LINKED ISSUE #291: "in three weeks" doesn't work
```
In [2]: dateparser.parse('in two weeks')
Out[2]: datetime.datetime(2017, 3, 31, 23, 42, 45, 268823)

In [3]: dateparser.parse('in three weeks')

In [4]: dateparser.parse('in 3 weeks')
Out[4]: datetime.datetime(2017, 4, 7, 23, 42, 53, 854388)
```

### LINKED ISSUE #383: Unable to parse 'one week ago'
After updating to the latest release, we are unable to parse 'one week ago', although '1 week ago' works. It seems to be a regression from 0.6.0 as downgrading fixes the problem. I'm running Python 3.6.3, under macOS.

```python
>>> import dateparser
>>> dateparser.__version__
'0.7.0'
>>> dateparser.parse('one week ago')
>>> dateparser.parse('1 week ago')
datetime.datetime(2018, 2, 1, 18, 47, 8, 496179)
>>>
```

```python
>>> import dateparser
>>> dateparser.__version__
'0.6.0'
>>> dateparser.parse('one week ago')
datetime.datetime(2018, 2, 1, 18, 43, 28, 833883)
>>>
```

christianfm10:
Fixes #239, fixes #291, fixes #383


dateparser_data/supplementary_language_data/date_translation_data/en.yaml
```diff
@@ -48,3 +48,18 @@ simplifications:
     - (\d+)h(\d+)m?: \1:\2
     - (?<=from\s+)now: in
     - less than 1 minute ago: 45 second ago
```
eLRuLL: I see all of this makes sense, but exactly for that I don't understand why wasn't this add before. @ArturGaspar would you like give your input here please?

eLRuLL:
@christianfm10 please be sure to support also days and exactly what was expected by Pawel here: https://github.com/scrapinghub/dateparser/issues/239#issuecomment-282965615

eLRuLL:
@christianfm10 also please check the recommendation of Eugene here: https://github.com/scrapinghub/dateparser/issues/46

I would actually love to follow that one, as I also agree that having the numbers just thrown there as simplifications shouldn't be the way to go, but every language should be able to define their "numbers" and use them to parse the dates better.

Please let me know what do you think and if you think that is the right approach or if we should improve it somehow. Thanks

dateparser_data/supplementary_language_data/date_translation_data/en.yaml
asadurski: Because of precision. It may well enough be `1 second ago`, `30 seconds ago` or `59.99 seconds ago`. I think we need a setting for that, because `45 seconds` is a size that will not fit all.

Previously discussed under: https://github.com/scrapinghub/dateparser/issues/411#issuecomment-387643982

asadurski:
Regarding https://github.com/scrapinghub/dateparser/issues/46, it was planned for, and we already have quite a lot of data to cover that (see https://github.com/scrapinghub/dateparser/tree/master/dateparser_data/cldr_language_data/numeral_translation_data).
My own, shallow, 3 year old efforts are here: https://github.com/scrapinghub/dateparser/tree/feature-numerals-to-numbers. I remember getting to the stage where single (even compound) numeral in a date was recognised correctly, but when it came to several of them, e.g. for year and days, it didn't work.

dateparser_data/supplementary_language_data/date_translation_data/en.yaml
Gallaecio: You seem to be discussing a line that is not part of these changes here.

Gallaecio (approved):
I think we should, after solving conflicts, merge this as is.

This solves the three linked issues (for numbers up to 12). Even if it is not a complete solution for #46, it does not aim to be, and it is an improvement over the current state of things nonetheless.

Regarding @pawelmhm‘s https://github.com/scrapinghub/dateparser/issues/239#issuecomment-282965615 I think that deserves a separate ticket (if no existing one fits).

dateparser_data/supplementary_language_data/date_translation_data/en.yaml
eLRuLL: I think I was talking about the numbers or I don't remember, so I'll be closing this conversation threa.

eLRuLL (approved):

noviluni:
I think we could proceed with merging this, but shouldn't this:

```
>>> parse("today at three o'clock", settings={'PREFER_DATES_FROM': 'future'})
datetime.datetime(2020, 4, 30, 3, 0)
```

return:
```
datetime.datetime(2020, 4, 30, 15, 0)
```

?

cc: @Gallaecio

noviluni:
(of course, my last comment assumes that you are trying it before 3 PM)

Gallaecio:
I agree, although I would not mind opening a separate issue to handle that.

PR #481, opened 2019-01-09
add exception

luisandia:


Gallaecio:
In order to support future changes to `python-hijiri-ummalqura` which may extend the supported period, shouldn’t we just catch the exception and indicate that the date is not supported by `python-hijiri-ummalqura`, providing a link to an upstream resource that indicates the limits?

PR #484, opened 2019-01-09
Russian abbreviation weekdays

### LINKED ISSUE #436: Missing abbreviation of Russian Thursday
Parsing the following string 
`'19/07/18 Чтв 21:14:09'`
Will result in None.
If you look at [the code](https://github.com/scrapinghub/dateparser/blob/master/dateparser/data/date_translation_data/ru.py)
You'll see that the entry for Thursday contains 
```
        "четверг",
        "чт"
```
I.e missing чтв.

christianfm10:
Added russian abbreviation weekdays: 
Monday: Пнд
Tuesday: Втр
Wednesday: Срд
Thursday: Чтв
Friday: Птн
Saturday: Сбт
Sunday: Вск

Fixes #436


eLRuLL:
We'll need a maintener that knows russian that could approve this.

Also it becomes more evident that we should be splitting all this "dates" related information to something more structured. Maybe inside every language we should separate the files to:

- days
- months

Just to give more perspective, but maybe this approach could affect performance. Maybe just one a better structure defining all that information by sections.

@asadurski I would also like your input here. Thanks

dateparser_data/supplementary_language_data/date_translation_data/ru.yaml
```diff
@@ -2,13 +2,23 @@ skip: ["около", "в", "и"]
 
 sentence_splitter_group : 1
 
+monday:
+    - Пнд
+tuesday:
+    - Втр
 wednesday:
+    - СрД
```
lopuhin: Probably the last letter shouldn't be capitalized, so that would be ``Срд``

asadurski:
Thanks for taking care of these!
However, for those updates to work, you need to run `scripts/write_complete_data.py`, which will merge them with already existing language tokens and will update `ru.py` language data file only with differences.
Then you need to commit that file too.
(please see https://github.com/scrapinghub/dateparser/blob/master/CONTRIBUTING.rst#guidelines-for-editing-translation-data)
Also, as @eLRuLL already mentioned, tests are essential here, so please include them too.

dateparser_data/supplementary_language_data/date_translation_data/ru.yaml
christianfm10: I have already fixed it. By the way, can you review the tests that i added? :pray:

christianfm10:
@asadurski thanks for your reply. I added "ru.py" file and tests, what do you think?

dateparser_data/supplementary_language_data/date_translation_data/ru.yaml
```diff
@@ -2,13 +2,23 @@ skip: ["около", "в", "и"]
...
+thursday:
+    - Чтв
 friday:
+    - Птн
     - Пятницу
 saturday:
+    - Сбт
     - Субботу
 sunday:
+    - Вск
```
lopuhin: Btw, another common one is ``Вскр``

dateparser/data/date_translation_data/ru.py
```diff
@@ -66,34 +66,41 @@
     ],
     "monday": [
         "понедельник",
-        "пн"
+        "пн",
+        "Пнд"
```
lopuhin: Probably not related to this PR, but I'm curious if case matters here, and why is ``пн`` lower but ``Пнд`` capitalized?

Gallaecio:
@christianfm10 Do you think you will have time to address @lopuhin’s feedback?

Otherwise, @eLRuLL, @lopuhin, @asadurski, shall we merge as is?

eLRuLL:
totally up to @lopuhin and @asadurski at this point I think

lopuhin:
In terms of Russian language, changes are good. From the dataparser side, I have the following questions:

- does capitalization matter? If yes then I can check in more detail which capitalization is better
- yaml and py files are different, is this fine?

Gallaecio:
> yaml and py files are different, is this fine?

It is not fine. We need to add a check for this to the CI.

dateparser/data/date_translation_data/ru.py
noviluni: Capitalization doesn't matter when processing dates. However, If we add two different versions of the same word both are added to the py file. This is going to be fixed in this PR by @asadurski: https://github.com/scrapinghub/dateparser/pull/487

noviluni:
Hi @lopuhin 
There is a PR adding tests to avoid merging different `yaml` - `py` files: https://github.com/scrapinghub/dateparser/pull/663. 

On the other side, as answered in the inline comment, capitalization doesn't matter.

PR #486, opened 2019-01-10
Check if the input is a non-empty string

### LINKED ISSUE #470: If pass empty string to dateparser.parse it tooks around 3-4 sec
If we pass empty string to parse method, it takes around 3-4 seconds to process it.
It should have some validations checks into it.

```
import dateparser
import time
st = time.time()
date = dateparser.parse('')
ed = time.time()
print('Date::', date)
print('%2.3f sec' % (ed - st))
```
```
Output:
>> Date:: 
>> 3.434 sec
```

rodp63:
I consider that checking if the input is a non-empty string is essential to start the parse process.

Fixes #470, closes #631


tests/test_input.py
```diff
@@ -0,0 +1,11 @@
...
+from tests import BaseTestCase
+
+class InputTest(BaseTestCase):
+    def empty_input_test(self):
+        result = parse('')
+        self.assertFalse(result)
+
+    def whitespace_input_test(self):
+        result = parse(' \t\n\r\f\v')
+        self.assertFalse(result)
```
BurnzZ: Could we have a test case to check out **Null Bytes**?

Rodp63: Do you mean add a special condition? Something like:
`if (date_string == '\x00'): return`

BurnzZ: Yes :)

Rodp63: Got it! ;)

dateparser/__init__.py
```diff
@@ -44,6 +45,10 @@ def parse(date_string, date_formats=None, languages=None, locales=None, region=N
     :rtype: :class:`datetime <datetime.datetime>`.
     :raises: ValueError - Unknown Language
     """
+
+    if not(re.compile('\S').search(date_string.replace('\x00',''))):
```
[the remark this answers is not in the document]
Rodp63: Got it! You're right

Gallaecio:
Re-running again. If the coverage data is to be believed, this is not working.

PR #487, opened 2019-01-10
CLDR script update

asadurski:
- Skip entries in locales with capitalisation differences.
- Should also slightly influence language detection through reducing the number of options.

TODO: documentation update must follow.

### DIFF
--- a/scripts/utils.py
+++ b/scripts/utils.py
@@ -4,6 +4,10 @@ from collections import OrderedDict
 
 from git import Repo
 
+# Languages with insufficient translation data are excluded
+# TODO: Automate with exclusion criteria.
+AVOID_LANGUAGES = {'cu', 'kkj', 'nds', 'prg', 'tk', 'vai', 'vai-Latn', 'vai-Vaii', 'vo'}
+
 
 def get_raw_data():
     cldr_dates_full_url = "https://github.com/unicode-cldr/cldr-dates-full.git"
@@ -28,10 +32,11 @@ def get_dict_difference(parent_dict, child_dict):
         if not parent_value:
             child_specific_value = child_value
         elif isinstance(child_value, list):
-            child_specific_value = list(set(child_value)-set(parent_value))
+            child_specific_value = list(
+                set(map(unicode.lower, child_value)) - set(map(unicode.lower, parent_value)))
         elif isinstance(child_value, dict):
             child_specific_value = get_dict_difference(parent_value, child_value)
-        elif child_value != parent_value:
+        elif child_value.lower() != parent_value.lower():
             child_specific_value = child_value
         if child_specific_value:
             difference_dict[key] = child_specific_value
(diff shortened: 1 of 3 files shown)


scripts/utils.py
```diff
@@ -1,7 +1,12 @@
 # -*- coding: utf-8 -*-
+import os
 from collections import OrderedDict
+
 from git import Repo
-import os
+
+# Languages with insufficient translation data are excluded
+# TODO: Automate with exclusion criteria.
```
Gallaecio: :lipstick: Do you plan to do this a part of this pull request? Otherwise, it may be better to create a Github issue for it instead and remove this line.

asadurski: The plan was to merge the script together with freshly pulled data from CLDR, but there are too many changes there. I'm postponing that to later than 0.7.2.

Gallaecio:
@asadurski What else is missing here? The pull request description mentions updating the documentation, what needs to be updated there?

scripts/write_complete_data.py
```diff
@@ -38,16 +43,15 @@ def _modify_relative_data(relative_data):
             string = RELATIVE_PATTERN.sub(r'(\\d+)', string)
             value[i] = string
         modified_relative_data[key] = value
-    return modified_relative_data
```
noviluni, line 41: If this is not going to return the `modified_relative_data` dict we don't even need to create it because we don't need to track it.

suggestion: deleting `modified_relative_data = OrderedDict()` and `modified_relative_data[key] = value`.

scripts/utils.py
```diff
@@ -38,6 +39,10 @@ def get_dict_difference(parent_dict, child_dict):
 
 
 def combine_dicts(primary_dict, supplementary_dict):
+    if not primary_dict:
+        return supplementary_dict
+    elif not supplementary_dict:
+        return primary_dict
```
noviluni, line 52:
```suggestion
    if not primary_dict or not supplementary_dict:
        return primary_dict or supplementary_dict
```

this is simpler, but I'm not sure about the readability of my code :thinking:

noviluni:
Hi @asadurski ! :smile: 

I think we could merge it as-is and merge the new CLDR data in a different PR.

I'm editing the same file here to support testing it: https://github.com/scrapinghub/dateparser/pull/663
and I think it would be a good idea to merge this first and resolve the conflicts in the other PR.

What docs would you update? If you don't have time just let me know and we take this PR.

scripts/write_complete_data.py
```diff
@@ -17,18 +19,20 @@
 numeral_translation_directory = '../dateparser/data/numeral_translation_data/'
 
 os.chdir(os.path.dirname(os.path.abspath(__file__)))
+logging.basicConfig(level=logging.INFO, format='%(asctime)s %(message)s')
+log = logging.getLogger('data_scripts')
```
noviluni, line 23: why did you add the log here?

PR #493, opened 2019-01-28
update versions in requirements.txt

### LINKED ISSUE #489: requirements.txt out of date
since #265, the `requirements.txt` file is "pinned" to older versions of dependencies. in some cases they go back to three year old (2016) dependencies. for normal operation, that isn't much of a problem, but it makes the test suite run with an outdated set of packages, which do not reflect real world operation.

in particular, while trying to build a Debian package for this module, I have struggled to make the test suite pass because the modules packaged in Debian are newer than the dependencies specified in `requirements.txt`. at least I *think* that's the reason the test suite is failing in my environment and not travis.

it would be nice to update those dependencies to more recent versions. for example pytz has a 2018 release, but this affects basically other dependencies as well.

thank you for your consideration and for the extensive test suite. :)

dotlambda:
fixes #489

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
-convertdate==2.0.8
-jdatetime==1.8.1
-python-dateutil==2.6.0
-pytz==2016.10
-regex==2016.11.21
-tzlocal==1.3
+convertdate==2.1.3
+jdatetime==3.1.0
+python-dateutil==2.7.5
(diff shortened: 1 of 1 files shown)


asadurski:
Thank you very much for this!
The tests passed after @rennerocha fixed Python 3.7 regex change in https://github.com/scrapinghub/dateparser/pull/495.

PR #495, opened 2019-01-29
Fixed regex pattern to avoid new behaviour of re.split in Python 3.7 …

rennerocha:
…when pattern match an empty string

Backward incompatible changed introduced in Python 3.7
https://docs.python.org/3/library/re.html#re.split


dotlambda:
Is there an ETA for a release with this fix?

asadurski:
Yes @dotlambda, planned for 12th February.

PR #503, opened 2019-02-15
Fix #501 parsing relative time with timezone

### LINKED ISSUE #501: Parsing relative date with timezone returns wrong datetime
Ubuntu 18.10, Python 3.7, dateparser 0.7.0

```
>>> datetime.now().strftime('%Y-%m-%d %H:%M')  # I'm in CET
'2019-02-06 18:42'
>>> dateparser.parse('tomorrow 08:30 CST').strftime('%Y-%m-%d %H:%M %Z')
'2019-02-07 11:42 CST'
```
Two things are wrong IMO:
- the time (08:30) is ignored
- the timezone is converted (to CST)

The expected outcome would be: `'2019-02-07 08:30 CST'`. With absolute dates or without timezone, everything works as expected:
```
>>> dateparser.parse('Feb 7, 2019 08:30 CST').strftime('%Y-%m-%d %H:%M %Z')
'2019-02-07 08:30 CST'
>>> dateparser.parse('tomorrow 08:30').strftime('%Y-%m-%d %H:%M')
'2019-02-07 08:30'
```

ejulio:
Just added a new test case for the issue #501 and noticed that the timezone info was breaking time parsing.
Just moved the timezone info to be parsed before the time info.

Things I noticed and will probably work in future PRs:
* There are some tests breaking, but it could be my environment
* The docs state that `TIMEZONE` default is `UTC`, but actually, it is `local`
* There is a condition in the `parse` that does seem to be covered by a test case

Fixes #501

### DIFF
--- a/dateparser/freshness_date_parser.py
+++ b/dateparser/freshness_date_parser.py
@@ -46,11 +46,9 @@ class FreshnessDateDataParser(object):
         return get_localzone()
 
     def parse(self, date_string, settings):
-
-        _time = self._parse_time(date_string, settings)
-
         date_string = strip_braces(date_string)
         date_string, ptz = pop_tz_offset_from_string(date_string)
+        _time = self._parse_time(date_string, settings)
 
         _settings_tz = settings.TIMEZONE.lower()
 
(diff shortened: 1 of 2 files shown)


noviluni:
Hi @ejulio, 
thank you for fixing this! :rocket: 

Those tests you mention were failing because you didn't have some locales installed on your computer. It was fixed by @Gallaecio here: https://github.com/scrapinghub/dateparser/pull/605 :raised_hands:

PR #509, opened 2019-03-10
Test for isinstance "set" instead of collections.Set

scop:
The latter is deprecated in favor of collections.abc.Set, but just "set"
should suffice here.


dateparser/date.py
```diff
@@ -167,7 +160,7 @@ def __init__(self, locale, date_string, date_formats, settings=None):
         if isinstance(date_formats, six.string_types):
             warn(self.DATE_FORMATS_ERROR_MESSAGE, FutureWarning)
             date_formats = [date_formats]
-        elif not (date_formats is None or isinstance(date_formats, (list, tuple, Set))):
+        elif not (date_formats is None or isinstance(date_formats, (list, tuple, set))):
```
noviluni, line 163: It's not exactly the same. `Set` would include subclasses of `collections.abc.Set` different than `set`, but I agree that it's better to simplify here, as another implementation shouldn't be used here.

PR #511, opened 2019-03-10
Travis setup improvements

scop:


Gallaecio:
Looking at the conflicts, it looks like this should be simplified now to just adding 3.8 tests.

noviluni:
Hi @scop, thank you for adding this PR! We added support for Python 3.8 and this generated conflicts, but your changes related to `pip` look good. 

Could you solve the conflicts again? I will merge this PR just after you do it  :raised_hands: 

Thanks in advance :slightly_smiling_face:

PR #515, opened 2019-04-09
Fix incomplete dates when STRICT_PARSING is True

### LINKED ISSUE #356: Dateparser parses years as a date when STRICT_PARSING = True
Dateparser with `STRICT_PARSING` parses a year wrongly: 
```
>>> dateparser.parse('2017', settings={'STRICT_PARSING': True})
datetime.datetime(2020, 1, 7, 0, 0)
```
I would have expected it to return `None` instead. Did I understand the docs wrongly?

For reference, the parsing seems ok without `STRICT_PARSING`:
```
>>> dateparser.parse('2017')
datetime.datetime(2017, 10, 30, 0, 0)
```

Setup:
- Python 3.5.2
- dateparser==0.6.0

marcosmadr:
Add check to only try to parse complete dates if the date string has at least 6 digits.
Fix cases when it's impossible to guess the date
Ex: 
Date: 11117
Possible dates: 11/01/17 or 01/11/17 or 11/11/07

So I suggest a minimum length of 6 to parse complete dates.

Closes #356

### DIFF
--- a/dateparser/parser.py
+++ b/dateparser/parser.py
@@ -134,7 +134,7 @@ class _no_spaces_parser(object):
     def _filter_datetime_format(self, date_order):
         return filter(
             lambda dt: dt.lower().startswith(date_order),
-            self.date_formats
+            self.date_formats[date_order]
         )
 
     @classmethod
--- a/dateparser/parser.py
+++ b/dateparser/parser.py
@@ -131,6 +131,12 @@ class _no_spaces_parser(object):
             '%d%y%m': sorted(self._all, key=lambda x: x.lower().startswith('%d%y%m'), reverse=True),
         }
 
+    def _filter_datetime_format(self, date_order):
+        return filter(
+            lambda dt: dt.lower().startswith(date_order),
+            self.date_formats
+        )
+
     @classmethod
     def _get_period(cls, format_string):
         for pname, pdrv in sorted(cls.period.items(), key=lambda x: x[0]):
@@ -154,9 +160,13 @@ class _no_spaces_parser(object):
         else:
             order = cls._default_order
         nsp = cls()
+        if settings.STRICT_PARSING:
+            date_formats = nsp._filter_datetime_format(order)
+        else:
+            date_formats = nsp.date_formats[order]
         ambiguous_date = None
         for token, _ in tokens.tokenize():
-            for fmt in nsp.date_formats[order]:
+            for fmt in date_formats:
                 try:
                     dt = strptime(token, fmt), cls._get_period(fmt)
                     if len(str(dt[0].year)) < 4:
--- a/dateparser/parser.py
+++ b/dateparser/parser.py
@@ -11,11 +11,17 @@ from dateparser.utils.strptime import strptime
 
 
 NSP_COMPATIBLE = re.compile(r'\D+')
+NSP_DIGITS = re.compile(r'\d+')
 MERIDIAN = re.compile(r'am|pm')
 MICROSECOND = re.compile(r'\d{1,6}')
+NSP_MIN_DIGITS = 6
 
 
 def no_space_parser_eligibile(datestring):
+    src = NSP_DIGITS.search(datestring)
+    if not src or len(src.group()) < NSP_MIN_DIGITS:
+        return False
+
     src = NSP_COMPATIBLE.search(datestring)
     if not src or ':' == src.group():
         return True
--- a/dateparser/parser.py
+++ b/dateparser/parser.py
@@ -11,17 +11,11 @@ from dateparser.utils.strptime import strptime
 
 
 NSP_COMPATIBLE = re.compile(r'\D+')
-NSP_DIGITS = re.compile(r'\d+')
 MERIDIAN = re.compile(r'am|pm')
 MICROSECOND = re.compile(r'\d{1,6}')
-NSP_MIN_DIGITS = 6
 
 
 def no_space_parser_eligibile(datestring):
-    src = NSP_DIGITS.search(datestring)
-    if not src or len(src.group()) < NSP_MIN_DIGITS:
-        return False
-
     src = NSP_COMPATIBLE.search(datestring)
     if not src or ':' == src.group():
         return True
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -121,7 +121,7 @@ class TestNoSpaceParser(BaseTestCase):
     ])
     def test_min_input_date_is_not_parsed(self, date_string):
         self.given_parser()
-        self.given_settings(settings={'DATE_ORDER': ''})
+        self.given_settings()
         self.when_date_is_parsed(date_string)
         self.then_date_is_not_parsed()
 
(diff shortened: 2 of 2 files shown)


Gallaecio:
Looking at the removed tests, I think we would be removing support for some valid use cases.

I wonder if, with a more convoluted solution, we could drop support for unclear scenarios while maintaining support for other cases where less than 6 digits are used but there is only 1 possible date or 1 of the possible dates is the obvious choice.

marcosmadr:
Hi @Gallaecio , I investigated a bit more to find an alternative approach and I have a suggestion.

The thing is that _no_spaces_parser class tries so many date formats that it almost always get an complete date, even if it doesnt look like correct.

My best guess would be limit no_spaces to parse only the given DATE_ORDER, currently it priorizes the order given but it tries all other combinations. For example:

>>> dp.parse('2011', settings={'STRICT_PARSING': True, 'DATE_ORDER': 'DMY'})
values tried = ['%d%m%Y', '%d%m%y', '%d%m%Y%H%M%S.%f', '%d%m%Y%H%M%S', '%d%m%Y%H%M', '%d%m%Y%H', '%d%m%y%H%M%S.%f', '%d%m%y%H%M%S', '%d%m%y%H%M', '%d%m%y%H', '%Y%m%d', '%Y%d%m', '%m%Y%d', '%m%d%Y', '%d%Y%m', '%y%m%d', '%y%d%m', '%m%y%d', '%m%d%y', '%d%y%m', '%Y%m%d%H%M%S.%f', '%Y%m%d%H%M%S', '%Y%m%d%H%M', '%Y%m%d%H', '%Y%d%m%H%M%S.%f', '%Y%d%m%H%M%S', '%Y%d%m%H%M', '%Y%d%m%H', '%m%Y%d%H%M%S.%f', '%m%Y%d%H%M%S', '%m%Y%d%H%M', '%m%Y%d%H', '%m%d%Y%H%M%S.%f', '%m%d%Y%H%M%S', '%m%d%Y%H%M', '%m%d%Y%H', '%d%Y%m%H%M%S.%f', '%d%Y%m%H%M%S', '%d%Y%m%H%M', '%d%Y%m%H', '%y%m%d%H%M%S.%f', '%y%m%d%H%M%S', '%y%m%d%H%M', '%y%m%d%H', '%y%d%m%H%M%S.%f', '%y%d%m%H%M%S', '%y%d%m%H%M', '%y%d%m%H', '%m%y%d%H%M%S.%f', '%m%y%d%H%M%S', '%m%y%d%H%M', '%m%y%d%H', '%m%d%y%H%M%S.%f', '%m%d%y%H%M%S', '%m%d%y%H%M', '%m%d%y%H', '%d%y%m%H%M%S.%f', '%d%y%m%H%M%S', '%d%y%m%H%M', '%d%y%m%H', '%H%M%S.%f', '%H%M%S', '%H%M', '%H']

result = datetime.datetime(2020, 1, 1, 0, 0)

IMHO, if DATE_ORDER is specified, it should only try with the given order. For example, the command above would try:
>>> dp.parse('2011', settings={'STRICT_PARSING': True, 'DATE_ORDER': 'DMY'})
values tried: ['%d%m%Y', '%d%m%y', '%d%m%Y%H%M%S.%f', '%d%m%Y%H%M%S', '%d%m%Y%H%M', '%d%m%Y%H', '%d%m%y%H%M%S.%f', '%d%m%y%H%M%S', '%d%m%y%H%M', '%d%m%y%H']

result = None

This would fix STRICT_PARSING bug if DATE_ORDER is specified, and I'd go for using the default order if STRICT_PARSING = True and no order is specified.

Gallaecio:
Sounds good to me.

noviluni:
I was about to merge this, but after merging master it stops working... Could you check it?

marcosmadr:
> I was about to merge this, but after merging master it stops working... Could you check it?
> 
> I think that what is happening is that when trying with other locales that have another default order (i.e.: "zh" that has "YMD") it's parsed badly again:
> 
> ```python
> >>> dateparser.parse('2017', settings={'STRICT_PARSING': True})
> 2020-01-07 00:00:00
> 
> >>> dateparser.parse('2017', settings={'STRICT_PARSING': True}, locales=['en'])
> None
> 
> >>> dateparser.parse('2017', settings={'STRICT_PARSING': True}, locales=['zh'])
> 2020-01-07 00:00:00
> ```

Yes, sure I'll check it

marcosmadr:
It seems that `DATE_ORDER` was fixed on `master` and now it updates according to the `locates` parameter.

For example, the result with `ribeiro-fix-complete-dates` branch:
```
>>> dateparser.parse('2017', settings={'STRICT_PARSING': True}, locales=['en'])
--> DATE_ORDER = %m%d%y
>>>
>>> dateparser.parse('2017', settings={'STRICT_PARSING': True}, locales=['zh'])
--> DATE_ORDER = %y%m%d
datetime.datetime(2020, 1, 7, 0, 0)
```

The `strptime` method allows months and days with no leading zero, so it can consider `2017` as `datetime(2020, 1, 7, 0, 0)` depending on `DATE_ORDER` value. There are some tests with 5 digits dates, so probably those 4 digits like `2017` should also be considered valid cases.

tests/test_parser.py
```
test_date_are_parsed_in_order_supplied
...
        param(
            date_string=u"18216",
            expected_date=datetime(2018, 2, 16),
            date_order='YMD',
            expected_period='day',
        ),


        param(
            date_string=u"14271",
            expected_date=datetime(2014, 1, 27),
            date_order='YDM',
            expected_period='day',
        ),


        param(
            date_string=u"21813",
            expected_date=datetime(2018, 2, 13),
            date_order='MYD',
            expected_period='day',
        ),
```
One of the issues on `master` is that it tries to parse non space dates with all possible date formats, including different date order, but prioritizing the date order provided as parameter. Fox example, if `DATE_ORDER` is set to`'DMY`, it will prioritize `DMY` dates, but will try also other combinations like `MDY`, `YMD`, etc... This might lead to undesired matches when the input is an incomplete date, as it tries all possible combinations of date formats, it almost always matches a date (even if its incorrect).

We can see on master, the same test with locales=['en'] parses `2017` as `YMD` even if `DATE_ORDER = MDY`
```
>>> dateparser.parse('2017', settings={'STRICT_PARSING': True}, locales=['en'])
--> DATE_ORDERr = '%m%d%y'
datetime.datetime(2020, 1, 7, 0, 0)
```

This PR implements a filter method that, when `STRICT_PARSING` is true, it filters the date formats attribute keeping only those that comply with `DATE_ORDER`. It helps with those overmatches, but does not solve all the cases, since there are cases that a 4 digit number could be a valid complete date.

Another option to handle cases like this could be a new parameter to define the minimum date length when `STRICT_PARSING` is True, in this case the user can change it according to their needs.

But yeah, by default I think probably it should keep parsing those 4 digits date. What do you think @noviluni  ?

Gallaecio:
@serhii73 noticed https://github.com/scrapinghub/dateparser/issues/356 might have addressed this issue, and indeed looks like that to me.

```python
>>> import dateparser
>>> dateparser.parse("2011", settings={"STRICT_PARSING": True})
>>> dateparser.parse("2017", settings={"STRICT_PARSING": True})
>>> 
```

PR #56, opened 2015-02-13
Implements issue #31

tsrdatatech:
Transformed code to a given-when-then approach.

PR #523, opened 2019-05-09
Support date_formats being a string in parse_with_formats()

### LINKED ISSUE #310: Wrong parsing unless 'date_formats' is a list/tuple.
```
$ python --version
Python 3.6.0

$ pip freeze | grep dateparser
dateparser==0.6.0
```

``` python
In [60]: dateparser.parse("20030731150333", "%Y%m%d%H%M%S")
~/.virtualenvs/foo/lib/python3.6/site-packages/dateparser/date.py:154: FutureWarning: Date formats should be list, tuple or set of strings
  warn(self.DATE_FORMATS_ERROR_MESSAGE, FutureWarning)
Out[60]: datetime.datetime(2033, 6, 22, 20, 11, 55)

In [61]: dateparser.parse("20030731150333", ["%Y%m%d%H%M%S"])
Out[61]: datetime.datetime(2003, 7, 31, 15, 3, 33)
```
I understand that using a plain string should be avoided, but since plain strings are not forbidden parsing should still be correct.

With a quick look, this seems to be easily solvable by adding the following lines in `parse_with_formats(date_string, date_formats, settings)`
``` diff
     period = 'day'
+    if isinstance(date_formats, six.string_types):
+        warn(_DateLanguageParser.DATE_FORMATS_ERROR_MESSAGE, FutureWarning)
+        date_formats = [date_formats]
     for date_format in date_formats:
         try:
```
On second thought, using a function would probably be better. E.g. :
``` python
def normalize_date_formats(date_formats):
    if isinstance(date_formats, six.string_types):
        warn(_DateLanguageParser.DATE_FORMATS_ERROR_MESSAGE, FutureWarning)
        date_formats = [date_formats]
    return date_formats
```
perhaps even:
``` python
def validate_date_formats(date_formats):
    if isinstance(date_formats, six.string_types):
        warn(DateLanguageParser.DATE_FORMATS_ERROR_MESSAGE, FutureWarning)
        date_formats = [date_formats]
    elif not (date_formats is None or isinstance(date_formats, (list, tuple, collections.Set))):
        raise TypeError(DateLanguageParser.DATE_FORMATS_ERROR_MESSAGE)
    return date_formats
```
I could try to make a pull request.

Gallaecio:
Fixes #310

### DIFF
--- a/dateparser/date.py
+++ b/dateparser/date.py
@@ -136,6 +136,9 @@ def parse_with_formats(date_string, date_formats, settings):
     :returns: :class:`datetime.datetime`, dict or None
 
     """
+    if isinstance(date_formats, six.string_types):
+        warn(_DateLocaleParser.DATE_FORMATS_ERROR_MESSAGE, FutureWarning)
+        date_formats = [date_formats]
     period = 'day'
     for date_format in date_formats:
         try:
--- a/tests/test_date.py
+++ b/tests/test_date.py
@@ -287,6 +287,15 @@ class TestParseWithFormatsFunction(BaseTestCase):
                                           month=expected_month,
                                           day=get_last_day_of_month(expected_year, expected_month)))
 
+    @parameterized.expand([
+        param(date_string='25-03-14', date_formats='%d-%m-%y', expected_result=datetime(2014, 3, 25)),
+    ])
+    def test_should_support_a_string_as_date_formats(self, date_string, date_formats, expected_result):
+        self.when_date_is_parsed_with_formats(date_string, date_formats)
+        self.then_date_was_parsed()
+        self.then_parsed_period_is('day')
+        self.then_parsed_date_is(expected_result)
+
     def given_now(self, year, month, day, **time):
         now = datetime(year, month, day, **time)
         datetime_mock = Mock(wraps=datetime)


dateparser/date.py
```diff
@@ -136,6 +136,9 @@ def parse_with_formats(date_string, date_formats, settings):
     :returns: :class:`datetime.datetime`, dict or None
 
     """
+    if isinstance(date_formats, six.string_types):
+        warn(_DateLocaleParser.DATE_FORMATS_ERROR_MESSAGE, FutureWarning)
```
asadurski: I'm not sure about the warning part.
Since we already support strings as formats, why should we throw the `Date formats should be list, tuple or set of strings` warning?

Gallaecio: I don’t really mind either way, I just went with a deprecation warning because it seemed consistent with the current code (due to https://github.com/scrapinghub/dateparser/pull/56/commits/136d301e250540b3fa76bf78a33f345e81fc9e6a).

If we don’t issue a deprecation warning here, maybe we should also remove the warnings from other places. What do you think?

asadurski: Good digging!
Well, let's leave the warning then.

PR #530, opened 2019-06-14
[MRG+1] Check prefer_dates_from when parsing strings like "3 days"

### LINKED ISSUE #563: PREFER_DATES_FROM future and parse("1 min") returns time 1 min ago
```
In [13]: parse("1 min", settings={"PREFER_DATES_FROM" : "future"}) - datetime.now()                                                                                                          
Out[13]: datetime.timedelta(-1, 86339, 999551)
```
Even when 'PREFER_DATES_FROM' is set to 'future', without the 'in' keyword, the time returned in 1 min before.

Expectation is that if future dates are preferred, then `1 min` should return 1 min in future.

Watchful1:
This replaces #414. I wanted to work on different features and didn't want to pollute this pull request, and you can't change the source branch of a pull request after submitting it, so I had to submit a new one.

Fixes #563

### DIFF
--- a/dateparser/freshness_date_parser.py
+++ b/dateparser/freshness_date_parser.py
@@ -126,7 +126,11 @@ class FreshnessDateDataParser(object):
                     break
 
         td = relativedelta(**kwargs)
-        if re.search(r'\bin\b', date_string) or 'future' in prefer_dates_from:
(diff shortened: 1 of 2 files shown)


noviluni:
Thank you very much @Watchful1 ! You did a really good job here :)

PR #531, opened 2019-06-14
Support floats in relative dates

Watchful1:
Support relative dates with floats like "2.5 hours ago".

This includes the commits from #530, but I can recreate this without them if that ends up not being approved.

### DIFF
--- a/dateparser/freshness_date_parser.py
+++ b/dateparser/freshness_date_parser.py
@@ -126,7 +126,11 @@ class FreshnessDateDataParser(object):
                     break
 
         td = relativedelta(**kwargs)
-        if re.search(r'\bin\b', date_string) or 'future' in prefer_dates_from:
+        if re.search(r'\bin\b', date_string):
+            date = self.now + td
+        elif re.search(r'\bago\b', date_string):
+            date = self.now - td
+        elif 'future' in prefer_dates_from:
             date = self.now + td
         else:
             date = self.now - td
--- a/dateparser/freshness_date_parser.py
+++ b/dateparser/freshness_date_parser.py
@@ -126,11 +126,11 @@ class FreshnessDateDataParser(object):
                     break
 
         td = relativedelta(**kwargs)
-        if re.search(r'\bin\b', date_string):
-            date = self.now + td
-        elif re.search(r'\bago\b', date_string):
-            date = self.now - td
-        elif 'future' in prefer_dates_from:
+        if (
+            re.search(r'\bin\b', date_string) or
+            ('future' in prefer_dates_from and
+             not re.search(r'\bago\b', date_string))
+        ):
             date = self.now + td
         else:
             date = self.now - td
--- a/dateparser/freshness_date_parser.py
+++ b/dateparser/freshness_date_parser.py
@@ -93,7 +93,7 @@ class FreshnessDateDataParser(object):
             else:
                 self.now = datetime.now(self.get_local_tz())
 
-        date, period = self._parse_date(date_string)
+        date, period = self._parse_date(date_string, settings)
 
         if date:
             date = apply_time(date, _time)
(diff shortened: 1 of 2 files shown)


Gallaecio:
Even if #530 is merged, its commits will probably squashed into one, requiring you to rewrite this change rebasing it against the master branch. And because of how simply it is, I think this change may even get merged earlier.

So, up to you, but please do consider dropping the unrelated commits from this pull request.

Watchful1:
It looks like dateutil.reletivedelta does not support fractional months or years. See [here](https://github.com/dateutil/dateutil/issues/920). So with this set of changes, if one is passed, it just errors.

That's arguable better than the current situation though, which just grabs the part after the decimal. IE, the current version parses "in 1.5 months" as "5 months". So maybe erroring is better than that.

We could wait for the changes to go through in dateutil to support this, but it's a big library and this is an edge case, so who knows when that might happen. Or we could just pick one of the options I listed in that thread and I could implement it here myself. Just having a default of 30 days would make it trivially easy and still cover 99% of situations.

Watchful1:
My git-fu failed me, but I got everything not associated reverted.

This is ready for review/merging.

As noted above, this change makes parse() throw an error in cases like "in 1.5 months" instead of incorrectly parsing it as "5 months". That will happen for any non-integer relative dates with months or years, but it will work for everything else. I'm working on the pull request to dateutil that will fix that.

Watchful1:
`search_dates` splits sentences before trying to parse them. This splitting used the regex `[\.!?;…\r\n]+(?:\s|$)*`, which, because of the `*` on the non-capturing group at the end, did not require whitespace after the separator. Which meant that a phrase like `in 2.5 years` got split into the sentences `in 2` and `5 years`, which resulted in search dates parsing the string as `5 years`.

I've updated that regex to end in a `+` instead of `*`, so it requires the whitespace to end a sentence. Oddly, the other 5 regex's for other languages all already ended in `+`, so I'm unsure if the `*` was an accident, or if it was intended.

If it was intended, I could build a more complicated regex that only prevented splitting of sentences when the period was followed by a number.

I also added a test for this case, and fixed a different, unrelated test that relied on the old behavior to check for an error condition.

tests/test_search.py
```diff
@@ -665,7 +668,8 @@ def test_detection(self, shortname, text):
         param(text="DECEMBER 21 19.87 87",
               languages=None,
               settings=None,
-              expected=[('DECEMBER 21 19', datetime.datetime(2019, 12, 21, 0, 0))]
+              expected=[('DECEMBER', datetime.datetime(2019, 12, 21, 0, 0)),
+                        ('19.87', datetime.datetime(1987, 12, 19, 0, 0))]
```
Gallaecio: Changing this test does not look right. Interpreting `19.87` as the year `1987` definitely looks wrong.

Watchful1: This is a bit odd because the parser can't figure out anything from the full string `DECEMBER 21 19.87 87`. So it splits it up into a bunch of different substrings and tries to figure out what returns the best results. In this case, if it splits on spaces it ends up with the most results.

When trying to parse `19.87`, it splits on the `.`, tries to figure out what `19` could represent, comes up with a day of the month, then tries to figure out what `87` could mean, and comes up with a year. It then guesses that the year `87` refers to `1987` rather than `2087 and returns that.

You can replicate this by just running `search_dates("19.87")`. Or if you run `search_dates("21.87")` it returns the 21st of the month, still in 1987. I can edit the test to use 21 instead of 19 to make it a bit more clear if you want.

Gallaecio:
@Watchful1 Do you plan to resume work on this?

Watchful1:
Other than the merge conflicts I don't think there's anything else to change here. It was just waiting for you to sign off on it.

It looks like dateutil just added my MR over there to a milestone, so maybe they will merge it in soon, but it's not necessary for this work.

Were there further changes you wanted me to make?

Gallaecio:
> Were there further changes you wanted me to make?

I was hoping you could address https://github.com/scrapinghub/dateparser/pull/531#discussion_r347118918, but re-reading your comment I guess it is what it is. The previous test interpretation was not much better. So I guess fixing the conflicts is enough.

Issue #529, opened 2019-06-11
Required Date Parts (semi-strict parsing)


IMP1:
It would be really nice if there was something like `STRICT_PARSING`, but for parts of the date. I'd like to be able to specific to only return dates that have a year and a month, but I don't mind a missing day part.

I haven't had a lot of time to look through the code, but it seems like the only place `STRICT_PARSING` is used is in [`dateparser/parser.py`](https://github.com/scrapinghub/dateparser/blob/master/dateparser/parser.py#L326):

```python
    def _results(self):
        if self.settings.STRICT_PARSING:
            errors = []
            if not self.day:
                errors.append('Day')
            if not self.month:
                errors.append('Month')
            if not self.year:
                errors.append('Year')
            if errors:
                raise ValueError('%s not found in the date string' % ''.join(errors))
```

If this is the *only* place where it's used, and people are happy with this change/addition, I could make up a pull-request sometime soon.

noviluni:
Closed by: https://github.com/scrapinghub/dateparser/pull/703

PR #532, opened 2019-06-17
Require Parts of Date

IMP1:
This was to address the issue I raised in #529. 

I've added to the settings `REQUIRE_PARTS` which takes a string that can include `'day'`, `'month'` and/or `'year'`. I added the setting to the documentation and also realised that there was some docs missing from the README page (with links that didn't go anywhere), so I copied over the documentation from `docs/usage.rst`.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -227,6 +227,138 @@ You can extract dates from longer strings of text. They are returned as list of 
 .. automodule:: dateparser.search
    :members: search_dates
 
+Settings
+========
+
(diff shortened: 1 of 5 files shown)


Gallaecio:
(closing and reopening to see if the Python 3.4 issue passes this time around)

Gallaecio:
Shouldn’t the value be a tuple, list or set, instead of a string?

PR #540, opened 2019-07-17
Add UTC -2:30 as a valid offset. 

jkukul:
At the moment parsing a string like `2019-07-17T12:30:00.000-02:30` doesn't work, because `-02:30` is not present in the valid offsets list.

Timestamps with this offset are observed in Newfoundland Time Zone (NT) during daylight saving time (NDT timezone). This offset is not listed on the linked wikipedia page of UTC time offsets, because "Places that observe daylight saving time (DST) during their respective summer periods are listed only once, at the offset for their winter (usually known as "standard") period;"


jkukul:
Hi all! A friendly reminder - anything holding back a merge of this PR?

PR #542, opened 2019-07-21
Stabilize tests at midnight

orivej:
Ensure that tests that compare times around now use identical now.

---

I was running the tests at midnight and noticed some failures.

### DIFF
--- a/tests/test_date.py
+++ b/tests/test_date.py
@@ -342,15 +342,16 @@ class TestDateDataParser(BaseTestCase):
         param(' Yesterday \n', days_ago=1),
         param('Ontem', days_ago=1),
         param('Ieri', days_ago=1),
+        param(u'вчера', days_ago=1),
+        param(u'снощи', days_ago=1),
         # Day before yesterday
         param('the day before yesterday', days_ago=2),
         param('The DAY before Yesterday', days_ago=2),
         param('Anteontem', days_ago=2),
-        param('Avant-hier', days_ago=2),
-        param(u'вчера', days_ago=1),
-        param(u'снощи', days_ago=1)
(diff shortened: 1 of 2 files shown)


orivej:
Even though this change seems useful, it turned out not to fix my original issue: when `TZ` environment variable is set to a value not supported by pytz/tzlocal but supported by libc/Python, such as `TZ=NZST`, dateparser miscalculates dates.

tests/test_date.py
```diff
@@ -489,11 +490,12 @@ def test_get_date_tuple(self, date_string, expected_result):
...
 
     def given_now(self, year, month, day, **time):
+        now = datetime(year, month, day, **time)
         datetime_mock = Mock(wraps=datetime)
-        datetime_mock.utcnow = Mock(return_value=datetime(year, month, day, **time))
-        self.add_patch(
-            patch('dateparser.date_parser.datetime', new=datetime_mock)
-        )
+        datetime_mock.utcnow = Mock(return_value=now)
+        datetime_mock.now = Mock(return_value=now)
```
Gallaecio, line 496: This seems wrong, `datetime.now` should have timezone information.

orivej: The docs say it shouldn't unless `now` is called with an optional `tz` argument, which seems not to be the case during these tests.

Gallaecio: I did _not_ know that, interesting!

tests/test_search.py
```diff
@@ -1,3 +1,5 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
```
wRAR, line 2: Without Python 2 support this line is not needed.

Issue #488, opened 2019-01-11
search_dates neglects timezone information


gypapp:
It looks quite strange to me that while `parse()` manages to parse timezone information and returns timezone-aware datetime object by default:
```
>>> from dateparser import parse
>>> parse('January 12, 2012 10:00 PM EST')
datetime.datetime(2012, 1, 12, 22, 0, tzinfo=<StaticTzInfo 'EST'>)
>>> parse("Aug 06, 2018 05:05 PM CDT")
datetime.datetime(2018, 8, 6, 17, 5, tzinfo=<StaticTzInfo 'CDT'>)
```
The same does not apply to `dateparser.search.search_dates()`:
```
>>> from dateparser.search import search_dates
>>> search_dates("Aug 06, 2018 05:05 PM CDT")
[('Aug 06, 2018 05:05 PM', datetime.datetime(2018, 8, 6, 17, 5))]
```

asadurski:
Closed with https://github.com/scrapinghub/dateparser/pull/544.

PR #544, opened 2019-07-24
capture timezone info for dates search

peonone:
`dateparser.search.search_dates` does not capture timezone info while `dateparser.parse` does. (See #488 )
Added timezone support for the search.

It may bring some false positives if the text contains a word stands for a timezone but not intent to

### DIFF
--- a/dateparser/languages/locale.py
+++ b/dateparser/languages/locale.py
@@ -8,7 +8,7 @@ from collections import OrderedDict
 
 from dateutil import parser
 
-from dateparser.timezone_parser import pop_tz_offset_from_string
+from dateparser.timezone_parser import pop_tz_offset_from_string, is_word_match_any_tz
 from dateparser.utils import normalize_unicode, combine_dicts
 
 from .dictionary import Dictionary, NormalizedDictionary, ALWAYS_KEEP_TOKENS
@@ -205,6 +205,9 @@ class Locale(object):
                         translated_chunk.append(dictionary[word.strip('()\"\'{}[],.،')])
                     original_chunk.append(original_tokens[i])
                 elif self._token_with_digits_is_ok(word):
+                    translated_chunk.append(word)
+                    original_chunk.append(original_tokens[i])
+                elif translated_chunk and is_word_match_any_tz(word):
                     translated_chunk.append(word)
                     original_chunk.append(original_tokens[i])
                 else:
--- a/dateparser/languages/locale.py
+++ b/dateparser/languages/locale.py
@@ -207,7 +207,8 @@ class Locale(object):
                 elif self._token_with_digits_is_ok(word):
                     translated_chunk.append(word)
                     original_chunk.append(original_tokens[i])
-                elif translated_chunk and is_word_match_any_tz(word):
+                # Use original token because is_word_match_any_tz is case sensitive
+                elif translated_chunk and is_word_match_any_tz(original_tokens[i]):
                     translated_chunk.append(word)
                     original_chunk.append(original_tokens[i])
                 else:
--- a/dateparser/languages/locale.py
+++ b/dateparser/languages/locale.py
@@ -8,7 +8,7 @@ from collections import OrderedDict
 
 from dateutil import parser
 
-from dateparser.timezone_parser import pop_tz_offset_from_string, is_word_match_any_tz
+from dateparser.timezone_parser import pop_tz_offset_from_string, word_is_tz
 from dateparser.utils import normalize_unicode, combine_dicts
 
 from .dictionary import Dictionary, NormalizedDictionary, ALWAYS_KEEP_TOKENS
@@ -207,8 +207,8 @@ class Locale(object):
                 elif self._token_with_digits_is_ok(word):
                     translated_chunk.append(word)
                     original_chunk.append(original_tokens[i])
-                # Use original token because is_word_match_any_tz is case sensitive
-                elif translated_chunk and is_word_match_any_tz(original_tokens[i]):
+                # Use original token because word_is_tz is case sensitive
+                elif translated_chunk and word_is_tz(original_tokens[i]):
                     translated_chunk.append(word)
                     original_chunk.append(original_tokens[i])
                 else:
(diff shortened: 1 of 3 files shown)


tests/test_search.py
```diff
@@ -250,9 +252,8 @@ def test_search_date_string(self, shortname, datetime_string):
               settings={'RELATIVE_BASE': datetime.datetime(2000, 1, 1)}),
 
         # Dutch
-        param('nl', ' De meest dramatische uitbreiding van het conflict vond plaats op 22 juni 1941 met de '
```
peonone: Updated the text as it contains `CET` which is a valid timezone and causes a false positive.

tests/test_search.py
```diff
@@ -279,10 +285,8 @@ def test_search_date_string(self, shortname, datetime_string):
               settings={'RELATIVE_BASE': datetime.datetime(2000, 1, 1)}),
 
         # French
-        param('fr', 'La 2e Guerre mondiale, ou Deuxième Guerre mondiale4, est un conflit armé à '
```
peonone: Similar as the one for Dutch, it contains `est`

dateparser/timezone_parser.py
```diff
@@ -44,6 +44,16 @@ def pop_tz_offset_from_string(date_string, as_offset=True):
         return date_string, None
 
 
+def is_word_match_any_tz(word):
+    # The regex only interests in timezone not at the beginning
+    word = ' ' + word
+    for name, info in _tz_offsets:
+        timezone_re = info['regex']
+        if timezone_re.match(word):
```
asadurski: Could we work on eliminating false positives a bit more? The reason is that there's quite a lot of time zone abbreviations that exist as words in some languages, .e.g.: WITA, VET, UT, SWAT, NUT.
I think that would require accepting only timezones in all caps, and perhaps also modifying the timezone regex with additional conditions, to avoid matching a timezone in regular sentence, e.g.:
`(?:<?(AM|PM|\d|\s?))timezone_regex`.
Not sure about this though, maybe it would be easier to only match timezones next to already detected time components of a date? This couldn't be done on single token level though.

tests/test_search.py
asadurski: Let's try a little harder to make these tests work without modifying, OK?

dateparser/timezone_parser.py
```diff
@@ -44,6 +44,16 @@ def pop_tz_offset_from_string(date_string, as_offset=True):
         return date_string, None
 
 
+def is_word_match_any_tz(word):
+    # The regex only interests in timezone not at the beginning
+    word = ' ' + word
+    for name, info in _tz_offsets:
```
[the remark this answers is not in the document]
peonone: You're right. I fixed it to use single regex

dateparser/timezone_parser.py
peonone: I fixed it to check case-sensitively, which means only in all caps matches it.
Adding additional conditions here needs more change, as it tests the tokenized word only.
But we have something similar in the caller of this function, it only runs this test if any previous token matches a digit, month, week, etc. It only helps to improve performance as it does not need to test the regex for every word.

peonone:
> Thank you for that, @peonone!
> 
> I'm leaving some comments for your consideration.
> Would be good if tests could pass without modifying them, there's also a performance improvement we could try.
> Let me know what you think about those.

Thanks for the feedback @asadurski 
I made the fixes and now all original test cases passed.

tests/test_search.py
peonone: I reverted the test case and now it passed.

tests/test_search.py
peonone: Reverted the test case and it passed

dateparser/timezone_parser.py
```diff
@@ -44,11 +44,15 @@ def pop_tz_offset_from_string(date_string, as_offset=True):
         return date_string, None
 
 
+def is_word_match_any_tz(word):
```
Gallaecio: `is_word_match_any_tz` does not sound right to me. What about `word_is_tz`?

peonone: Yep, it's better, fixed as it.

dateparser/languages/locale.py
```diff
@@ -207,6 +207,10 @@ def translate_search(self, search_string, settings=None):
                 elif self._token_with_digits_is_ok(word):
                     translated_chunk.append(word)
                     original_chunk.append(original_tokens[i])
+                # Use original token because word_is_tz is case sensitive
+                elif translated_chunk and word_is_tz(original_tokens[i]):
+                    translated_chunk.append(word)
+                    original_chunk.append(original_tokens[i])
                 else:
                     if translated_chunk:
```
Gallaecio: :lipstick: I’m thinking maybe we could replace

```
elif translated_chunk and word_is_tz(original_tokens[i]):
    # 1
else:
    if translated_chunk:
        # 2
```

with

```
elif translated_chunk:
    if word_is_tz(original_tokens[i]):
        # 1
    else:
        # 2
```

dateparser/timezone_parser.py
```diff
@@ -44,11 +44,15 @@ def pop_tz_offset_from_string(date_string, as_offset=True):
...
+def word_is_tz(word):
+    return bool(_search_regex.match(word))
+
+
 def convert_to_local_tz(datetime_obj, datetime_tz_offset):
     return datetime_obj - datetime_tz_offset + local_tz_offset
 
 
-def get_tz_offsets():
+def build_tz_offsets(search_regex_parts):
```
Gallaecio: @asadurski Should backward-compatibility be a concern here? Otherwise, should we prefix this function with `_`?

dateparser/languages/locale.py
peonone: I prefer to keep the current way as I think the logic is more clear with such case conditions:
```
if word-is-in-dictionary:
  # 1
elif word-is-digit:
  # 2
elif word-is-timezone:
  #3 
else:
  if has-chunk:
    # 4
```

dateparser/timezone_parser.py
asadurski: No, I think the function is hidden so deep - and used only once - that we shouldn't concern ourselves about it.

PR #547, opened 2019-07-30
fix trailing colon

### LINKED ISSUE #537: Failed to parse, if date ends with colon, "06/25/2019:"

andersonberg:
Fixes #537 
I added code to strip trailing colon and also a test for this code

### DIFF
--- a/dateparser/date.py
+++ b/dateparser/date.py
@@ -39,6 +39,7 @@ APOSTROPHE_LOOK_ALIKE_CHARS = [
 RE_NBSP = re.compile(u'\xa0', flags=re.UNICODE)
 RE_SPACES = re.compile(r'\s+')
 RE_TRIM_SPACES = re.compile(r'^\s+(\S.*?)\s+$')
+RE_TRIM_COLONS = re.compile(r':*(\S.*?):*$')
 
 RE_SANITIZE_SKIP = re.compile(r'\t|\n|\r|\u00bb|,\s\u0432|\u200e|\xb7|\u200f|\u064e|\u064f', flags=re.M)
 RE_SANITIZE_RUSSIAN = re.compile(r'([\W\d])\u0433\.', flags=re.I | re.U)
@@ -113,6 +114,7 @@ def sanitize_date(date_string):
     date_string = sanitize_spaces(date_string)
     date_string = RE_SANITIZE_PERIOD.sub('', date_string)
     date_string = RE_SANITIZE_ON.sub(r'\1', date_string)
+    date_string = RE_TRIM_COLONS.sub(r'\1', date_string)
 
     date_string = RE_SANITIZE_APOSTROPHE.sub(u"'", date_string)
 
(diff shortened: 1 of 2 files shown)


tests/test_date.py
```diff
@@ -637,6 +637,7 @@ def test_remove_year_in_russian(self):
         self.assertEqual(date.sanitize_date(u'2005г.'), u'2005 ')
         self.assertEqual(date.sanitize_date(u'2005 г.'), u'2005 ')
         self.assertEqual(date.sanitize_date(u'Авг.'), u'Авг')
+        self.assertEqual(date.sanitize_date(u'2019:'), u'2019')
```
Gallaecio: I don’t think `test_remove_year_in_russian` is the right test function for this change.

dateparser/date.py
```diff
@@ -39,6 +39,7 @@
 RE_NBSP = re.compile(u'\xa0', flags=re.UNICODE)
 RE_SPACES = re.compile(r'\s+')
 RE_TRIM_SPACES = re.compile(r'^\s+(\S.*?)\s+$')
+RE_TRIM_COLONS = re.compile(r':*(\S.*?):*$')
```
Gallaecio: I don’t understand how it works. I would expect that greedy `:*` at the beginning to cause issues, but it seems to behave as non-greedy here, as opposed to `:+`:

```
>>> re.sub(r':+(\S.*?):*$', r'\1', '2019-07-31 12:00:')
'2019-07-31 1200'
>>> re.sub(r':*(\S.*?):*$', r'\1', '2019-07-31 12:00:')
'2019-07-31 12:00'
```

In any case, what would you think about adding a `^` at the beginning of the expression to make it more obvious that the `:` capture before the capture group is only meant to strip colons from the beginning?

andersonberg: I think we can safely change to `^`. I added at the begining to be more safe, but I think we don't need it.

dateparser/date.py
```diff
@@ -39,6 +39,7 @@
 RE_NBSP = re.compile(u'\xa0', flags=re.UNICODE)
 RE_SPACES = re.compile(r'\s+')
 RE_TRIM_SPACES = re.compile(r'^\s+(\S.*?)\s+$')
+RE_TRIM_COLONS = re.compile(r'(\S.*?):*$')
```
Gallaecio: If we only strip from the beginning, I think `\s*:+$` and `.sub(r'', date_string)` make a bit more sense: we don’t find a match unless there is at least one colon, and we also remove any extra spacing there might be before the colon (I assume space after the colon would have be stripped already somewhere else, but I did not check).

asadurski: > I assume space after the colon would have be stripped already somewhere else, but I did not check

It wouldn't, and it's probably a bug... `'31/07/2019: '` won't be parsed, because `RE_TRIM_SPACES` assumes spaces present at both sides of the date string.
I'm not sure if it should be handled in this PR though.

andersonberg: not sure what's the best approach here. what do you think?

asadurski: I'd leave it for another PR. Your approach undeniably fixes the target issue - and that's the point.

PR #556, opened 2019-09-03
convert indents to spaces

noviluni:
The `rst` files are mixing spaces and tabs. This PR fixes the inconsistent indentation (all converted to spaces).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -3,16 +3,16 @@ dateparser -- python parser for human readable dates
 ====================================================
 
 .. image:: https://img.shields.io/travis/scrapinghub/dateparser/master.svg?style=flat-square
-    :target: https://travis-ci.org/scrapinghub/dateparser
-    :alt: travis build status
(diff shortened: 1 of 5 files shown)


README.rst
```diff
@@ -3,16 +3,16 @@ dateparser -- python parser for human readable dates
 ====================================================
 
 .. image:: https://img.shields.io/travis/scrapinghub/dateparser/master.svg?style=flat-square
-    :target: https://travis-ci.org/scrapinghub/dateparser
-    :alt: travis build status
+	:target: https://travis-ci.org/scrapinghub/dateparser
```
Gallaecio: The convention is to align these options to the directive word, so 3 spaces should be used.

I also see some other instances below where, as far as it looks on Github, you replaced a tab with a single space. I would imagine a tab would be usually replaced by 3 spaces.

noviluni: totally agree, I tried to fix it

Issue #560, opened 2019-09-05
Adding `DEBUG_INFO` setting


noviluni:
I was thinking about how to improve the "debugging" process when weird things happen...

And I realized that it could be really useful to add an option to "see" what happens behind the scenes. It could be, for example, a setting called `DEBUG_INFO` that when `True` shows relevant information to help you to understand what's happening.

I was trying a little and that's what I got.

- - - - -

Input:
```
dateparser.parse('07/03/2009')
```

Log:
```
Interpretation: 07 - 03 - 2009
Order: month, day, year
Used locale: en
```

Out:
```
2009-07-03 00:00:00`
```
- - - - -

Input:
```
dateparser.parse('2009')
```

Log:
```
Day, Month not found in the date string
Interpretation: 2009
Order: year
Used locale: en
```

Out:
```
2009-09-05 00:00:00
```

- - - - -

Input:
```
dateparser.parse('Feb 2009')
```

Log:
```
Day not found in the date string
Interpretation: february - 2009
Order: year
Used locale: en
```

Out:
```
2009-02-05 00:00:00
```
- - - - -

Input:
```
dateparser.parse('07/03/2009', locales=['fr'])
```
Log:
```
Interpretation: 07 - 03 - 2009
Order: day, month, year
Used locale: fr
```

Out:
```
2009-03-07 00:00:00
```
- - - - -

Input:
```
dateparser.parse('en 2 días')
```

Log:
```
Used locale: es
```

Out:
```
2019-09-07 17:46:02.653798
```

Of course there are a lot of new information that could appear in the `log`, that's just an idea.

What do you think?

noviluni:
I just added a draft PR with the code I used to get the `log` in the examples.

noviluni:
It could help in some situations like this: https://github.com/scrapinghub/dateparser/issues/520#issuecomment-489556731

noviluni:
I expect to continue with this PR after merging this: https://github.com/scrapinghub/dateparser/pull/603 as it would be a good starting point :)

noviluni:
Related to this: https://github.com/scrapinghub/dateparser/pull/729

PR #561, opened 2019-09-05
wip: first implementation

noviluni:
Just a PoC to illustrate issue: https://github.com/scrapinghub/dateparser/issues/560


asadurski:
Interesting! This would definitely be helpful when filing new issues.
Please keep in mind to avoid impact to performance when DEBUG is False.

noviluni:
Waiting to merge this: https://github.com/scrapinghub/dateparser/pull/603 before continuing

noviluni:
I think once we implement this: https://github.com/scrapinghub/dateparser/pull/778 this will be trivial, as we only have to log the information from the DateData object.

PR #569, opened 2019-09-20
faster pop_tz_offset_from_string, 2x faster dateparser.parse if date has no timezone info

lopuhin:
When most strings don't have tz offset, this is massively faster, as we avoid a loop over all timezones (around 800 of them).
But it's possible to improve this.

TODO:

- [x] benchmark results
- [ ] improve case when we do find a match

### DIFF
--- a/dateparser/timezone_parser.py
+++ b/dateparser/timezone_parser.py
@@ -33,15 +33,17 @@ class StaticTzInfo(tzinfo):
 
 
 def pop_tz_offset_from_string(date_string, as_offset=True):
-    for name, info in _tz_offsets:
-        timezone_re = info['regex']
-        timezone_match = timezone_re.search(date_string)
-        if timezone_match:
-            start, stop = timezone_match.span()
-            date_string = date_string[:start + 1] + date_string[stop:]
-            return date_string, StaticTzInfo(name, info['offset']) if as_offset else name
-    else:
-        return date_string, None
+    if _search_regex.search(date_string):
+        for name, info in _tz_offsets:
+            timezone_re = info['regex']
+            timezone_match = timezone_re.search(date_string)
+            if timezone_match:
+                start, stop = timezone_match.span()
+                date_string = date_string[:start + 1] + date_string[stop:]
+                return (
+                    date_string,
+                    StaticTzInfo(name, info['offset']) if as_offset else name)
+    return date_string, None
 
 
 def word_is_tz(word):
--- a/dateparser/timezone_parser.py
+++ b/dateparser/timezone_parser.py
@@ -33,7 +33,7 @@ class StaticTzInfo(tzinfo):
 
 
 def pop_tz_offset_from_string(date_string, as_offset=True):
-    if _search_regex.search(date_string):
+    if _search_regex_ignorecase.search(date_string):
         for name, info in _tz_offsets:
             timezone_re = info['regex']
             timezone_match = timezone_re.search(date_string)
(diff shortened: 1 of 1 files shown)


lopuhin:
Tests run quite a bit faster after this change (e.g. ``tests/test_date_parser.py`` went from 1.7s to 1.5s, ``tests/test_search.py`` went from 2.2s to 1.9s), also I wrote a little benchmark specifically for this PR:

```
$ cat t.py 
import time
import random
import dateparser

def main():
    inputs_no_tz = []
    inputs_tz = []
    random.seed(42)
    for _ in range(1000):
        day = random.randint(10, 31)
        year = random.randint(1970, 2030)
        hour = random.randint(12, 23)
        date = f'{year}.12.{day} {hour}:45:00'
        inputs_no_tz.append(date)
        inputs_tz.append(date + ' EST')

    t0 = time.perf_counter()
    for x in inputs_tz:
        assert dateparser.parse(x)
    t_tz = (time.perf_counter() - t0) / len(inputs_tz)

    t0 = time.perf_counter()
    for x in inputs_no_tz:
        assert dateparser.parse(x)
    t_no_tz = (time.perf_counter() - t0) / len(inputs_no_tz)

    print(f'Time with tz: {t_tz:.6f} s, t_no_tz: {t_no_tz:.6f} s')


if __name__ == '__main__':
    main()
```

And here are results (on branch first):

```
$ python t.py
Time with tz: 0.001375 s, t_no_tz: 0.000579 s
$ python t.py
Time with tz: 0.001363 s, t_no_tz: 0.000578 s
$ git checkout master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
$ python t.py
Time with tz: 0.001353 s, t_no_tz: 0.001223 s
$ python t.py
Time with tz: 0.001344 s, t_no_tz: 0.001216 s
```
so parsing without TZ is more than 2x faster, while with TZ it's a tiny bit smaller. It should also be possible to bring the time of TZ parsing down.

lopuhin:
@asadurski do you think it's fine to merge this PR as it is now (big speedup for dates without TZ, small slowdown for dates with TZ), or it's better to improve the case with TZ as well? It should be possible to eliminate the loop from ``pop_tz_offset_from_string`` entirely, but it would be a bit more changes than now.

lopuhin:
I also checked import time, and it's affected a little as well:
```
# PR
$ for i in $(seq 10); do python -c "import time; t0 = time.perf_counter(); import dateparser; print(f'{(time.perf_counter() - t0):.3f} s')"; done
0.126 s
0.128 s
0.129 s
0.127 s
0.129 s
0.126 s
0.130 s
0.127 s
0.126 s
0.126 s
# master
$ for i in $(seq 10); do python -c "import time; t0 = time.perf_counter(); import dateparser; print(f'{(time.perf_counter() - t0):.3f} s')"; done
0.114 s
0.115 s
0.112 s
0.112 s
0.112 s
0.112 s
0.111 s
0.112 s
0.112 s
0.111 s
```

lopuhin:
> I also checked import time, and it's affected a little as well:

It should be possible to reverse this though if we lazy-init stuff used by timezone parser.

lopuhin:
Thank you @asadurski 👍 

> And I'll look some more into performance benchmark options with Travis.

I think even a benchmark suite which can be run locally to compare with master would be very useful, and there are plenty of nice cases in the tests which could be used as benchmark input.

PR #572, opened 2019-10-05
Fix milliseconds parsing

d10xa:
```python
print(dateparser.date.get_date_from_timestamp(u'1570308760263', None).microsecond)
```

```
0
```


tests/test_date.py
```diff
@@ -681,5 +681,16 @@ def then_date_object_is_invalid(self):
...
 
+class TestTimestampParser(BaseTestCase):
+    def setUp(self):
+        super(TestTimestampParser, self).setUp()
+
+    def test_timestamp_millis(self):
+        self.assertEqual(
+            date.get_date_from_timestamp(u'1570308760263', None),
+            datetime(2019, 10, 5, 23, 52, 40, 263)
+        )
```
Gallaecio: Your test is currently failing. I wonder if it has to do with the fact that you are passing milliseconds as the `microsecond` parameter in your implementation.

And you should also cover cases where the number has a different number of digits: 9, 11, 12, 14. I assume in those case the resulting date would be `None`. Maybe 16 digits (microseconds) should be supported as well.

Also, `setUp` can be removed from your test class.

Gallaecio:
I think you may be over complicating your implementation. I think it may be better to modify `RE_SEARCH_TIMESTAMP` to match the timestamp with capture groups for the different relevant numbers here (e.g. `r'^(\d{10})(\d{3})?(\d{3})?$'`), and work with the captured groups.

As for tests, you should be able to simply extend the existing `test_parse_timestamp` test with new inputs and outputs.

dateparser/date.py
```diff
@@ -47,7 +47,7 @@
 RE_SANITIZE_ON = re.compile(r'^.*?on:\s+(.*)')
 RE_SANITIZE_APOSTROPHE = re.compile(u'|'.join(APOSTROPHE_LOOK_ALIKE_CHARS))
 
-RE_SEARCH_TIMESTAMP = re.compile(r'^\d{10}(?![^\d.])')
+RE_SEARCH_TIMESTAMP = re.compile(r'^(\d{10})(\d{3})?(\d{3})?$')
```
Gallaecio: I guess we should allow periods after the number, to match the intent of the original regular expression:

```suggestion
RE_SEARCH_TIMESTAMP = re.compile(r'^(\d{10})(\d{3})?(\d{3})?(?![^.])')
```

Could you add a test for it as well? (having a period after the timestamp)

dateparser/date.py
```diff
@@ -122,9 +122,15 @@ def sanitize_date(date_string):
...
-        date_obj = datetime.fromtimestamp(int(date_string[:10]))
-        date_obj = apply_timezone_from_settings(date_obj, settings)
+    match = RE_SEARCH_TIMESTAMP.search(date_string)
+    if match:
+        seconds = int(match.group(1))
+        millis = int(match.group(2) or 0)
+        micros = int(match.group(3) or 0)
+        date_obj = datetime.fromtimestamp(seconds)
+        date_obj = date_obj.replace(microsecond=millis * 1000 + micros)
+        if date_obj:
```
Gallaecio: I am puzzled by why this `if` statement is needed, although according to the code coverage report it is indeed visited. What happens if you remove it, which tests fail?

Gallaecio:
Please, mind https://github.com/scrapinghub/dateparser/pull/572/files#r338457223

PR #577, opened 2019-10-21
Update fr.json

AmenEllah:


asadurski:
Thanks a lot for your effort, @AmenEllah.
However, when you include tests for your changes (that's also a requirement for your contribution to be merged), they will most likely fail, because editing `.json` files ― which are only one of the sources originated from CLDR database ― doesn't have any direct impact on `.py` files used with the parsers.
Instead you should edit `.yaml` files from https://github.com/scrapinghub/dateparser/tree/master/dateparser_data/supplementary_language_data/date_translation_data and run the script https://github.com/scrapinghub/dateparser/blob/master/scripts/write_complete_data.py after that.
Please refer to recent comment https://github.com/scrapinghub/dateparser/pull/592#issuecomment-556343811 and to https://dateparser.readthedocs.io/en/latest/contributing.html#guidelines-for-editing-translation-data for more details.
If those comments are not enough, I'll be happy to help.

PR #585, opened 2019-11-07
fixbug/587-return-none-when-pasre-indonesia-date-string

### LINKED ISSUE #584: Return None when parser Indonesia date string
### Problem
- Input : `6 Nop 2019`
- Output: return `None`

toixtran:
Fixes #584


toixtran:
For isssue: #585

Gallaecio:
From recent similar pull requests, I gather that editing the JSON files directly is not the way dateparser works. You need to edit some YAML files that are then used to generate the JSON files, which merge data from alternative sources (ICU).

PR #591, opened 2019-11-17
Utc test fix

Watchful1:
Some tests that use "today()" to get the current date fail when not run on a utc based system. This replaces those references with "utcnow()".


noviluni:
thanks @Watchful1 !

PR #592, opened 2019-11-19
add 'days ago' for norwegian

hakanw:
"2 days ago" in norwegian didn't work.

### DIFF
--- a/dateparser/timezone_parser.py
+++ b/dateparser/timezone_parser.py
@@ -33,15 +33,17 @@ class StaticTzInfo(tzinfo):
 
 
 def pop_tz_offset_from_string(date_string, as_offset=True):
-    for name, info in _tz_offsets:
-        timezone_re = info['regex']
-        timezone_match = timezone_re.search(date_string)
-        if timezone_match:
-            start, stop = timezone_match.span()
-            date_string = date_string[:start + 1] + date_string[stop:]
-            return date_string, StaticTzInfo(name, info['offset']) if as_offset else name
-    else:
-        return date_string, None
+    if _search_regex_ignorecase.search(date_string):
+        for name, info in _tz_offsets:
+            timezone_re = info['regex']
+            timezone_match = timezone_re.search(date_string)
+            if timezone_match:
+                start, stop = timezone_match.span()
+                date_string = date_string[:start + 1] + date_string[stop:]
+                return (
+                    date_string,
+                    StaticTzInfo(name, info['offset']) if as_offset else name)
+    return date_string, None
 
 
 def word_is_tz(word):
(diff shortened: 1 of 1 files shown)


dateparser/data/date_translation_data/nb.py
```diff
@@ -93,11 +93,13 @@
     ],
     "week": [
         "uke",
-        "u"
+        "u",
+        "uker"
```
Gallaecio: :lipstick: I know this is not currently the case, but it may be a good idea to take this opportunity to include a comma in the last item as well, and maybe even sort the entries alphabetically. Same for the rest of the changes.

```suggestion
        "uker",
```

hakanw:
All your comments done! Please have a look and merge if it looks OK

dateparser/data/date_translation_data/nb.py
```diff
@@ -92,12 +92,14 @@
...
+        "u",
         "uke",
-        "u"
+        "uker",
     ],
     "day": [
+        "d",
         "dag",
-        "d"
+        "dager"
```
Gallaecio: :lipstick: 
```suggestion
        "dager",
```

hakanw:
Happy with the fixes? :)

asadurski:
Thank you for making this contribution @hakanw!
There's a small problem though, language data `.py` files are created automatically and should not be edited.
Instead, you should have created `no.yml` with necessary changes here: https://github.com/scrapinghub/dateparser/tree/master/dateparser_data/supplementary_language_data/date_translation_data, and then run the script https://github.com/scrapinghub/dateparser/blob/master/scripts/write_complete_data.py, which would have made necessary changes to dateparser/data/date_translation_data/nb.py.
The reason for this is that the next time the CLDR data gets updated and new files are generated based on that, those valuable changes to `.py` files will be lost.
That's described in https://dateparser.readthedocs.io/en/latest/contributing.html#guidelines-for-editing-translation-data. If you feel that it's not enough or it's not comprehensible, please feed me back.
Hang on, you are almost there!

hakanw:
Thanks @asadurski.  Is editing the JSON file not enough (since there already exists a `nb.json`and no `nb.yaml` since previously) or do I have to reverse engineering the yaml format and port all of the existing `nb.py` to it?

hakanw:
@asadurski I have now tried to create a .yaml file from the existing `nb.py`. See my commit; am I on the right track? For some reason I get double lines in `nb.py` now after running the script... but would be good to get a confirmation from you that I'm doing the right thing here.

hakanw:
@asadurski ping?

asadurski:
Thanks for pinging, @hakanw. Can you elaborate on double lines in `nb.py`?
Essentially, `nb.json` should not be touched, as it's a resource file. Please revert the changes there and run the `write_complete_data.py` script again.

PR #299, opened 2017-03-28
Added numbers upto twelve (fix for #291)

atultherajput:
I had added numbers upto twelve in languagefiles/en.yaml 
This is a fix for #291 
Please review this patch @waqasshabbir and @asadurski

PR #593, opened 2019-11-20
Support "one"

Watchful1:
When pull request #299 was merged [here](https://github.com/scrapinghub/dateparser/commit/11a761c99d3ee522a3c63756b70c106a579e8b5c), the number for "one" was somehow lost. So parsing "two weeks" works, but "one week" doesn't.


noviluni:
Thank you very much @Watchful1 !!!

PR #595, opened 2019-12-10
Allow datetime.parse("3 seconds") to respect PREFER_DATES_FROM

### LINKED ISSUE #325: Relative dates in FreshnessDateDataParser
If I want to parse relative dates that don't include the word `in` it will always return a date in the past, which it's not ideal and it should take into account the `PREFER_DATES_FROM` setting.
Example:
```
In [2]: datetime.now()
Out[2]: datetime.datetime(2017, 6, 14, 18, 1, 4, 654628)
In [5]: dateparser.parse('1 month', settings={'PREFER_DATES_FROM': 'future'})
Out[5]: datetime.datetime(2017, 5, 14, 18, 1, 26, 815558)
```

swingtempo:
**Previous behavior:**
If the caller of datetime.parse("3 seconds") sets PREFER_DATES_FROM="future", it would be ignored.

e.g. datetime.parse("3 seconds", settings={'PREFER_DATES_FROM="future"}) would return a time 3 seconds ago

**After this change:**
This code path will work now. Previously, if the date did not contain the word "in", then a parsing of a date will _always_ return a past time. However, it makes sense that if PREFER_DATES_FROM="future", we should send the future date.

datetime.parse("3 seconds", settings={'PREFER_DATES_FROM="future"}) would return a time 3 seconds in the future

Fixes #325


swingtempo:
Updated with tests. `test_relative_future_dates_via_settings` which got many of its test cases from `test_relative_future_dates`

Gallaecio:
@swingtempo Did you close this on purpose?

swingtempo:
accident!! sorry!

dateparser/date.py
```diff
@@ -326,7 +326,7 @@ class DateDataParser(object):
     locale_loader = None
 
     @apply_settings
-    def __init__(self, languages=None, locales=None, region=None, try_previous_locales=True,
+    def __init__(self, languages=None, locales=None, region=None, try_previous_locales=False, #jayongg updated
```
Gallaecio: Why did you add this now? I’m a bit confused. At first sight it looks to me like an unrelated, backward-incompatible change.

msftjayongg: So Sorry!!! I was experimenting and thought I was in a branch.  Let me revert it now

noviluni:
Closed in favor of https://github.com/scrapinghub/dateparser/pull/530, as it is older and it considers when "`ago`" is present. 

Notice that in this PR, this: `parse("1 min ago", settings={"PREFER_DATES_FROM" : "future"})` would fail, as would be returning a "future" date.

Anyway, thank you very much for your efforts, @swingtempo :)

PR #596, opened 2019-12-10
Resolve DeprecationWarning related to raw string escape sequences

mshriver:
Just trying to clean up some Deprecation Warnings that I'm seeing in use of the library with python 3.7.

Looking at contributions guidelines - I didn't find any open issues related to this deprecation warning, and there's no functional change so I have not added any additional tests.


mshriver:
Thanks @Gallaecio !

PR #311, opened 2017-05-02
conversion improvements of russian human dates

avostryakov:

PR #599, opened 2019-12-13
Add во as a synonym of в for Russian

Gallaecio:
Implements one of the improvements from #311.

I don’t speak Russian, so please do check that I did not make a language mistake.

Regarding tests, I might have gone overboard, specially in `tests/test_search.py`.

Finally, I did not edit a JSON file because I recall it being automatically generated. However, I did not find information at https://dateparser.readthedocs.io/en/latest/contributing.html#guidelines-for-editing-translation-data on how to rebuild it. Also, I am not sure if `ru.py` is meant to be modified like I did.


kmike:
`в` and `во` have the same meaning, but are not interchangeable, so unfortunately all currently added examples in this PR are ungrammatical. 

The only possible usage I can think of is "во вторник" (Tuesday); "в вторник" would be incorrect. In all other cases `в` should be used, if I'm not mistaken.

Issue #397, opened 2018-03-16
Necessary documentation improvements


asadurski:
- extensive entry for `search_dates()`
- extend list of supported languages and locales

TBC

rennerocha:
I believe that removing references to Scrapy Cloud is interesting too (https://dateparser.readthedocs.io/en/latest/installation.html#deploying-dateparser-in-a-scrapy-cloud-project), as the library is general enough and not directly related to Scrapy.

Gallaecio:
>     * extensive entry for `search_dates()`

Could you elaborate? What do we need to cover beyond what is already covered in https://dateparser.readthedocs.io/en/latest/#search-for-dates-in-longer-chunks-of-text?

lucasdutraf:
> I believe that removing references to Scrapy Cloud is interesting too (https://dateparser.readthedocs.io/en/latest/installation.html#deploying-dateparser-in-a-scrapy-cloud-project), as the library is general enough and not directly related to Scrapy.

Is this change still needed, @rennerocha ? I could work on it

noviluni:
I have just created a ticket to improve the `search_dates` section in the docs: https://github.com/scrapinghub/dateparser/issues/737

As we currently have a list of locales (it is in the README, I think we should move it to the docs, but that's another point) and we have an open ticket regarding the section about Scrapy Cloud in the docs: https://github.com/scrapinghub/dateparser/issues/724, I think we can close this ticket, as it's too broad and incomplete.

Thank you, guys!

PR #601, opened 2019-12-13
 Add a script to update the README table of supported languages and locales

Gallaecio:
Partial fix for #397.

Although sorted, the table contents are the same. Should the script be feeding additional languages and locales from sources other than `dateparser.data.languages_info.language_locale_dict`?


scripts/update_supported_languages_and_locales.py
```diff
@@ -0,0 +1,48 @@
...
+        result += '\n'
+    return result
+
+
+def main():
+    readme_path = os.path.join(os.path.dirname(__file__), '..', 'README.rst')
+    new_data = to_string(language_locale_dict)
+    temporary_file = TemporaryFile('w+')
+    with open(readme_path) as readme_file:
+        delimiter = '============    ================================================================\n'
```
noviluni, line 29: what about refactoring this to something like this?:
```suggestion
        delimiter = '='*12+'    '+'='*64+'\n'
```

noviluni:
I like the idea, but I don't think adding such a big table of supported locales to the README has any sense, so I will keep this PR open until we finish this: https://github.com/scrapinghub/dateparser/issues/723 and then we can work on this, to see if we want to move that table to the docs or whatever.

noviluni:
Hi @Gallaecio!

This work would help a lot when updating the CLDR data: https://github.com/scrapinghub/dateparser/issues/826

The table was moved to the docs (supported_locales.rst). Could you update the code to work with it?

Regarding the source, I think it's OK to extract the language list from `dateparser.data.languages_info.language_locale_dict`. We could also get the list of supported locales from the modules in `dateparser.data.date_translation_data`, while the derived locales (locale + region) are in the `locale_specific` key, but I think that this is too tricky and it should be the same.

CONTRIBUTING.rst
```diff
@@ -160,3 +160,16 @@ language, you must:
...
+----------------------------------------------------
+
+Whenever the content of
+``dateparser.data.languages_info.language_locale_dict`` is modified, use
+``dateparser_scripts/update_supported_languages_and_locales.py`` to update 
+the corresponding documentation table::
+
+    python3 -m venv venv
+    . venv/bin/activate
+    pip install -e .
```
noviluni, line 174: do we need to add these three lines? I think it's too verbose and it shouldn't be necessary. @Gallaecio

noviluni (approved):
Great! :muscle:

CONTRIBUTING.rst
noviluni: @Gallaecio 
In order to merge it, let me know if you change this or the reasons if you leave it :)

PR #603, opened 2019-12-17
Implement a PARSERS setting

### LINKED ISSUE #312: Disabling relative times
Hi all,

I am very happy to use dateparser. It is very successful for me in parsing a lot of date formats.
I have a use case however where relative dates do not make sense. Therefore I would like to disable this functionality, limiting the parsing to absolute times only.

I am happy to contribute a patch, if you are interested and if can point me in the right direction.

Cheers,
                 Johannes

### LINKED ISSUE #525: can be prevent searching relative dates
I am looking for dates only in my string like 
"25th march 2015 , 26th july 2018 , i need this report today."

I need only dates when search using search.search_dates()
but it also showing today.

Is any way to prevent this.

Gallaecio:
Fixes #312, fixes #525


noviluni:
I thinks it is a really good idea. Moreover, it is necessary to implement this in some way to allow showing what parser has been used when implementing the `DEBUG_INFO` flag (draft PR: https://github.com/scrapinghub/dateparser/pull/561/files)

noviluni:
@Gallaecio

Gallaecio:
Added the list of parsers to the documentation with a brief description about each one of them.

dateparser/date.py
```diff
@@ -182,21 +182,22 @@ def __init__(self, locale, date_string, date_formats, settings=None):
...
-        for parser in (
-            self._try_timestamp,
-            self._try_freshness_parser,
-            self._try_given_formats,
-            self._try_parser,
-            self._try_hardcoded_formats,
-        ):
-            date_obj = parser()
+        for parser in self._settings.PARSERS:
+            date_obj = self._parsers[parser]()
```
noviluni: If we add a non existent parser to the settings this will raise `KeyError`. 

I think that adding a warn like this:
```python
    def _parse(self):
        for parser in self._settings.PARSERS:
            parser_fnx = self._parsers.get(parser)
            if not parser_fnx:
                warn('"{}" is not a valid parser.'.format(parser))
                continue
            date_obj = parser_fnx()
            if self._is_valid_date_obj(date_obj):
                return date_obj
        else:
            return None
```
would do the trick. To test it, you can change the `PARSER` settings to something like this: `'PARSERS': ['aaaaa'] + default_parsers,`.

What do you think @Gallaecio ?

Feel free to adapt the warn message, as maybe we could mention the settings or whatever, and if you want, it could be a good idea to add a test with wrong parsers to see that all works as expected.

docs/usage.rst
```diff
@@ -146,6 +146,45 @@ Defaults to `False`.
...
+    are interpreted as `Unix time <https://en.wikipedia.org/wiki/Unix_time>`_.
+
+-   ``'relative-time'``: Parses dates and times expressed in relation to the
+    current date and time (e.g. “1 day ago”, “in 2 weeks”).
+
+-   ``'custom-formats'``: Parses dates that match one of the date formats in
+    the list of the ``date_formats`` parameter of :func:`dateparser.parse` or
+    :meth:`DateDataParser.get_date_data
+    <dateparser.date.DateDataParser.get_date_data>`.
+
```
noviluni: Good job here!!

tests/test_search.py
```diff
@@ -274,6 +275,14 @@ def test_search_date_string(self, shortname, datetime_string):
...
                   datetime.datetime(2018, 8, 6, 17, 5, tzinfo=StaticTzInfo('CDT', datetime.timedelta(seconds=-18000))))],
               settings={'RELATIVE_BASE': datetime.datetime(2000, 1, 1)}),
+        param('en', '25th march 2015 , i need this report today.',
+              [('25th march 2015', datetime.datetime(2015, 3, 25))],
+              settings={'PARSERS': [parser for parser in default_parsers
+                                    if parser != 'relative-time']}),
+        param('en', '25th march 2015 , i need this report today.',
+              [('25th march 2015', datetime.datetime(2015, 3, 25)),
+               ('today', datetime.datetime(2000, 1, 1))],
+              settings={'RELATIVE_BASE': datetime.datetime(2000, 1, 1)}),
```
noviluni: as mentioned before, it could be a good idea to add a test with non existing parsers.

dateparser/date.py
noviluni: Of course, if you prefer the "easier to ask for forgiveness than permission" version, you can go with:

```python
    def _parse(self):
        for parser in self._settings.PARSERS:
            try:
                date_obj = self._parsers[parser]()
            except KeyError:
                warn('"{}" is not a valid parser.'.format(parser))
                continue
            date_obj = parser_fnx()
            if self._is_valid_date_obj(date_obj):
                return date_obj
        else:
            return None
```

However, this would catch any `KeyError` occurring inside the parser function and we could generate false warns.

Gallaecio: I’m thinking that it may be better to raise an exception, instead of logging a warning. Warnings are good for deprecation, when something will stop working in the future. But in this case, I cannot think of a good reason why a user would pass the wrong parser, so I think we should raise an exception.

That said, it would be great to raise a `ValueError` with a message in line with the one you propose for the warning. I’ll update the changes accordingly and include a test as you suggested below.

PR #605, opened 2019-12-17
Skip locale-dependent tests if the required locale is not installed

Gallaecio:


Gallaecio:
I see your point, but to skip the required French locale in Travis CI we would have to remove line https://github.com/scrapinghub/dateparser/blob/master/.travis.yml#L30 so I’m not really concerned about such an scenario.

asadurski:
Good one!
We should probably add similar test for locale that's further away from English (not as part of this PR though).

Issue #607, opened 2020-01-10
dateparser.parse('02/2013', ['%m/%Y']) gives wrong datetime


mcarans:
dateparser.parse('02/2013', ['%m/%Y']) gives datetime(2013, 2, 28, 0, 0) instead of datetime(2013, 2, 1, 0, 0)

datetime.strptime('02/2013', '%m/%Y') gives datetime(2013, 2, 1, 0, 0)

noviluni:
In this moments `dateparser` returns the last day of the month for this kind of formats. I think the reason behind that decission is that when finding only a month and a year we usually refer to the last day of that month (i.e. expiration date).

HOWEVER that is a bug/inconsistency, because there is a setting `PREFER_DAY_OF_MONTH` that should be followed, as when we just write `dateparser.parse('February 2013')`.

Examples:

```
# It works

dateparser.parse('February 2013', settings={'PREFER_DAY_OF_MONTH': 'current'})                                                                                                                          
>>> datetime.datetime(2013, 2, 17, 0, 0)

dateparser.parse('February 2013', settings={'PREFER_DAY_OF_MONTH': 'first'})                                                                                                                            
>>> datetime.datetime(2013, 2, 1, 0, 0)

dateparser.parse('February 2013', settings={'PREFER_DAY_OF_MONTH': 'last'})                                                                                                                             
>>> datetime.datetime(2013, 2, 28, 0, 0)

# It doesn't work

dateparser.parse('02/2013', settings={'PREFER_DAY_OF_MONTH': 'current'})                                                                                                                          
>>> datetime.datetime(2013, 2, 28, 0, 0)

dateparser.parse('02/2013', settings={'PREFER_DAY_OF_MONTH': 'first'})                                                                                                                            
>>> datetime.datetime(2013, 2, 28, 0, 0)

dateparser.parse('02/2013', settings={'PREFER_DAY_OF_MONTH': 'last'})                                                                                                                             
>>> datetime.datetime(2013, 2, 28, 0, 0)


```

I added a PR with a patch.

jbcooley7:
I've just run across this issue too. Supposing the PR gets final approval and is merged, is there a projected release date at which this fix will go out? I saw the most recent release was September 2019, but prior releases seemed to be on a nearly yearly cadence around the first months of the year.

noviluni:
Hi @jbcooley7 this has been fixed in the last version `0.7.4` :tada: 

We will try to start doing more frequent releases. 

Cheers!

PR #611, opened 2020-01-17
apply PREFER_DAY_OF_MONTH in date.py

noviluni:
fixes https://github.com/scrapinghub/dateparser/issues/607

### DIFF
--- a/dateparser/parser.py
+++ b/dateparser/parser.py
@@ -7,7 +7,8 @@ from collections import OrderedDict
 from datetime import datetime
 from datetime import timedelta
 
-from dateparser.utils import set_correct_day_from_settings
+from dateparser.utils import set_correct_day_from_settings, \
+    get_last_day_of_month
 from dateparser.utils.strptime import strptime
 
 
@@ -302,8 +303,7 @@ class _parser(object):
                 (error_msgs[0] in error_text or error_msgs[1] in error_text) and
                 not(self._token_day or hasattr(self, '_token_weekday'))
             ):
-                _, tail = calendar.monthrange(params['year'], params['month'])
-                params['day'] = tail
+                params['day'] = get_last_day_of_month(params['year'], params['month'])
                 return datetime(**params)
             else:
                 raise e
@@ -430,10 +430,8 @@ class _parser(object):
         ):
             return dateobj
 
-        _, tail = calendar.monthrange(dateobj.year, dateobj.month)
-
         dateobj = set_correct_day_from_settings(
-            dateobj, self.settings, last_day=tail, current_day=self.now.day
+            dateobj, self.settings, current_day=self.now.day
         )
         return dateobj
 
--- a/dateparser/parser.py
+++ b/dateparser/parser.py
@@ -7,6 +7,7 @@ from collections import OrderedDict
 from datetime import datetime
 from datetime import timedelta
 
+from dateparser.utils import set_correct_day_from_settings
 from dateparser.utils.strptime import strptime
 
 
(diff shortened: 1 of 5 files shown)


noviluni:
(Explanation and example in a comment in the PR)

Gallaecio:
What I found most strange about #607 was that, depending on whether the date format input parameter was a list (`['%m/%Y']`) or a string (`'%m/%Y'`), the output date was different.

I think that is something we need to address and cover in the tests before we can consider #607 solved.

noviluni:
@Gallaecio 
I think there was a confusion.

Currently, doing this:
```python
dateparser.parse('02/2013', ['%m/%Y'])                                                                                                                                                                  
>>> datetime.datetime(2013, 2, 28, 0, 0)
```
Is the same than doing this:
```python
>>> dateparser.parse('02/2013', '%m/%Y')                                                                                                                                                                    
/home/marc/workprojects/dateparser/dateparser/date.py:142: FutureWarning: Date formats should be list, tuple or set of strings
  warn(_DateLocaleParser.DATE_FORMATS_ERROR_MESSAGE, FutureWarning)
datetime.datetime(2013, 2, 28, 0, 0)
```
(except for the deprecation warning). In the example in the issue, the second statement is
```
datetime.strptime('02/2013', '%m/%Y')
``` 
which is using the `datetime.strptime`, not the `dateparse.parse` method.

Gallaecio:
:man_facepalming:

dateparser/utils/__init__.py
```diff
@@ -133,6 +135,24 @@ def apply_timezone_from_settings(date_obj, settings):
     return date_obj
 
 
+def get_last_day_of_month(year, month):
+    return calendar.monthrange(year, month)[1]
+
+
+def set_correct_day_from_settings(date_obj, settings, last_day=None, current_day=None):
```
Gallaecio: :lipstick: What about `apply_day_of_the_month_from_settings`, to be consistent with `apply_timezone_from_settings`?

dateparser/parser.py
```diff
@@ -430,16 +431,11 @@ def _correct_for_day(self, dateobj):
             return dateobj
 
         _, tail = calendar.monthrange(dateobj.year, dateobj.month)
```
noviluni: I have just seen that we don't really need to generate here the `tail` as can be directly generated in the `set_correct_day_from_settings`. I will change it when I have a moment.

dateparser/utils/__init__.py
```diff
@@ -133,6 +135,24 @@ def apply_timezone_from_settings(date_obj, settings):
...
+    """ Set correct day attending the `PREFER_DAY_OF_MONTH` setting."""
```
noviluni: any better idea for the naming of this function?

Gallaecio: https://github.com/scrapinghub/dateparser/pull/611#discussion_r368165205

noviluni:
Note to self: I could also add a unit test for the function `get_last_day_of_month` (including all months and at least one example of leap year).

dateparser/utils/__init__.py
noviluni: Hmmm... This was my first approach, but then I didn't like that naming. First of all "apply" seems to mean that you are "adding" something. It maybe has sense for the timezone, as sometimes you add it, but I'm not sure if this has sense for the day as what you are doing is to change it.

Secondly, `apply_day_of_the_month_from_settings` seems to mean that it is getting the day from the settings; but what it is doing is using the rules in the settings to find the correct day.

For both reasons I prefer this naming: `set_correct_day_from_settings`.

What do you think @Gallaecio ? Any better idea?

PR #616, opened 2020-02-13
fix self.filtered_tokens bug that cause an exception for some characters in datestring

emadkha:
I use jalali parser for parse some jalali datetime string in my crawler.  it throws an exception for this string: "۱۳۹۸/۰۶/۳۱ در  ۱۴:۵۲"  because of "در" word. in line 192 of dateparser/parser.py , when use t[1] < 1 , for that word, tokens contain ('د','') and ('ر','') items and for these items, t[1] is str , not int. then it cause an exception. 'در' == "at"

### DIFF
--- a/dateparser/parser.py
+++ b/dateparser/parser.py
@@ -189,7 +189,7 @@ class _parser(object):
     def __init__(self, tokens, settings):
         self.settings = settings
         self.tokens = list(tokens)
-        self.filtered_tokens = [t for t in self.tokens if t[1] <= 1]
+        self.filtered_tokens = [t for t in self.tokens if isinstance(t[1],int) and t[1] <= 1]
 
         self.unset_tokens = []
 


Gallaecio:
This probably affects `_no_spaces_parser` as well, so the fix should be either in `tokenizer` or in both `_parser` and `_no_spaces_parser`. Can you have a look at `tokenizer`, see if you can improve it to support this scenario?

It would also be great to have tests for the fix.

noviluni:
Hi @emadkha, thank you for raising this!

The `TypeError` you mention was fixed in this PR: https://github.com/scrapinghub/dateparser/pull/622 and it was released within the new version `0.7.4`. :tada: 

Now you can use:
```
>>> from dateparser.calendars.jalali import JalaliCalendar
>>> JalaliCalendar("۱۳۹۸/۰۶/۳۱ در ۱۴:۵۲").get_date()
{'date_obj': datetime.datetime(2019, 9, 22, 14, 52), 'period': 'day'}
```
And it works as expected. I will close this PR but feel free to open it again if your problem persists! 

Thanks

<hr>
P.S: I'm trying to implement a new `jalali` parser to allow us to do this directly:

```
>>> dateparser.parse("۱۳۹۸/۰۶/۳۱ در ۱۴:۵۲", settings={'PARSERS': ['jalali']})  
datetime.datetime(2019, 9, 22, 14, 52)
```

However, it doesn't work for now. This is the PR: https://github.com/scrapinghub/dateparser/pull/627 if you want to take a look.

Issue #621, opened 2020-02-20
TypeError: '<=' not supported between instances of 'str' and 'int'


lopuhin:
``dateparser.parse`` crashes when trying to parse some dates, for example:

```
>>> dateparser.parse('Oct 1 2018 4:40 PM EST —')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "dateparser/conf.py", line 82, in wrapper
    return f(*args, **kwargs)
  File "dateparser/__init__.py", line 53, in parse
    data = parser.get_date_data(date_string, date_formats)
  File "dateparser/date.py", line 417, in get_date_data
    locale, date_string, date_formats, settings=self._settings)
  File "dateparser/date.py", line 189, in parse
    return instance._parse()
  File "dateparser/date.py", line 199, in _parse
    date_obj = parser()
  File "dateparser/date.py", line 222, in _try_parser
    self._get_translated_date(), settings=self._settings)
  File "dateparser/conf.py", line 82, in wrapper
    return f(*args, **kwargs)
  File "dateparser/date_parser.py", line 26, in parse
    date_obj, period = parse(date_string, settings=settings)
  File "dateparser/parser.py", line 70, in parse
    raise exceptions.pop(-1)
  File "dateparser/parser.py", line 64, in parse
    res = parser(datestring, settings)
  File "dateparser/parser.py", line 446, in parse
    po = cls(tokens.tokenize(), settings)
  File "dateparser/parser.py", line 192, in __init__
    self.filtered_tokens = [t for t in self.tokens if t[1] <= 1]
  File "dateparser/parser.py", line 192, in <listcomp>
    self.filtered_tokens = [t for t in self.tokens if t[1] <= 1]
TypeError: '<=' not supported between instances of 'str' and 'int'
```

This is using dateparser master (2dd4c5962e16de208c6ef7a9a9f9d5efe8754eb7)

noviluni:
It seems that the dash and the long dash weren't included in the `nowords` variable. I added a PR to fix it. Thanks for raising it, @lopuhin !

noviluni:
Hi @lopuhin the last release includes this fix :tada:

lopuhin:
Great, thank you so much @noviluni 👍

PR #622, opened 2020-02-21
fix tokenizer for non recognized characters

noviluni:
fixes: https://github.com/scrapinghub/dateparser/issues/621

### DIFF
--- a/dateparser/parser.py
+++ b/dateparser/parser.py
@@ -523,13 +523,12 @@ class _parser(object):
 class tokenizer(object):
     digits = u'0123456789:'
     letters = u'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ'
-    nonwords = u"./\()\"',.;<>~!@#$%^&*|+=[]{}`~?-—–     "
 
-    def _isletter(self, tkn): return tkn in self.letters
+    def _isletter(self, tkn):
+        return tkn in self.letters
 
-    def _isdigit(self, tkn): return tkn in self.digits
-
-    def _isnonword(self, tkn): return tkn in self.nonwords
+    def _isdigit(self, tkn):
+        return tkn in self.digits
 
     def __init__(self, ds):
         self.instream = StringIO(ds)
@@ -538,13 +537,11 @@ class tokenizer(object):
         if self._isdigit(chara):
             return 0, not self._isdigit(charb)
 
-        if self._isletter(chara):
+        elif self._isletter(chara):
             return 1, not self._isletter(charb)
 
-        if self._isnonword(chara):
-            return 2, not self._isnonword(charb)
-
-        return 3, True
+        else:
+            return 2, self._isdigit(charb) or self._isletter(charb)
 
     def tokenize(self):
         token = ''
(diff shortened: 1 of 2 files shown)


dateparser/parser.py
```diff
@@ -524,7 +524,7 @@ def parse_alpha(token, skip_component=None):
 class tokenizer(object):
     digits = u'0123456789:'
     letters = u'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ'
-    nonwords = u"./\()\"',.;<>~!@#$%^&*|+=[]{}`~?-     "
+    nonwords = u"./\()\"',.;<>~!@#$%^&*|+=[]{}`~?-—–     "
```
lopuhin: Thanks @noviluni ! As I understand, this still won't fix cases with other unexpected symbols, e.g. ``dateparser.parse('Oct 1 2018 4:40 PM EST 😊')`` would still fail. I wonder if it's possible to handle "unknown" symbols in another place? But maybe it's a bad idea, I'm not familiar with this part of the code.

noviluni: Hi @lopuhin 

My PR was to fix the problem as soon as possible, as a dash is a common sign, but I have noted this down to investigate it further.

It seems that the tokenizer sepparate between "words" (returns `1`), "digits" (returns `0`, it can include `:`, etc.) and "no-words" (returns `2`) and, after that, it selects the valid tokens by discarding the "no words". However, those unexpected symbols return just an empty string and that breaks the filtering because it compares with 1 and gives the error you had. Fixing this should be as easy as changing this:

```
self.filtered_tokens = [t for t in self.tokens if t[1] <= 1]
```
by this:
```
self.filtered_tokens = [t for t in self.tokens if t[1] != '' and t[1] <= 1]
```

But I'm not sure of the implications and why it was decided to be implemented in this way...

I will do some more tests and push a "fixed" version.

lopuhin: > My PR was to fix the problem as soon as possible, as a dash is a common sign, but I have noted this down to investigate it further.

This makes sense, thank you @noviluni :+1:

I checked the code and I think this fix with ``filtered_tokens`` looks good and should not break anything it seems :+1: 

I was also thinking why was this code written like this in the first place, and maybe I have an idea: in python 2 it's not an error to compare an empty string with an int, and ``1 < ''`` is True - so maybe the idea behind a ``''`` default in ``_switch`` method was to have a value which would be greater than any number that would be added next. If this logic is correct, then indeed your fix would be good, and another potential fix would be to default to a large number in ``_switch``.

noviluni: Oh, right, Python 2... It was only 1.5 months ago when it died and I've already forgotten...

If that's the case Instead of filtering the tokens it would be better to return just `3` by default (I mean, for unrecognized characters), as returning the same datatype seems easier to mantain / follow and we will avoid comparing all the valid cases with `""` in the `filtered_tokens` (better performance). However, I think we don't need the `nonwords` list at all, as we could just return 2 when it isn't neither a letter nor a word... 

@lopuhin I have just pushed the new code. Let me know what you think

Gallaecio: Since `tokenizer` is not a private class (i.e. `_tokenizer`), and these changes change the behavior (for the better) instead of simply fixing the cast bug, I think we might need to deprecate `tokenizer`, copy it as `_tokenizer`, fix `_tokenizer` and use `_tokenizer` in our internal code.

I was worried that these changes would also break our code, as all non-ASCII characters are now considered "non-words", but it looks like in the parts of Dateparser where the tokenizer is used that is not an issue (or our test suite is incomplete around this; let’s hope not).

lopuhin: > I was worried that these changes would also break our code, 

I also looked for cases where we would distinguish between non-words and "other" and didn't find any.

PR #623, opened 2020-02-28
add parse_timedelta

noviluni:


dshefman1:
In my opinion, this would be a good addition to the dateparser library. It would get rid of a lot of the perceived and real complexity of having to convert absolute dates to relative dates. It would also more closely align the functionality of the dateparser library with the functionality of the datetime library.

PR #627, opened 2020-03-06
add jalali parser

noviluni:
As we implemented the `PARSERS` settings I thought that it could be a good idea to add an optional `jalali` parser.

In this way, it shouldn't be necessary to do this (as explained in the docs):
```
from dateparser.calendars.jalali import JalaliCalendar
JalaliCalendar(<jalali-string>).get_date()
```

And we could directly do this:
```
>>> dateparser.parse("۱۳۹۸/۰۶/۳۱ در ۱۴:۵۲", settings={'PARSERS': ['jalali']})  
datetime.datetime(2019, 9, 22, 14, 52)
```

(My idea is not to include it in the default parsers list, and to allow to enable it by changing the settings or through the settings param)

The current implementation I've done it's working for the example above, however, it's not working for this example:
```
>>> dateparser.parse(u'جمعه سی ام اسفند ۱۳۸۷', settings={'PARSERS': ['jalali']})
```

However, using the `JaliliCalendar` directly works for both examples:
```
>>> from dateparser.calendars.jalali import JalaliCalendar
>>> JalaliCalendar(u'جمعه سی ام اسفند ۱۳۸۷').get_date()
{'date_obj': datetime.datetime(2009, 3, 20, 0, 0), 'period': 'day'}

```

As far as I see this is due to the fact that before calling the parsers, in `get_date_data` we try to decide which language is used, and while the first example is correctly detected as Persian (`fa`), the second example is not detected as Persian and it returns an empty date object.

I'm not sure what to do. We could improve the Persian support, but maybe it doesn't make sense if we don't need to know that it's Persian to use the new `jalili` parser. 

What do you think? any feedback? @Gallaecio 

Thanks in advance

P.S: Of course, my idea after deciding how to proceed is to do the same with the Hijri calendar.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -465,11 +465,12 @@ Supported Calendars
 
 * Persian Jalali calendar. For more information, refer to `Persian Jalali Calendar <https://en.wikipedia.org/wiki/Iranian_calendars#Zoroastrian_calendar>`_.
 
-* Hijri/Islamic Calendar. For more information, refer to `Hijri Calendar <https://en.wikipedia.org/wiki/Islamic_calendar>`_.
-
     >>> from dateparser.calendars.jalali import JalaliCalendar
     >>> JalaliCalendar(u'جمعه سی ام اسفند ۱۳۸۷').get_date()
     {'date_obj': datetime.datetime(2009, 3, 20, 0, 0), 'period': 'day'}
+
+
+* Hijri/Islamic Calendar. For more information, refer to `Hijri Calendar <https://en.wikipedia.org/wiki/Islamic_calendar>`_.
 
     >>> from dateparser.calendars.hijri import HijriCalendar
     >>> HijriCalendar(u'17-01-1437 هـ 08:30 مساءً').get_date()
--- a/dateparser/date.py
+++ b/dateparser/date.py
@@ -9,6 +9,7 @@ import six
 import regex as re
 from dateutil.relativedelta import relativedelta
 
+from dateparser.calendars.jalali import JalaliCalendar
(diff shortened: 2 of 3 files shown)


Gallaecio:
I think it should be OK to skip language detection and translation on this new parser if that improves input data support.

If we later find an input string where we need language detection, we can work on it then. As long as we have those two input strings in the tests, everything should be fine.

noviluni:
Hi @Gallaecio, I don't see an easy way to bypass the lang detection, as it's expected to have a language to start parsing the date.

I added a commit with a workaround setting default locales, but maybe it's too custom for this. Let me know what you think about this approach.

README.rst
```diff
@@ -465,12 +465,13 @@ Supported Calendars
 
 * Persian Jalali calendar. For more information, refer to `Persian Jalali Calendar <https://en.wikipedia.org/wiki/Iranian_calendars#Zoroastrian_calendar>`_.
 
-* Hijri/Islamic Calendar. For more information, refer to `Hijri Calendar <https://en.wikipedia.org/wiki/Islamic_calendar>`_.
-
     >>> from dateparser.calendars.jalali import JalaliCalendar
     >>> JalaliCalendar(u'جمعه سی ام اسفند ۱۳۸۷').get_date()
     {'date_obj': datetime.datetime(2009, 3, 20, 0, 0), 'period': 'day'}
```
Gallaecio, line 470: We need to update this, right?

Gallaecio (commented):
Works for me.

PR #629, opened 2020-03-07
fix prefer dates from to exclude wrong words

noviluni:
At this moment if we use a wrong word in `PREFER_DATES_FROM` containing the word "future" (for example "afuture") the relative-time parser takes it as equivalent to "future". 

Now:
```
>>> dateparser.parse('3 days', settings={'PREFER_DATES_FROM': 'afuture'}) 
datetime.datetime(2020, 3, 10, 10, 40, 58, 68963)  # future (today is 7th)
```

After merging this PR: 
```
>>> dateparser.parse('3 days', settings={'PREFER_DATES_FROM': 'afuture'}) 
datetime.datetime(2020, 3, 4, 10, 40, 58, 68963) # past

```

As "afuture" is an invalid value, we shouldn't treat it as "future". 

I think we should add a warning when the value of `PREFER_DATES_FROM` is not `future` or `past`.

### DIFF
--- a/dateparser/freshness_date_parser.py
+++ b/dateparser/freshness_date_parser.py
@@ -128,8 +128,8 @@ class FreshnessDateDataParser(object):
         td = relativedelta(**kwargs)
         if (
             re.search(r'\bin\b', date_string) or
-            ('future' in prefer_dates_from and
-             not re.search(r'\bago\b', date_string))
+            re.search(r'\bfuture\b', prefer_dates_from) and
+            not re.search(r'\bago\b', date_string)
         ):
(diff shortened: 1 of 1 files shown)


Gallaecio:
I would rather raise a `ValueError` if a setting is assigned an invalid value.

noviluni:
Hi @Gallaecio ,

I created this PR: https://github.com/scrapinghub/dateparser/pull/630 to discuss how to raise an error if a setting value is not valid. In the meanwhile, I think this could be merged, as it fixes a current bug. 

What do you think?

PR #637, opened 2020-03-19
Update timezones.py

bazingarj:
The IST timezone was wrong - it was 7200 (2 hours) but in actual it is 19800 (5 hours 30minutes)


noviluni:
I think we should rethink how the timezones are applied, as we have more than one case of abbreviations with different meanings. However, I will close this PR because the solution can't be just overriding the default behavior. If you have any idea don't hesitate to comment on the related issue.

PR #639, opened 2020-03-19
Changing parse behaviour for 8 digit date

### LINKED ISSUE #618: Improve the default parsing of 20110101
While I don’t think https://github.com/scrapinghub/dateparser/issues/606 is necessarily a bug, I do think that it would be an improvement for Dateparser to parse '20110101' the same as '2011-01-01' by default.

arnavkapoor:
Fixes #618 
An attempt at #618 , the objective is to improve the parsing of dates of the form XXXXXXXX (Any 8 digit number).

Updated parsing results after this fix

- ` 20202001  datetime.datetime(2020, 1, 20, 0, 0)`
- ` 01201702  datetime.datetime(1702, 1, 20, 0, 0)`
- ` 01202020  datetime.datetime(2020, 1, 20, 0, 0)`
- ` 20202020  datetime.datetime(2002, 2, 2, 0, 2)`

For the last case `20202020` it's going back to the default behavior.

### DIFF
--- a/dateparser/parser.py
+++ b/dateparser/parser.py
@@ -25,11 +25,6 @@ def no_space_parser_eligibile(datestring):
 
     return False
 
-def is_8_digit_number(datestring):
-    if EIGHT_DIGIT.match(datestring) is None:
-        return False
-    return True
-
 def get_unresolved_attrs(parser_object):
     attrs = ['year', 'month', 'day']
     seen = []
@@ -153,7 +148,7 @@ class _no_spaces_parser(object):
     def _find_best_matching_date(cls, datestring):
         for fmt in cls._preferred_formats_ordered_8_digit:
             try:
-                dt = strptime(datestring, fmt), cls._get_period(fmt)
+                dt = strptime(datestring, fmt), cls._get_period(fmt)                
                 if len(str(dt[0].year)) == 4:
                     return dt
             except:
@@ -173,7 +168,7 @@ class _no_spaces_parser(object):
             order = resolve_date_order(settings.DATE_ORDER)
         else:
             order = cls._default_order
-            if is_8_digit_number(datestring):
+            if bool(EIGHT_DIGIT.match(datestring)):
                 dt = cls._find_best_matching_date(datestring)
                 if dt is not None:    
                     return dt               
--- a/dateparser/parser.py
+++ b/dateparser/parser.py
@@ -25,6 +25,7 @@ def no_space_parser_eligibile(datestring):
 
     return False
 
+
 def get_unresolved_attrs(parser_object):
     attrs = ['year', 'month', 'day']
     seen = []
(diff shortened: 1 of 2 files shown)


dateparser/parser.py
```diff
@@ -24,6 +25,10 @@ def no_space_parser_eligibile(datestring):
 
     return False
 
+def is_8_digit_number(datestring):
+    if EIGHT_DIGIT.match(datestring) is None:
+        return False
+    return True
```
Gallaecio: This could be simplified as `return bool(EIGHT_DIGIT.match(datestring))`, so it may not be worth having a separate function for this.

arnavkapoor: Thanks for the suggestion. Updated it.

dateparser/parser.py
```diff
@@ -142,6 +144,17 @@ def _get_period(cls, format_string):
         else:
             return 'year'
 
+    @classmethod
+    def _find_best_matching_date(cls, datestring):
+        for fmt in cls._preferred_formats_ordered_8_digit:
+            try:
+                dt = strptime(datestring, fmt), cls._get_period(fmt)
```
Gallaecio: There are extra spaces at the end here.

dateparser/parser.py
```diff
@@ -155,6 +168,10 @@ def parse(cls, datestring, settings):
             order = resolve_date_order(settings.DATE_ORDER)
         else:
             order = cls._default_order
+            if bool(EIGHT_DIGIT.match(datestring)):
```
Gallaecio: The `if` makes the `bool()` unnecessary.

```suggestion
            if EIGHT_DIGIT.match(datestring):
```

dateparser/parser.py
```diff
@@ -120,7 +122,7 @@ def __init__(self, *args, **kwargs):
         self._all = (self._dateformats +
                      [x+y for x in self._dateformats for y in self._timeformats] +
                      self._timeformats)
-
+
```
Gallaecio: Careful with adding extra spaces.

tests/test_parser.py
```diff
@@ -270,6 +270,36 @@ def test_default_order_used_if_date_order_not_supplied(self, date_string, expect
...
+            date_string=u"01202020",
+            expected_date=datetime(2020, 1, 20),
+            expected_period='day',
+        ),
+        param(
+            date_string=u"20202001",
+            expected_date=datetime(2020, 1, 20),
+            expected_period='day',
+        ),
+    ])
```
Gallaecio: See if you can add a test that results in `None`, to cover that scenario and complete [test coverage on the changes](https://codecov.io/gh/scrapinghub/dateparser/pull/639/diff#D1-156).

arnavkapoor: Sure added.

dateparser/parser.py
arnavkapoor: I have fixed all the whitespace issues in adherence flake8 rules in the latest commit.

PR #641, opened 2020-03-21
Fix issue #571  : Add 24hr format support

Avilad0:
Fix issue #571

adds 24hr format support
adds decimal format support i.e. single digit specifications in time

PR #634, opened 2020-03-12
Handle 24hr time format with Meridian Info

Gs-001:
Fixes #571 
Closes https://github.com/scrapinghub/dateparser/pull/654

- Handle 24hr time format with meridian
- Added and Updated tests

The idea was to ignore meridian info. when the time is following 24hr format. 
I added a new time format to handle this case. The output for this case is now same as when no meridian info. is provided for 24hr time format.

This case was specified among the date strings that would throw an error, but as now it is being handled I have updated the tests accordingly.

PR #654, opened 2020-03-31
Added format and test cases for the 24 hrs - decimal time parsing

### LINKED ISSUE #571: dateparser.parse returns None for any datetime using 24hr format and time period when hour is > 12
Whils't parsing some data I got an error when processing `Sep 28, 2016, 20:13 pm`, I believe this is due to a conflict between 24 hour time format and a time period being supplied.
```
>>> print(dateparser.parse('Sep 28, 2016, 12:13 PM'))
2016-09-28 12:13:00
>>> print(dateparser.parse('Sep 28, 2016, 11:13 PM'))
2016-09-28 23:13:00
>>> print(dateparser.parse('Sep 28, 2016, 13:13 PM'))
None
>>> print(dateparser.parse('Sep 28, 2016, 13:13 AM'))
None
>>> 
```
I know this isn't technically valid since the period is not needed with 24hr format. Instead of returning `None` I believe that it would be useful to instead just ignore the given period and fallback on the given 24hr format.

i.e.:
`Sep 28, 2016, 13:13 AM` -> `Sep 28, 2016, 13:13`
`Sep 28, 2016, 13:13 PM` -> `Sep 28, 2016, 13:13`

Avilad0:
Fixes #571, closes #641


Gallaecio:
Could you share references to real life scenarios where these formats are used? Also, I noticed you added 4 formats but only 2 tests.

Avilad0:
@Gallaecio  sorry for the delay, i was occupied in some other work. I have added 2 more test cases which do not have am/pm reference which are the remaining 2 test cases. Please refer it.

Gallaecio:
Could you share URLs to a couple of online resources where these formats are used?

Avilad0:
https://strftime.org/   -  this link defines it purpose 

it basically adds the functionality to accept time without mentioning in specific format HH:MM:SS (double digits).    eg. 1:1:1 instead of writing 01:01:01

Gallaecio:
The line switches at `dateparser/parser.py` seem wrong, an accident?

Avilad0:
Yes, sorry it was an an accident.

dateparser/parser.py
```diff
@@ -81,6 +81,10 @@ class _time_parser(object):
         '%I %p',
         '%H:%M:%S.%f',
         '%I:%M:%S.%f %p',
+        '%-H:%-M',
+        '%-H:%-M:%-S',
+        '%-H:%-M %p',
+        '%-H:%-M:%-S %p'
```
Gallaecio, line 87: Woudn’t the fix for https://github.com/scrapinghub/dateparser/issues/571 be to simply add `'%H:%M %p'`? I’m worried that we are opening the door here for wrong interpretations, to parse as dates stuff that is not a date. None of the input strings from the test look to me like real-life date strings. I could understand the hours part having a single digit, and if that requires `%-H` then `%-H` should be OK, but `%-M` and `%-S` seem wrong to me, and I would rather keep them out of the default list.

Avilad0: Yeah, That may be the case then we can simply add the single digit format for hours only as in '%-H:%M' etc.

noviluni: Hi @Gallaecio 
I needed some time to realize it, but it doesn't make any sense to add `'%-H:%M %p'` as is equivalent to `'%I:%M %p'` for values where hour < 10 and equivalent to `'%H:%M %p'` for values where hour >= 10

Anyway, I have just merged this: https://github.com/scrapinghub/dateparser/pull/634

PR #664, opened 2020-05-02
add python 3.8 support and clean old tests references

noviluni:
Add Python 3.8 support and clean some references to the old test suite.


noviluni:
The build is passing, but for some unkown reason it doesn't appear: https://travis-ci.org/github/scrapinghub/dateparser/builds/682239290

PR #674, opened 2020-05-13
Fix languge unique chars

Ronserruya:
This PR fixed 1 bug and **introduces(unveils?) one bug**

##Bug # 1:
```python
In [7]: search_dates("july 2018 - december 2020", add_detected_language=True)
Out[7]:
[('july 2018', datetime.datetime(2018, 7, 13, 0, 0), 'en'),
 ('december 2020', datetime.datetime(2020, 12, 13, 0, 0), 'en')]

In [8]: search_dates("july 2018 - december 2020", add_detected_language=True, languages=['en','es'])
Out[8]:
[('july 2018', datetime.datetime(2018, 7, 13, 0, 0), 'en'),
 ('december 2020', datetime.datetime(2020, 12, 13, 0, 0), 'en')]

In [9]: search_dates("july 2018 - december 2020", add_detected_language=True, languages=['en','fr'])
Out[9]:
[('2018', datetime.datetime(2018, 5, 13, 0, 0), 'fr'),
 ('2020', datetime.datetime(2020, 5, 13, 0, 0), 'fr')]
```

As you can see, including some languages, like French **(in addition to English)** causes issues with finding dates. This is caused by the `character_check` / `get_unique_characters` methods, some languages had punctuations in their "unique characters", for example "-" for French or '.' for Thai.

Due to this reason the parser things that the text `July 2018 - December 2020` is in French and ignores the months.

This is fixed by making sure that punctuations don't get into the unique chars set.

##Bug # 2

```
In [11]: search_dates("01.09 – 03.09.2017", add_detected_language=True)
Out[11]:
[('01.09', datetime.datetime(2020, 1, 9, 0, 0), 'th'),
 ('03.09.2017', datetime.datetime(2017, 3, 9, 0, 0), 'th')]
```

After I fixed this issue I found another issue that was "hidden" before, the following test: https://github.com/scrapinghub/dateparser/blob/master/tests/test_search.py#L670
Only ever worked because of the previous bug, `01.09 – 03.09.2017` contains dots, so the parser thought it was in Thai, and used the specific Thai word splitter https://github.com/scrapinghub/dateparser/blob/master/dateparser/languages/locale.py#L256

Now that the parser correctly recognizes that the text is not in Thai (so English), it uses the English splitter that splits on dots, which screws up the date.

I tried fixing it as well but I wasn't successful, hopefully you can get it working with your experience with the codebase.

Obviously, due to the second bug, 1 test is failing

### DIFF
--- a/dateparser/search/text_detection.py
+++ b/dateparser/search/text_detection.py
@@ -1,11 +1,14 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
+from string import digits, punctuation
 from dateparser.search.detection import BaseLanguageDetector
 from dateparser.conf import apply_settings
 from dateparser.utils import normalize_unicode
 
 
 class FullTextLanguageDetector(BaseLanguageDetector):
+    SYMBOL_SET = set(list(digits + punctuation + ' ' + '–'))
+
     def __init__(self, languages):
         super(BaseLanguageDetector, self).__init__()
         self.languages = languages[:]
@@ -21,6 +24,7 @@ class FullTextLanguageDetector(BaseLanguageDetector):
 
         for char_set in self.language_chars:
             unique_chars = char_set
+            unique_chars -= self.SYMBOL_SET
             for other_char_set in self.language_chars:
                 if other_char_set != char_set:
                     unique_chars = unique_chars - other_char_set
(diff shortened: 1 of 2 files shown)


Gallaecio:
I wonder how hard it would be to change the French and Thai language detection to only trigger if there are no spaces surrounding their respective characters.

Ronserruya:
> I wonder how hard it would be to change the French and Thai language detection to only trigger if there are no spaces surrounding their respective characters.

What would that achieve?

Gallaecio:
I imagine the goal of including those characters in language detection was to detect dates in the formats common for those locales. That’s what we loose with this patch. Maybe we can figure out a way to keep the old indented behavior while still fixing the issue.

PR #675, opened 2020-05-14
clean supplementary language data

noviluni:
I cleaned the duplicated entries in the `yaml` files. I also deleted the `dateorder` property, as it was wrong (it should be `date_order`) and it was already in the `.py` files.


noviluni:
I think the import of duplicated entries will be fixed in: https://github.com/scrapinghub/dateparser/pull/487

However, it could be interesting to add a warning if there are duplicated entries in the `yaml` file and data coming from CLDR (`json`), as this will help to keep the `yaml` files cleaner.

I will submit a PR with this.

PR #682, opened 2020-05-16
Interpret BST as British Summer Time

Baviaan:
I have checked https://www.timeanddate.com/time/zones/ and apparently BST can also mean Bougainville Standard Time, which I guess is why the BST offset is set to +11.

I understand it is ambiguous, but I believe the majority of users will mean UTC+1 with BST.


noviluni:
Hi @Baviaan, thank you for your feedback here!

I totally agree that the current timezone implementation should be improved and I just opened an issue to discuss this: https://github.com/scrapinghub/dateparser/issues/683

**BST means:**
* Bangladesh Standard Time: UTC+06
* British Summer Time (British Standard Time from Feb 1968 to Oct 1971): UTC+01
* Bougainville Standard Time: UTC+11

This is happening with other timezones, and we can't just change the default behavior, as it wouldn't be backward compatible and on the other side, the problem would remain the same.

I will close this PR as we can't merge it but please, if you could read the mentioned issue (https://github.com/scrapinghub/dateparser/issues/683) and give feedback with your ideas it would be really appreciated. :raised_hands:

Issue #685, opened 2020-05-17
Bad handling of big numbers (OverflowError)


noviluni:
There are a lot of situations where using a big number we get an `OverflowError`:

```python3
>>> dateparser.parse('123456789 hour')
OverflowError: date value out of range

>>> dateparser.parse('123456789123 hour')
OverflowError: Python int too large to convert to C int

>>> dateparser.parse('1234567 days') 
OverflowError: date value out of range

>>> dateparser.parse('1234567891 days') 
OverflowError: days=-1234567891; must have magnitude <= 999999999

>>> dateparser.parse('12345678912 days') 
OverflowError: Python int too large to convert to C int

>>> dateparser.parse('123455678976543 month')
OverflowError: signed integer is less than minimum
```
It happens the same for `week`, `year`, etc.

This case, handled by python3.6 calendar seems better:
```python
>>> dateparser.parse('123456789 year') 
ValueError: year -123454769 is out of range
```

However, I think we should just ignore or return `None` if any of these cases is found.

PR #686, opened 2020-05-17
fix OverflowError in freshness dateparser

noviluni:
Fixes https://github.com/scrapinghub/dateparser/issues/685

~I pushed this branch with the tests failing to ask for your feedback.~

Until now, when entering some wrong values in the `relative-time` parser (aka `freshness_parser`), like `"4000 years ago"`, the `dateparser.parse()` raises a `ValueError` (this is validated in the tests). 

However, this is working only partially, as bigger values return an `OverflowError` (more context here: https://github.com/scrapinghub/dateparser/issues/685).

When fixing this I considered that we should always return `None` instead of a `ValueError` because when parsing long texts or in the middle of a script we don't want the `dateparser.parse()` method to raise an error ~, however, it's possible that some people use this behavior when entering a string like "X years ago", expecting an error when the year is not valid.~

~What do you think? It's ok to change the current logic? Should we reraise the `OverflowError` as a `ValueError`? In the first case, as it isn't backward compatible, should we release a major version (or wait until it is ready to be released)?~

@Gallaecio 

Update:
After checking this issue: https://github.com/scrapinghub/dateparser/issues/687 it seems obvious that we can't raise `ValueError`s because it breaks `search_dates`, etc. I know this won't be backward compatible, but I think it's the best solution. @Gallaecio , I would like to hear your opinion on this. Thanks in advance.

Fixes: https://github.com/scrapinghub/dateparser/issues/687

### DIFF
--- a/dateparser/date.py
+++ b/dateparser/date.py
@@ -208,7 +208,10 @@ class _DateLocaleParser(object):
         }
 
     def _try_freshness_parser(self):
-        return freshness_date_parser.get_date_data(self._get_translated_date(), self._settings)
+        try:
+            return freshness_date_parser.get_date_data(self._get_translated_date(), self._settings)
+        except (OverflowError, ValueError):
+            return None
 
     def _try_parser(self):
         _order = self._settings.DATE_ORDER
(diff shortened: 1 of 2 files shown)


Gallaecio:
The [pull request](https://github.com/scrapinghub/dateparser/pull/26) that [introduced that behavior](https://github.com/scrapinghub/dateparser/commit/b96cbde9e9f95323e7b3073d55e79d9c23df00d0#diff-ff270e36b5b263db2509412ab28935cfR214) seems unrelated.

I wonder if we should add some logging, even if it is debug level, about this. But that’s a different topic.

PR #696, opened 2020-05-28
Fix name params in the DateDataParser docstring.

wRAR:
Similar to #557 (no idea if there are other copy-paste errors, I just searched for `allow_redetect_language`).


noviluni:
Good catch! Thank you @wRAR!

Issue #688, opened 2020-05-18
Test raise an unnecessary FutureWarning 


eamanu:
Hello, 

This issues was reported originally here [0]. If I'm not wrong
the warning is emit here [1]. 

The test `test_should_support_a_string_as_date_formats`  wait for a
FutureWarning intenttionally?

IMO this test must have a `self.assertWarns` or if there aren't intentions
to test the Warning emited could be a good option change 
`date_formats='%d-%m-%y'` for `date_formats=['%d-%m-%y']`

[0] https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=954907
[1] https://github.com/scrapinghub/dateparser/blob/master/tests/test_date.py#L302

PR #699, opened 2020-05-29
Add a assertWarns to avoid unnecessary messages.

eamanu:
This PR propose create a new method that assert Warnings
to avoid print on screen the future warns.

This patch just work for Python3, becuase ssertWarns` isn't
available for Python2.

Closes: #688


noviluni:
thanks  @eamanu! :smile:

PR #703, opened 2020-06-04
Add `REQUIRE_PARTS` setting

noviluni:
I decided to continue this PR: https://github.com/scrapinghub/dateparser/pull/532 as it seems stalled.

You can fin what I did in my commits:
* Update with current code (merging master)
* Change from `None` and comma-separated strings to a list.
* Remove settings from README.md (I don't think we should move the same content to the README.md, instead we should improve it reducing the content and pointing to the documentation.
* Improve logic for STRICT_PARSING and REQUIRE_PARTS.

closes: https://github.com/scrapinghub/dateparser/pull/532


dateparser/parser.py
```diff
@@ -346,16 +346,20 @@ def _get_date_obj(self, token, directive):
...
-                errors.append('Month')
-            if not self.year:
-                errors.append('Year')
+        missing = []
+        if not self.day:
+            missing.append('day')
+        if not self.month:
+            missing.append('month')
+        if not self.year:
+            missing.append('year')
```
Gallaecio, line 355: :lipstick: 

```suggestion
        missing = [field for field in ('day, 'month', 'year')
                   if not getattr(self, field)]
```

dateparser/parser.py
```diff
@@ -346,16 +346,20 @@ def _get_date_obj(self, token, directive):
...
+
+        if self.settings.STRICT_PARSING and missing:
+            raise ValueError('%s not found in the date string' % ', '.join(missing))
```
Gallaecio, line 358: :lipstick: “day, month not found in the date string” feels weird to me. Same applies below.

```suggestion
            raise ValueError('Fields missing from the date string: %s' % ', '.join(missing))
```

docs/usage.rst
```diff
@@ -182,9 +182,25 @@ The following parsers exist:
...
 
->>> from dateparser.settings import default_parsers
->>> parsers = [parser for parser in default_parsers if parser != 'relative-time']
->>> parse('today', settings={'PARSERS': parsers})
+    >>> from dateparser.settings import default_parsers
+    >>> parsers = [parser for parser in default_parsers if parser != 'relative-time']
+    >>> parse('today', settings={'PARSERS': parsers})
+
+
+``REQUIRE_PARTS`` This option ensures results are dates that have some specified part. It defaults to ``[]`` and can include ``day``, ``month`` and/or ``year``.
```
Gallaecio, line 190:
```suggestion
``REQUIRE_PARTS`` This option ensures results are dates that have all specified parts. It defaults to ``[]`` and can include ``day``, ``month`` and/or ``year``.
```

dateparser/parser.py
noviluni: Beautiful! :heart:

dateparser/parser.py
noviluni: This message is not shown, in fact, before my changes it wasn't separated by commas, creating a weird/wrong message like "DayMonthYear not found in the date string". 

However, I decided to follow your advice and I changed the code to use a better message. Thanks @Gallaecio!

dateparser/parser.py
noviluni: There is a single quote (`'`) missing after `day`, but don't worry, I added it :smile:

PR #709, opened 2020-06-17
Use chain.from_iterable in dictionary.py

cool-RR:
This is a faster and more idiomatic way of using `itertools.chain`. Instead of computing all the items in the iterable and storing them in memory, they are computed one-by-one and never stored as a huge list. This can save on both runtime and memory space.


noviluni:
I think we could also apply it here: 
* https://github.com/scrapinghub/dateparser/blob/4e4b6d9fd9c64df62b4471b5618ff6bee27d211a/dateparser/languages/locale.py#L349

cool-RR:
Correct, and another place in that module. I created a separate PR #710 for that (just because I'm too lazy to clone :) )

Issue #385, opened 2018-02-09
ummalqurra is supported on Python 3


dotlambda:
Since umalqurra is supported on Python 3 (cf. https://github.com/tytkal/python-hijiri-ummalqura/pull/5), the tests in https://github.com/scrapinghub/dateparser/blob/1764bcfb4932b8928541d8ff58c91b391c15e1e5/tests/test_hijri.py#L9 should be enabled.

wRAR:
Unfortunately, the latest release on pypi is still 0.2 which is older than that change.

PR #718, opened 2020-06-23
Change hijri calendar dependency to support Python 3

noviluni:
The current package we are using to parse dates from the **Hijri calendar** (https://pypi.org/project/ummalqura/) is not maintained and it doesn't support Python 3.

I think that we can use this other package, as it has been maintained and it works pretty well, containing all the features we need: https://github.com/dralshehri/hijri-converter

You can see in the diff below how simple it is to adapt this new package to the current dateparser status (tests are passing).

This library only supports Python 3.6+, but that's not a problem, as we will remove Python 2.7 support in the next release and the end of life of Python 3.5 is 2020-09-13 (the next release of `dateparser` is planned for September or October, we don't need to remove support for python 3.5 but just add a note about the hijri calendar compatibility). 

fixes: https://github.com/scrapinghub/dateparser/issues/385


tests/test_hijri.py
```diff
@@ -1,26 +1,27 @@
...
 from datetime import datetime
 import six
-import unittest
 
 from parameterized import parameterized, param
 
-# umalqurra does not support Python3 yet
-# see https://github.com/tytkal/python-hijiri-ummalqura/pull/5
-# let's skip these tests under Python3 for now
+is_python_supported = six.PY3 and sys.version_info.minor > 5
```
Gallaecio, line 11: :lipstick: 

```suggestion
is_python_supported = sys.version_info >= (3, 6)
```

Slightly more readable. And we should try to avoid using `six` going forward, removing its usage should be one of the cleanup goals now that Python 2 is no longer supported.

tests/test_hijri.py
```diff
@@ -1,26 +1,27 @@
...
+
 try:
     from dateparser.calendars.hijri import HijriCalendar
 except ImportError:
-    if not six.PY2:
+    if not is_python_supported:
         pass
     else:
         raise
```
Gallaecio, line 19: :lipstick: 

```suggestion
    if is_python_supported:
        raise
```

tests/test_hijri.py
noviluni: Totally agree!

README.rst
```diff
@@ -465,17 +465,18 @@ Supported Calendars
...
 
+
+* Hijri/Islamic Calendar. For more information, refer to `Hijri Calendar <https://en.wikipedia.org/wiki/Islamic_calendar>`_.
+
     >>> from dateparser.calendars.hijri import HijriCalendar
     >>> HijriCalendar(u'17-01-1437 هـ 08:30 مساءً').get_date()
     {'date_obj': datetime.datetime(2015, 10, 30, 20, 30), 'period': 'day'}
 
-.. note:: `HijriCalendar` has some limitations with Python 3.
+.. note:: `HijriCalendar` only works with Python > 3.6.
```
Gallaecio, line 479:
```suggestion
.. note:: `HijriCalendar` only works with Python >= 3.6.
```

noviluni:
Oh, it wouldn't be a breaking change.

I was waiting 2 weeks to ensure that the last version is stable enough before proceeding with merging those PRs that would break the Python 2.7 code and marked them with the "breaking change" label to be easily recognizable. I just wrote about this here: https://github.com/scrapinghub/dateparser/issues/727#issuecomment-651078076

I agree with you that in the next release notes this shouldn't be labeled as a "breaking change".

noviluni:
code rebased against the master branch to fix conflicts

PR #720, opened 2020-06-24
fix translate keep_formatting

noviluni:
When parsing a string it is "translated". When doing it, the commas are missed, and because of that some date formats in the `base-formats` parser doesn't work.

**Before this change:**

```python
>>> dateparser.parse('December 04, 1999, 11:04:59 PM', settings={'PARSERS': ['base-formats']})

```

It should work because `'%B %d, %Y, %I:%M:%S %p'` is the first date format when trying with `base-formats, but it doesn't.

**After this change:**

```python
>>> dateparser.parse('December 04, 1999, 11:04:59 PM', settings={'PARSERS': ['base-formats']})
datetime.datetime(1999, 12, 4, 23, 4, 59)
```

### DIFF
--- a/dateparser/languages/locale.py
+++ b/dateparser/languages/locale.py
@@ -141,7 +141,8 @@ class Locale(object):
                     date_string_tokens[i] = pattern.sub(replacement, word)
             else:
                 if word in dictionary:
-                    date_string_tokens[i] = dictionary[word] or ''
+                    fallback = word if keep_formatting and not word.isalpha() else ''
+                    date_string_tokens[i] = dictionary[word] or fallback
         if "in" in date_string_tokens:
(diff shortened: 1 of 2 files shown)


noviluni:
@Gallaecio maybe we don't need this :thinking:

please, take a look to this: https://github.com/scrapinghub/dateparser/pull/721

noviluni:
Even if we decided to remove the `base-formats` parser (https://github.com/scrapinghub/dateparser/pull/721), I will keep this PR open to check if this is affecting the `custom-formats` parser. From my first attempts, it seems that the `custom-formats` doesn't take into account commas, etc. so we could consider that it's working wrongly. I will investigate this further.

PR #630, opened 2020-03-07
wip:settings validation

noviluni:
I think we should add warnings if the values of the settings are not the expected. 

I have done this PR to illustrate the idea of a common framework to do it.

Examples:
```
>>> dateparser.parse('3 days', settings={'PREFER_DAY_OF_MONTH': 'aaaa'}) 
"aaaa" is not a valid value for PREFER_DAY_OF_MONTH, it should be: 'current', 'first' or 'last'. Using default behaviour: 'current'
datetime.datetime(2020, 3, 4, 16, 36, 3, 322839)
```

```
>>> dateparser.parse('3 days', settings={'PREFER_DATES_FROM': 'afuture'})
"afuture" is not a valid value for PREFER_DATES_FROM, it should be: 'past', 'future' or 'current_period'. Using default behaviour: 'current_period'
datetime.datetime(2020, 3, 10, 16, 37, 25, 516574)
```

```
>>> dateparser.parse('03/04/05', settings={'DATE_ORDER': 'DEE'}) 
ValueError: "DEE" is not a valid value for DATE_ORDER, it should be: 'MDY', 'MYD', 'DMY', 'DYM', 'YDM' or 'YMD'. Using default behaviour: 'MDY'
```

Let me know what you think or any idea you have.

PR #722, opened 2020-06-25
document settings

noviluni:
I've been checking the current settings as a previous step to work on this: https://github.com/scrapinghub/dateparser/pull/630

~I've been reordering the settings, reordering the settings in the docs, and checking the currently undocumented settings.~

~The undocumented settings are:~
~* SKIP_TOKENS_PARSER~
~* NORMALIZE~
~* FUZZY~

~I documented the `NORMALIZE` setting, however, after some checks, it seems that those settings are not documented because:~
~* It doesn't have any sense or use cases to change them (they were created because of legacy reasons).~
~* Nobody asked for it in issues, etc.~

~From this, I think that they are not useful at all, and it doesn't have any sense to keep them. For example, the default list in `SKIP_TOKENS_PARSER` can be set directly in the code, or `NORMALIZE=True` can be applied always as it is really specific (it only has sense when the checked source (CLDR json and yamls files) is not normalized, but if it's normalized it doesn't work).~

~My proposal is to remove them in the next major version but I need feedback on this. @Gallaecio~

I documented all the undocumented settings and made some improvements to the documentation structure.


noviluni:
Proposal for deleting `SKIP_TOKENS_PARSER: https://github.com/scrapinghub/dateparser/pull/728

Gallaecio:
I think removing them makes sense.

noviluni:
I will wait until merging this: https://github.com/scrapinghub/dateparser/pull/779 
to avoid conflicts

docs/conf.rst
```diff
@@ -1,5 +0,0 @@
-Configurations
```
noviluni, line 1: Removed as it was not used

docs/settings.rst
```diff
@@ -0,0 +1,189 @@
+.. _settings:
```
noviluni, line 1: Added specific section for the settings (extracted from `usage.rst`). Multiple errors and improvements have been performed.

docs/usage.rst
```diff
@@ -8,21 +11,8 @@ from the same source.
...
-assumes that all of the dates fed to it are in the same language.
-
-.. autoclass:: dateparser.date.DateDataParser
-   :members: get_date_data
-
-.. warning:: It fails to parse *English* dates in the example below, because *Spanish* was detected and stored with the ``ddp`` instance:
-
-    >>> ddp.get_date_data('11 August 2012')
-    {'date_obj': None, 'period': 'day'}
-
```
noviluni, line 25: This is not true now, so I removed it.

dateparser/conf.py
```diff
@@ -11,17 +11,20 @@ class Settings:
...
+    * `RETURN_AS_TIMEZONE_AWARE`
+    * `PREFER_DAY_OF_MONTH`
+    * `PREFER_DATES_FROM`
+    * `RELATIVE_BASE`
+    * `STRICT_PARSING`
+    * `REQUIRE_PARTS`
+    * `SKIP_TOKENS`
+    * `NORMALIZE`
+    * `RETURN_TIME_AS_PERIOD`
+    * `PARSERS`
```
noviluni, line 27: I added all the missing settings and removed the descriptions because it makes this harder to maintain. For up to date descriptions we have the settings section in docs.

dateparser/date.py
```diff
@@ -268,7 +268,7 @@ class DateDataParser:
 
     :param locales:
         A list of locale codes, e.g. ['fr-PF', 'qu-EC', 'af-NA'].
-        The parser uses locales to translate date string.
+        The parser uses only these locales to translate date string.
```
noviluni, line 271: Clarified because of the comment here: https://github.com/scrapinghub/dateparser/pull/789#discussion_r492621092

Gallaecio (approved):
Nice!

PR #715, opened 2020-06-19
apply STRICT_PARSING in no-spaces-time parser

noviluni:
fixes: https://github.com/scrapinghub/dateparser/issues/713

~It's just a WIP. I need to add some tests and probably move the functions to another place (like `utils`).~

Issue #700, opened 2020-05-29
Question: is there a way to know when a date is incomplete.


marcelooyaneder:
Is there a way too know when a date is incomplete. e.g "December 2015" 

```python
from dateparser import parse
parse(u'December 2015')  # default behavior
datetime.datetime(2015, 12, 16, 0, 0)
#is a way to do something like this
parse(u'December 2015', settings={'PREFER_DAY_OF_MONTH': 'incomplete'}) 
datetime.datetime(2015, 12, incomplete, 0, 0)
#the same with year or day
```
I need to parse dates wich contains incomplete dates, in a nutshell i need the following to be done

```python
from dateparser import parse
parse(u'December 2015')  # default behavior
datetime.datetime(2015, 12, 16, 0, 0)

**Comment by Gallaecio** [comment]
I think this is a duplicate of https://github.com/scrapinghub/dateparser/issues/553

**Comment by noviluni** [comment]
Just added a PR trying to find a solution for this: https://github.com/scrapinghub/dateparser/pull/729

Please, if you have any feedback let me know.

**Comment by marcelooyaneder** [comment]
I've actually develop a little solution for that... on my testing is consistent, the code is the following.

```
imported_list=[]
for elements in df.verbatimEventDate:
    try:
        if math.isnan(elements):
            imported_list.append([None,None,None,None,None,"Empty"])
        else:
            date_element = dateparser.parse(str(elements), settings={'PREFER_DATES_FROM': 'past', 'STRICT_PARSING': True},languages=['pt', 'es'])
            imported_list.append([elements,date_element.strftime('%F'),date_element.strftime("%Y"),date_element.strftime("X%m").replace('X0','X').replace('X',''),date_element.strftime("X%d").replace('X0','X').replace('X',''),"High"])
    except TypeError: 
        date_element = dateparser.parse(elements, settings={'PREFER_DATES_FROM': 'past', 'STRICT_PARSING': True},languages=['pt', 'es'])
        if date_element is None: 
            date_element = dateparser.parse(elements, settings={'PREFER_DATES_FROM': 'past','DATE_ORDER': 'MYD'},languages=['pt', 'es'])
            imported_list.append([elements,date_element.strftime('%Y-%m'),date_element.strftime("%Y"),date_element.strftime("X%m").replace('X0','X').replace('X',''),None,"Medium"]) #just windows
        else: 
[...]

Issue #553, opened 2019-08-30
Should have a 'zero' option for incomplete dates


ilienert:
Instead of forcing some default day of the month for incomplete dates, it would be very helpful if you could zero out the day so you'd know what part of the date string was missing.

Example:
```
parse(u'December 2015', settings={'PREFER_DAY_OF_MONTH': 'zero'})
>>> datetime.datetime(2015, 12, 0, 0, 0)
```
The same behaviour should be available for years, months, etc.

waltertross:
Unfortunately that would be an invalid datetime (if you try to create it, you get "ValueError: day is out of range for month")
There should be another way of signalling whether date, time or both were set in each datetime. I think that with an option in settings one should be able to request an extra member of the output tuples carrying this information.

noviluni:
Just added a PR trying to find a solution for this: https://github.com/scrapinghub/dateparser/pull/729

Please, if you have any feedback let me know.

noviluni:
We can't provide incomplete dates, but we can give more context by using `DateDataParser()` and `get_date_data()` as explained in this comment: https://github.com/scrapinghub/dateparser/issues/700#issuecomment-734326956


So I will close this issue in favor of the other as it's pointing to the correct solution we will work on in the future. 

Thanks for the feedback :)

Issue #424, opened 2018-06-18
How to parse when month or day is missing?


pengyu:
The following only says the year is 2017. But `dateparser` will fill in the month and day as the today. I don't think that this is the most appropriate behavior.

```
>>> from dateparser import parse
>>> print(parse('2017'))
2017-06-18 00:00:00
```

Is there a way to just print the most specific information like the following?

- '2017' -> '2017'
- '2017 Jan' -> '2017-01'

kamilnematli:
It is actually not today's date but the reference date. As default the reference date is the current date. One workaround can be to set the reference date to 1st of January. By this, if the day is missing it will set it to 1st day of month and if day and month missing it will set it to 1st of Jan. But please be careful because there are some other date phrases that are parsed considering reference date, like yesterday.

noviluni:
Hi @pengyu, the `parse` method returns a `datetime` object (not a str) and you can format the result by using `strftime`.

In your examples:
* `print(parse('2017').strftime('%Y'))`
* `print(parse('2017 Jan').strftime('%Y-%m'))`

Check this to know how to use it: https://strftime.org/

As it doesn't exist a `datetime` object without the `day` or without the `month`, it's not technically possible to return less information, so gaps are automatically filled with the reference date (as @kamilnematli explained).

If you want to automatically (programmatically) format the dates according to the parsed content now it's not possible and I don't think it will be possible in the future, as the result of parse will be always a `datetime`. However, it could be possible to add another function to get information about what has been parsed (like the info we use when setting `STRICT_PARSING=True`), but this has only sense in certain circumstances, as when parsing "2017 Jan", but it hasn't sense when parsing strings like "yesterday".

noviluni:
Just added a PR trying to find a solution for this: https://github.com/scrapinghub/dateparser/pull/729

Please, if you have any feedback let me know.

PR #729, opened 2020-06-26
WIP: implement missing info

noviluni:
I did a proof of concept to add the missing parts ("year", "month" or "day") to the result of `DateDataParser.get_date_data()`.

This could be a fix for https://github.com/scrapinghub/dateparser/issues/424, https://github.com/scrapinghub/dateparser/issues/553 and https://github.com/scrapinghub/dateparser/issues/700.

Of course **tests are not passing**, it's just a PoC to discuss it.

-----

### Precedents: 
Sometimes there is a part of a date that it's missing: "22 may", "1994", etc. and there are people interested in knowing the missing part/s to be able to print / handle the result. Examples coming from the issues:

When doing:

```python
>>> from dateparser import parse
>>> print(parse('2017'))
2017-06-18 00:00:00
```
to be able to print just:
```bash
2017
```
and when doing 

```python
>>> print(parse('2017 Jan'))
2017-01-18 00:00:00
```

to be able to print:

```bash
2017-01
```

### Feasibility
As `dateparser.parse()` returns a `datetime` object, and it's not possible to return those parts with zero (example from the mentioned issues: `datetime.datetime(2015, 12, 0, 0, 0)`), we can't just do it.

However, as we can use directly the `DateDataParser.get_date_data()` method and it returns more information (`locale`,  `period`, etc.), we could add another field with the missing parts (if applicable).

I added an `extra` field containing the `missing` field. We could probably simplify this, adding the `missing` field directly.

### Code snippets with the implementation from this PR:

This:

```python
from dateparser.date import DateDataParser
p = DateDataParser()

print(p.get_date_data("22 may"))
```

will print:

```bash
{'date_obj': datetime.datetime(2020, 5, 22, 0, 0), 'period': 'day', 'extra': {'missing': ['year']}, 'locale': 'en'}
```

And this:

```python
print(p.get_date_data("1999"))
```

will print:
```bash
{'date_obj': datetime.datetime(1999, 6, 26, 0, 0), 'period': 'year', 'extra': {'missing': ['day', 'month']}, 'locale': 'en'}
```

With this, we could just do something like:
```python
from dateparser.date import DateDataParser
p = DateDataParser()

def print_date_obj(result):
    missing = result.get('extra', {}).get('missing', [])
    if not missing:
        date_format = '%Y-%m-%d'
    elif 'day' and 'month' in missing:
        date_format = '%Y'
    elif 'day' in missing:
        date_format = '%Y-%m'
    elif 'year' in missing:
        date_format = '%m-%d'

   ...

    print(result.get('date_obj').strftime(date_format))


result = p.get_date_data("22 may")

if result['date_obj']:
    print_date_obj(result)

```

### Feedback

This PR can't be merged as-is, we should work on the tests and add more scenarios. Apart from that, it should be merged after merging this: https://github.com/scrapinghub/dateparser/pull/715  (and we should fix the conflicts) 

## Let me know what you think about this approach and if you see any blocker.

### DIFF
--- a/dateparser/date.py
+++ b/dateparser/date.py
@@ -219,13 +219,17 @@ class _DateLocaleParser(object):
             if self._settings.PREFER_LOCALE_DATE_ORDER:
                 if 'DATE_ORDER' not in self._settings._mod_settings:
                     self._settings.DATE_ORDER = self.locale.info.get('date_order', _order)
-            date_obj, period = date_parser.parse(
+            date_obj, period, extra = date_parser.parse(
                 self._get_translated_date(), settings=self._settings)
             self._settings.DATE_ORDER = _order
-            return {
+            result = {
                 'date_obj': date_obj,
                 'period': period,
             }
+            if extra:
+                result['extra'] = extra
+            return result
+
         except ValueError:
             self._settings.DATE_ORDER = _order
             return None
@@ -271,7 +275,7 @@ class _DateLocaleParser(object):
     def _is_valid_date_obj(self, date_obj):
         if not isinstance(date_obj, dict):
             return False
-        if len(date_obj) != 2:
+        if len(date_obj) != 2 and len(date_obj) != 3:
             return False
         if 'date_obj' not in date_obj or 'period' not in date_obj:
             return False
--- a/dateparser/date_parser.py
+++ b/dateparser/date_parser.py
@@ -23,7 +23,7 @@ class DateParser(object):
         date_string = strip_braces(date_string)
         date_string, ptz = pop_tz_offset_from_string(date_string)
 
-        date_obj, period = parse(date_string, settings=settings)
+        date_obj, period, extra = parse(date_string, settings=settings)
 
         _settings_tz = settings.TIMEZONE.lower()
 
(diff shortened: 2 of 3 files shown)


Gallaecio:
I think having a function that returns all this data is necessary for some use cases.

My first instinct would be to return an object, though, instead of a dictionary. We could create a custom class (e.g. `DateMatch`, `ParsedDate`), and return an instance of it. That way, we have room for deprecation logic in the future, as well as lazy loading where needed.

rmedaer:
It might be also useful to add `time` as possible missing field:

```python3
        self.missing = [field for field in ('day', 'month', 'year', 'time')
                   if not getattr(self, field)]
```

So you know if the time is coming from parsed hour/minute. For instance, both `04/09/2020 12AM` and `04/09/2020` return the same and we don't know if a time did match or not..

noviluni:
I added the `DateData` class (https://github.com/scrapinghub/dateparser/pull/778), so I will wait until it's merged before proceeding with this

Twister7:
Hi looks like (#778) got merged, is there any chance to have this change merged?

PR #730, opened 2020-06-26
fix E2XX and E3XX flake8 rules and ignore W503

noviluni:


docs/conf.py
```diff
@@ -20,7 +20,7 @@
 # directory, add these directories to sys.path here. If the directory is
```
Gallaecio, line 20: In this file, I would instead all the sections with commented-out code.

They exist to make the initial configuration file creation easy, but they are not needed after that, and they can even be misleading (as you could think these are all the settings there are, but there may be upstream changes that add, remove or change the default values of some of these settings).

pytest.ini
```diff
@@ -7,6 +7,8 @@ addopts =
     --ignore=dateparser/calendars/hijri_parser.py
     --ignore=dateparser/calendars/hijri.py
 flake8-ignore =
+    W503
```
Gallaecio, line 10: Maybe include a comment to indicate why it is ignored?

tests/test_freshness_date_parser.py
```diff
@@ -252,7 +252,7 @@ def now_with_timezone(self, tzinfo):
         param('2 tuần 3 ngày', ago={'weeks': 2, 'days': 3}, period='day'),
         # following test unsupported, refer to discussion at:
         # http://github.com/scrapinghub/dateparser/issues/33
-        #param('1 năm 1 tháng 1 tuần 1 ngày 1 giờ 1 chút',
+        # param('1 năm 1 tháng 1 tuần 1 ngày 1 giờ 1 chút',
```
Gallaecio, line 255: Maybe we should extract this into an `xfail`-marked test method.

docs/conf.py
noviluni: agree!

pytest.ini
noviluni: Done!

tests/test_freshness_date_parser.py
noviluni: There are other similar cases (commented tests) and I prefer not to fix all of them in this PR. If you want we can open a new issue to handle this.

PR #732, opened 2020-06-26
add Python 3.9 support

noviluni:
It will be released the 05/09/2020, just before the next planned release of `dateparser`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,9 @@ matrix:
     - python: 3.8
       env: TOXENV=py38
       dist: xenial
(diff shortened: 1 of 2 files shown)


noviluni:
should I change the used distro (xenial)? it is, in fact, necessary? :thinking:

Gallaecio:
> should I change the used distro (xenial)? it is, in fact, necessary?

I would remove the distro from all entries as long as test pass. In Scrapy we have a few jobs where it is needed, but in Dateparser it may not be needed at all.

PR #735, opened 2020-07-01
Fix flake8 line length e501

noviluni:
~~waiting for this to be merged: https://github.com/scrapinghub/dateparser/pull/731~~

----

I selected 119 as the max default line length because it's used by Scrapy and Django, and this is because it's the default line length supported by GitHub without needing to do horizontal scrolling.

I have some doubts when accepting exceptions so don't be afraid and comment if you see anything that could be improved.

### DIFF
--- a/dateparser/conf.py
+++ b/dateparser/conf.py
@@ -21,7 +21,7 @@ class Settings(object):
     Default: False.
     * `PARSERS`: list of date parsers to use, in order of preference. Default:
     :attr:`dateparser.settings.default_parsers`.
-    """
+    """ # noqa E501
 
     _default = True
     _pyfile_data = None
--- a/dateparser/conf.py
+++ b/dateparser/conf.py
@@ -14,14 +14,15 @@ class Settings(object):
     * `PREFER_DATES_FROM`: defaults to `current_period`. Options are `future` or `past`.
     * `PREFER_DAY_OF_MONTH`: defaults to `current`. Could be `first` and `last` day of month.
     * `SKIP_TOKENS`: defaults to `['t']`. Can be any string.
-    * `TIMEZONE`: defaults to `UTC`. Can be timezone abbreviation or any of `tz database name as given here <https://en.wikipedia.org/wiki/List_of_tz_database_time_zones>`_.
+    * `TIMEZONE`: defaults to `UTC`. Can be timezone abbreviation or any of
+      `tz database name as given here <https://en.wikipedia.org/wiki/List_of_tz_database_time_zones>`_.
     * `RETURN_AS_TIMEZONE_AWARE`: return tz aware datetime objects in case timezone is detected in the date string.
     * `RELATIVE_BASE`: count relative date from this base date. Should be datetime object.
     * `RETURN_TIME_AS_PERIOD`: returns period as `time` in case time component is detected in the date string.
     Default: False.
     * `PARSERS`: list of date parsers to use, in order of preference. Default:
     :attr:`dateparser.settings.default_parsers`.
-    """ # noqa E501
+    """
 
     _default = True
     _pyfile_data = None
--- a/dateparser/languages/locale.py
+++ b/dateparser/languages/locale.py
@@ -467,9 +467,10 @@ class Locale(object):
 
     def _set_splitters(self, settings=None):
         splitters = {
-            'wordchars': set(),  # The ones that split string only if they are not surrounded
-            # by letters from both sides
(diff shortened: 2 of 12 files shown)


dateparser/conf.py
```diff
@@ -21,7 +21,7 @@ class Settings(object):
     Default: False.
     * `PARSERS`: list of date parsers to use, in order of preference. Default:
     :attr:`dateparser.settings.default_parsers`.
-    """
+    """ # noqa E501
```
Gallaecio, line 24: Actually, I think this may be a good opportunity to improve the formatting of that docstring.

dateparser/search/__init__.py
```diff
@@ -45,7 +45,7 @@ def search_dates(text, languages=None, settings=None, add_detected_language=Fals
          ('on May 6th 2004', datetime.datetime(2004, 5, 6, 0, 0))]
 
 
-        """
+        """ # noqa E501
```
Gallaecio, line 48: In this case, it should also be possible to refactor the code, using multiple lines for the input line. And given that it’s code, so line breaks won’t be added as part of the documentation rendering, I would think it’s important to break that line.

noviluni: Done!

dateparser/timezones.py
```diff
@@ -10,9 +10,11 @@
         'replace':
             [
                 (r'(?:UTC|GMT)\\(\+|\-)0(\d):00', r'(?:UTC|GMT)\\\1\2'),  # UTC+n, UTC-n, GMT+n, GMT-n
-                (r'(?:UTC|GMT)\\(\+|\-)0(\d):(\d{2})', r'(?:UTC|GMT)\\\1\2:\3'),  # UTC+n:mm, UTC-n:mm,GMT+n:mm, GMT-n:mm
+                (r'(?:UTC|GMT)\\(\+|\-)0(\d):(\d{2})', r'(?:UTC|GMT)\\\1\2:\3'),
+                # UTC+n:mm, UTC-n:mm, GMT+n:mm, GMT-n:mm
```
Gallaecio, line 14: I believe inline comments are usually moved to the line above, not below. You can append a `:` at the end of the comment if you think it makes things more clear.

dateparser/languages/locale.py
```diff
@@ -467,7 +467,8 @@ def _get_splitters(self, settings=None):
 
     def _set_splitters(self, settings=None):
         splitters = {
-            'wordchars': set(),  # The ones that split string only if they are not surrounded by letters from both sides
+            'wordchars': set(),  # The ones that split string only if they are not surrounded
+            # by letters from both sides
```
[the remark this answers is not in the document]
eamanu: +1 for the lastt change

tests/test_utils_strptime.py
```diff
@@ -72,12 +72,18 @@ def test_parsing_date_should_fail_using_datetime_strptime_if_locale_is_non_engli
...
+        ),
         param('12 Dec 10 10:30:55.000001', '%d %b %y %H:%M:%S.%f', expected=datetime(2010, 12, 12, 10, 30, 55, 1)),
         param('12 Dec 10 10:30:55.000011', '%d %b %y %H:%M:%S.%f', expected=datetime(2010, 12, 12, 10, 30, 55, 11)),
         param('12 Dec 10 10:30:55.000111', '%d %b %y %H:%M:%S.%f', expected=datetime(2010, 12, 12, 10, 30, 55, 111)),
         param('12 Feb 2016 11:41:23', '%d %b %Y %I:%M:%S', expected=datetime(2016, 2, 12, 11, 41, 23)),
-        param('11 Dec 10 10:30:2011.999999', '%y %b %S %H:%M:%Y.%f', expected=datetime(2011, 12, 1, 10, 30, 10, 999999)),
+        param(
+            '11 Dec 10 10:30:2011.999999', '%y %b %S %H:%M:%Y.%f',
+            expected=datetime(2011, 12, 1, 10, 30, 10, 999999)
+        ),
```
eamanu, line 86: what about make the same change in the other lines?

noviluni: @eamanu because it would add a lot of noise to this PR, making it hard to read.

dateparser/languages/locale.py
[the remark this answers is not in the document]
noviluni: @Gallaecio I don't totally like these changes. As you mentioned in a comment below, inline comments are usually moved above, appending a colon, so I did this here too. Let me know what you think :+1:

Issue #679, opened 2020-05-14
TypeError: can't compare offset-naive and offset-aware datetimes


LeptoFlare:
This issue is not isolated inside the function. It permeates outside of the function.

It seems to start occurring after you use search_dates on relative and absolute datetimes at the same time.

Just use search_dates on this string:
```
May 2020
June 2020
2023
January UTC
June 5 am utc
June 23th 5 pm EST
May 31, 8am UTC
```

LeptoFlare:
```py
>>> from dateparser.search import search_dates
>>> search_dates("""May 2020
... June 2020
... 2023
... January UTC
... June 5 am utc
... June 23th 5 pm EST
... May 31, 8am UTC""")

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/leptospira/lohi/.venv/lib/python3.8/site-packages/dateparser/search/__init__.py", line 49, in search_dates
    result = _search_with_detection.search_dates(
  File "/home/leptospira/lohi/.venv/lib/python3.8/site-packages/dateparser/conf.py", line 84, in wrapper
    return f(*args, **kwargs)
  File "/home/leptospira/lohi/.venv/lib/python3.8/site-packages/dateparser/search/search.py", line 229, in search_dates
    return {'Language': language_shortname, 'Dates': self.search.search_parse(language_shortname, text,
  File "/home/leptospira/lohi/.venv/lib/python3.8/site-packages/dateparser/search/search.py", line 161, in search_parse
    parsed, substrings = self.parse_found_objects(parser=parser, to_parse=translated,
  File "/home/leptospira/lohi/.venv/lib/python3.8/site-packages/dateparser/search/search.py", line 125, in parse_found_objects
    parsed_item = self.parse_item(parser, item, translated[i], parsed, need_relative_base)
  File "/home/leptospira/lohi/.venv/lib/python3.8/site-packages/dateparser/search/search.py", line 111, in parse_item
    parsed_item = parser.get_date_data(item)
  File "/home/leptospira/lohi/.venv/lib/python3.8/site-packages/dateparser/date.py", line 417, in get_date_data
    parsed_date = _DateLocaleParser.parse(
  File "/home/leptospira/lohi/.venv/lib/python3.8/site-packages/dateparser/date.py", line 196, in parse
    return instance._parse()
  File "/home/leptospira/lohi/.venv/lib/python3.8/site-packages/dateparser/date.py", line 200, in _parse
[...]

PR #736, opened 2020-07-01
fix comparing naive and aware

noviluni:
This is a workaround to fix this: https://github.com/scrapinghub/dateparser/issues/679

However it's necessary to test it properly before merging it.


noviluni:
related to this: https://github.com/scrapinghub/dateparser/pull/814

Issue #520, opened 2019-04-30
wrong lead year


HQCfly:
When I type `dateparser.parse("2/29 10:30:02")`, the result returned to me is 2029-02-28 10:30:02. I think the correct answer should be the last leap year.

Gallaecio:
I’m unsure. I would expect these from return the same year:

    dateparser.parse("2/28 10:30:02")
    dateparser.parse("2/29 10:30:02")
    dateparser.parse("3/1 10:30:02")

lopuhin:
> I think the correct answer should be the last leap year.

+1 that this is not obvious, e.g. why not the next leap year, especially if it's closer?

Maybe it would be great to be able to know that they year is unknown.

Issue #693, opened 2020-05-20
Leap year not working as expected with `'PREFER_DATES_FROM'


noviluni:
I tried to execute:

```python3
dateparser.parse("29/2 10:30:02", settings={'PREFER_DATES_FROM': 'future'})
```

and it returns None.

kishan3:
Hi @noviluni I am getting response in the latest version `0.7.4` 

```
In [1]: import dateparser

In [2]: dateparser.__version__
Out[2]: '0.7.4'

In [3]: dateparser.parse("29/2 10:30:02", settings={'PREFER_DATES_FROM': 'future'})
   ...: 
Out[3]: datetime.datetime(2029, 2, 9, 10, 30, 2)

```

noviluni:
hhmmmm... I don't know why I was receiving `None`, but even if it returns something, I think that the result is wrong, as it's interpreting the 29 as a year without any apparent reason.

Look these two examples, especially the second one.

```python
>>> dateparser.parse("29 Feb", settings={'PREFER_DATES_FROM': 'future'})                                                  
datetime.datetime(2029, 2, 10, 0, 0)

>>> dateparser.parse("29 Feb", settings={'PREFER_DATES_FROM': 'past'})                                                    
datetime.datetime(2020, 2, 29, 0, 0)
```

```python
>>> dateparser.parse("29 Feb", settings={'PREFER_DATES_FROM': 'future','RELATIVE_BASE': datetime.datetime(2020, 1, 1)})  
datetime.datetime(2020, 2, 29, 0, 0)

>>> dateparser.parse("29 Feb", settings={'PREFER_DATES_FROM': 'past','RELATIVE_BASE': datetime.datetime(2020, 1, 1)})    
datetime.datetime(1929, 2, 1, 0, 0)
```

I think we should check it carefully.

Related with: https://github.com/scrapinghub/dateparser/issues/520

kishan3:
Right, Maybe we can update the issue so it reflects the actual issue. What do you think?

Gallaecio:
Fixed by #738

PR #738, opened 2020-07-02
Fix leap years when PREFER_DATES_FROM is set

noviluni:
fixes: https://github.com/scrapinghub/dateparser/issues/520, https://github.com/scrapinghub/dateparser/issues/693

### DIFF
--- a/dateparser/parser.py
+++ b/dateparser/parser.py
@@ -8,7 +8,7 @@ from datetime import datetime
 from datetime import timedelta
 
 from dateparser.utils import set_correct_day_from_settings, \
-    get_last_day_of_month
+    get_last_day_of_month, get_previous_leap_year, get_next_leap_year
 from dateparser.utils.strptime import strptime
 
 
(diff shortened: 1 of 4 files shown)


dateparser/parser.py
```diff
@@ -316,14 +316,29 @@ def _get_datetime_obj(self, **params):
...
+            year = get_next_leap_year(current_year)
+        elif prefer_dates_from == 'past':
+            year = get_previous_leap_year(current_year)
+        else:  # 'current_period': return closer leap year
+            next_leap_year = get_next_leap_year(current_year)
+            previous_leap_year = get_previous_leap_year(current_year)
+            year = next_leap_year \
+                if next_leap_year - current_year < current_year - previous_leap_year \
+                else previous_leap_year
+        return year
```
Gallaecio, line 341: :lipstick: This could be implemented with early returns.

dateparser/parser.py
```diff
@@ -415,12 +430,22 @@ def _correct_for_time_frame(self, dateobj):
             dateobj = dateobj + delta
 
         if self.month and not self.year:
-            if self.now < dateobj:
-                if 'past' in self.settings.PREFER_DATES_FROM:
-                    dateobj = dateobj.replace(year=dateobj.year - 1)
-            else:
-                if 'future' in self.settings.PREFER_DATES_FROM:
-                    dateobj = dateobj.replace(year=dateobj.year + 1)
+            is_leap_date = dateobj.day == 29 and dateobj.month == 2
```
Gallaecio, line 433: :lipstick: This probably does not need to be saved into a variable, and in any case it should be moved to the `except ValueError` section below.

dateparser/utils/__init__.py
```diff
@@ -139,6 +139,26 @@ def get_last_day_of_month(year, month):
...
+
+def _get_leap_year(year, future):
+    """
+    Iterate through previous or next years until it gets a valid leap year
+    This is performed to avoid missing or including centurial leap years
+    """
+    difference = 1 if future else -1
+    leap_year = year + difference
+    while not calendar.isleap(leap_year):
+        leap_year = leap_year + difference
```
Gallaecio, line 158: :lipstick: 

```suggestion
        leap_year += difference
```

dateparser/utils/__init__.py
```diff
@@ -139,6 +139,26 @@ def get_last_day_of_month(year, month):
...
+def get_next_leap_year(year):
+    return _get_leap_year(year, future=True)
+
+
+def _get_leap_year(year, future):
+    """
+    Iterate through previous or next years until it gets a valid leap year
+    This is performed to avoid missing or including centurial leap years
+    """
+    difference = 1 if future else -1
```
Gallaecio, line 155: :lipstick: I’m unsure about this one, but since I’m going full style suggestions, 1 more should not hurt :slightly_smiling_face: 

```suggestion
    step = 1 if future else -1
```

noviluni: hahaha it's ok

dateparser/parser.py
```diff
@@ -8,7 +8,7 @@
 from datetime import timedelta
 
 from dateparser.utils import set_correct_day_from_settings, \
-    get_last_day_of_month
+    get_last_day_of_month, get_previous_leap_year, get_next_leap_year
```
eamanu, line 11: Why don't write one method by line?, this way if we need add an import, just modify a line and avoid possible erros

noviluni: I don't see the point...

If you check the PEP8 guidelines, this is the normal way of using imports:
imagen
ref: https://www.python.org/dev/peps/pep-0008/#imports

Of course, I can use tuples to do multi-line imports, but that's normally used only when there are long names to be imported or heavy nested structures. 

In this case, I think that what I did is totally correct.

noviluni: @eamanu (comment above)

PR #739, opened 2020-07-03
Delete requirements.txt and Scrapy Cloud references

kishan3:
This should fix #724

Issue #727, opened 2020-06-25
Deprecate Python 2.7


noviluni:
As part of the next version 1.0.0 we will deprecate Python 2.7. I have just created this ticket to track it.

- [X] Merge as more finished PRs as possible (to avoid future file conflicts)
- [X] Change tox and Travis to test only with Python 3. (ref: https://github.com/scrapinghub/dateparser/pull/744)
- [x] Change the setup file. (ref: https://github.com/scrapinghub/dateparser/pull/755)
- [x] Apply 2to3 and review carefully (`__future__` imports, unicode strings...) (ref: https://github.com/scrapinghub/dateparser/pull/748)
- [x] Remove six package (ref: https://github.com/scrapinghub/dateparser/pull/754)
- [x] Remove utf declaration strings (ref: https://github.com/scrapinghub/dateparser/pull/758, #761, #764)
- [x] Check specific code for Python 2 that can be removed or refactored (https://github.com/scrapinghub/dateparser/pull/763, https://github.com/scrapinghub/dateparser/pull/749, https://github.com/scrapinghub/dateparser/pull/777)
- [x] Check and update docs and comments (https://github.com/scrapinghub/dateparser/pull/783)

kishan3:
Hi @noviluni Is it possible to list tasks we need to do perform deprecation?
Or we just need to check it file by file?

noviluni:
Hi @kishan3, we have some pending PRs labeled as `breaking-change` (https://github.com/scrapinghub/dateparser/labels/breaking-change) that I'm waiting to merge after seeing that this last release is stable enough. If we don't see any major issue raised before the 15th of July, we will merge those PRs and after that starting to migrate the library to Python 3 code (deprecating Python 2.7).

My idea is to:
* Merge as more finished PRs as possible (to avoid future file conflicts)
* Change tox and Travis to test only with Python 3.
* Change the setup file.
* Proceed with code changes.

From my experience, the best way to proceed is by using the `2to3' library (reviewing manually each change as some things are not always necessary). After that, we will remove the `six` package, remove the utf strings and check if there are specific code for Python 2 or Python 3 that can be removed or refactored.

To avoid really big PRs, I think that we could start by using 2to3 in some specific sections (for example in tests, in scripts, etc.).

noviluni:
Today is 15th, so we can start with the migration :muscle: 

I will spend some time merging PRs marked as "breaking change" to avoid too many conflicts when removing Python 2 reminiscents.

[...]

Issue #565, opened 2019-09-11
Pypy (2) breaks on importing parse


Bachmann1234:
When installing dateparser the most recent version of the regex library is installed which seems to break in pypy (works in pypy3 in my experiments)

When I uninstall it and install the version in your requirements.txt the import seems to work fine. Here is my session

```
venv ❯ pypy
Python 2.7.13 (8cdda8b8cdb8ff29d9e620cccd6c5edd2f2a23ec, Sep 04 2019, 09:54:49)
[PyPy 7.1.1 with GCC 4.2.1 Compatible Apple LLVM 10.0.1 (clang-1001.0.46.4)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>>> from dateparser import parse
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/bachmann/Desktop/venv/site-packages/dateparser/__init__.py", line 4, in <module>
    from .date import DateDataParser
  File "/Users/bachmann/Desktop/venv/site-packages/dateparser/date.py", line 13, in <module>
    from dateparser.date_parser import date_parser
  File "/Users/bachmann/Desktop/venv/site-packages/dateparser/date_parser.py", line 8, in <module>
    from .timezone_parser import pop_tz_offset_from_string
  File "/Users/bachmann/Desktop/venv/site-packages/dateparser/timezone_parser.py", line 79, in <module>
    _tz_offsets = list(get_tz_offsets())
  File "/Users/bachmann/Desktop/venv/site-packages/dateparser/timezone_parser.py", line 70, in get_tz_offsets
    yield get_offset(tz_obj, regex, repl=replace, replw=replacewith)
  File "/Users/bachmann/Desktop/venv/site-packages/dateparser/timezone_parser.py", line 57, in get_offset
    'regex': re.compile(re.sub(repl, replw, regex % tz_obj[0]), re.IGNORECASE),
  File "/Users/bachmann/Desktop/venv/site-packages/regex/regex.py", line 276, in sub
    endpos, concurrent)
ImportError: No module named regex.regex.regex
>>>>

~/Desktop 26s
venv ❯ pip freeze
DEPRECATION: A future version of pip will drop support for Python 2.7. More details about Python 2 support in pip, can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support
cffi==1.12.3
dateparser==0.7.1
greenlet==0.4.13
python-dateutil==2.8.0
pytz==2019.2
[...]

PR #740, opened 2020-07-03
Fix `calendars` extra

noviluni:
I noticed that `pip install dateparser[calendars]` didn't work. After some time investigating, it seems that there was a typo in `extras_require`.

I also updated the `tox` file after this change and the docs (that were wrong). The `requirements.txt` file will be deleted in this PR: https://github.com/scrapinghub/dateparser/pull/739

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,8 +35,9 @@ setup(
         'regex !=2019.02.19',
         'tzlocal',
     ],
-    extra_requires={
-        'calendars': ['convertdate', 'umalqurra', 'jdatetime', 'ruamel.yaml'],
+    extras_require={
+        'calendars:python_version<="2.7"': ['umalqurra', 'convertdate'],
+        'calendars:python_version>"2.7"': ['convertdate']
     },
     license="BSD",
     zip_safe=False,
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -35,8 +35,9 @@
         'regex !=2019.02.19',
         'tzlocal',
     ],
-    extra_requires={
-        'calendars': ['convertdate', 'umalqurra', 'jdatetime', 'ruamel.yaml'],
+    extras_require={
+        'calendars:python_version<="2.7"': ['umalqurra', 'convertdate'],
+        'calendars:python_version>"2.7"': ['convertdate']
```
Gallaecio, line 40: :lipstick: 

```suggestion
        'calendars:python_version>"2.7"': ['convertdate'],
```

noviluni:
I added a temporary workaround (pinning a working `regex` package version)  to make the `Pypy` pipeline pass.

As mentioned here: https://github.com/scrapinghub/dateparser/issues/565, it seems that `dateparser` doesn't work with `Pypy` out of the box because of the `regex` library which we depend on.

The next version will remove Python 2.7 support: https://github.com/scrapinghub/dateparser/issues/727 (and Pypy), so I hope we will remove this requirement pinning before releasing the next version.

I will investigate if this affects `pypy3` to know if we can support it or not.

PR #744, opened 2020-07-15
Remove pypy and python2 from tox and travis.

kishan3:
This should help one step in closing #727

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,10 +6,6 @@ matrix:
   include:
     - python: 3.8
       env: TOXENV=flake8
-    - python: 2.7
-      env: TOXENV=py27
-    - python: pypy
-      env: TOXENV=pypy
     - python: 3.5
       env: TOXENV=py35
     - python: 3.6
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = flake8, py2, py3
+envlist = flake8, py3
 
 [testenv]
 deps =


Gallaecio:
The coverage data reveals a few parts of the code that are specific to Python 2 and can be removed.

lopuhin:
re pypy removal - maybe it makes sense to add pypy3 then, maybe as part of another PR?

noviluni:
Hi @lopuhin, I tried it and the tests were totally broken. The `regex` package doesn't have official support for pypy and pypy3. I would love to have support for pypy3, but I think it's not as easy as it seems (maybe it's not possible) and it should be added in another, different, PR. Thank you for your comment, I opened this new ticket to discuss it: https://github.com/scrapinghub/dateparser/issues/745

noviluni:
Anyway, this is the first step on a list of steps to get rid of Python 2.

We will:
* Remove testing (this will make it easier to create new PRs without needing to support Python 2.7)
* Update setup file
* Migrate code (probably using 2to3 and checking it carefully)
* Delete other reminiscences (the `six` package, version-specific code...)

We don't need to perform these steps in this strict order, but I think it's a good way to go, let me know if there is any wrong on this approach or if you have any other idea.


@Gallaecio @lopuhin

lopuhin:
Thanks @noviluni yes makes sense to tackle pypy3 separately. Re the steps you outlined - they make sense, I don't have experience with 2to3, only with manual porting, but usually py2-specific code stands out.

kishan3:
Hi @noviluni @Gallaecio we should merge this only after coverage is fixed or can do now?

Gallaecio:
> Hi @noviluni @Gallaecio we should merge this only after coverage is fixed or can do now?

I think we can merge already and tackle the coverage part separately, as @noviluni suggested.

noviluni:
Good job! Now we can say that we don't support Python 3 officially :rofl:

PR #747, opened 2020-07-17
[WIP] Use 2to3 library on files under tests module.

kishan3:
This should help a step in closing #727

### DIFF
--- a/tests/test_data.py
+++ b/tests/test_data.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-from __future__ import unicode_literals
+
 
 from itertools import chain
 import regex as re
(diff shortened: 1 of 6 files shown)


tests/test_data.py
```diff
@@ -13,7 +13,7 @@
 DEFAULT_MONTH_PATTERN = re.compile(r'^M?\d+$', re.U)
 INVALID_AM_PM_PATTERN = re.compile(r'^[AaPp]\.?\s+[Mm]$')
 
-all_locale_shortnames = list(chain(language_locale_dict.keys(), *language_locale_dict.values()))
+all_locale_shortnames = list(chain(list(language_locale_dict.keys()), *list(language_locale_dict.values())))
```
Gallaecio, line 16: I understand why 2to3 does it, to err on the safe side, but I believe these changes are not needed.

I think 2to3 is meant mainly to convert code that is Python-2-only to Python-3-only. So, while the output provides invaluable information on potential cleanup (those `from __future__` statements), I don’t think we should apply its changes _as is_, as most of it is likely to be noise on a project that already supports Python 3.

kishan3: Thanks @Gallaecio 
Then I think we don't need to make this change (along with removing `from __future__` ) statements 
Probably I should delete this PR in this case.

noviluni: Hi @Gallaecio, I totally agree with you, I personally think that 2to3 it's really helpful to remove some things coming from Python 2, but, as described in the original issue (https://github.com/scrapinghub/dateparser/issues/727#issuecomment-651078076), we should check the output to see if a change it's necessary or not:

> From my experience, the best way to proceed is by using the 2to3 library (reviewing manually each change as some things are not always necessary). After that, we will remove the six package, remove the utf declaration strings and check if there are specific code for Python 2 or Python 3 that can be removed or refactored.

I opened a PR with the necessary changes: https://github.com/scrapinghub/dateparser/pull/748, so we can close this.

noviluni: BTW, I have just added a checklist in the issue: https://github.com/scrapinghub/dateparser/issues/727

PR #748, opened 2020-07-17
Apply 2to3 to clean code

noviluni:
closes: https://github.com/scrapinghub/dateparser/pull/747

### DIFF
--- a/dateparser/calendars/__init__.py
+++ b/dateparser/calendars/__init__.py
@@ -1,6 +1,6 @@
 # coding: utf-8
 
-from __future__ import unicode_literals
+
 from dateparser.parser import _parser
 from dateparser.conf import settings
 from datetime import datetime
--- a/dateparser/freshness_date_parser.py
+++ b/dateparser/freshness_date_parser.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-from __future__ import unicode_literals
+
 
 import regex as re
 from datetime import datetime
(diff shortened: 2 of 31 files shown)


dateparser/calendars/__init__.py
```diff
@@ -1,6 +1,6 @@
 # coding: utf-8
```
Gallaecio, line 1: Unless you prefer to do it in a later changeset, this may be a good chance to get rid of these encoding comments as well.

And that way we don‘t need to discuss whether the empty lines between those and the first import line should be 1 or 2 :slightly_smiling_face:

dateparser/freshness_date_parser.py
```diff
@@ -29,9 +29,9 @@ def _are_all_words_units(self, date_string):
 
         date_string = re.sub(r'\s+', ' ', date_string.strip())
 
-        words = filter(lambda x: x if x else False, re.split(r'\W', date_string))
-        words = filter(lambda x: not re.match(r'%s' % '|'.join(skip), x), words)
-        return not list(words)
+        words = [x for x in re.split(r'\W', date_string) if (x if x else False)]
```
Gallaecio, line 32:
```suggestion
        words = [x for x in re.split(r'\W', date_string) if x]
```

dateparser/calendars/__init__.py
noviluni: YeahYeah, this will be handled in another PR: #758 :+1:

PR #749, opened 2020-07-17
remove some python2 specific code

noviluni:
I fixed some cases with Python 2 specific code that were detected when removing Python 2 from the pipelines (https://github.com/scrapinghub/dateparser/pull/744) (https://codecov.io/gh/scrapinghub/dateparser/pull/744/changes)

### DIFF
--- a/dateparser/date.py
+++ b/dateparser/date.py
@@ -1,5 +1,5 @@
 import collections
-from collections import Set
+from collections.abc import Set
 from datetime import datetime, timedelta
 
 import regex as re
--- a/dateparser/date.py
+++ b/dateparser/date.py
@@ -1,4 +1,5 @@
 import collections
+from collections import Set
 from datetime import datetime, timedelta
 
 import regex as re
@@ -153,7 +154,7 @@ class _DateLocaleParser(object):
 
     def __init__(self, locale, date_string, date_formats, settings=None):
         self._settings = settings
-        if not (date_formats is None or isinstance(date_formats, (list, tuple, set))):
+        if not (date_formats is None or isinstance(date_formats, (list, tuple, Set))):
             raise TypeError("Date formats should be list, tuple or set of strings")
 
         self.locale = locale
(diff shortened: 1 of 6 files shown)


dateparser/search/search.py
```diff
@@ -180,7 +174,7 @@ def __init__(self):
         self.search = ExactLanguageSearch(self.loader)
 
     def detect_language(self, text, languages):
-        if isinstance(languages, (list, tuple, Set)):
+        if isinstance(languages, (list, tuple, set)):
```
Gallaecio, line 177: I think this is technically a backward-incompatible change, as it is technically possible for an object to be an instance of a class inheriting from `Set` that is not a `set`.

dateparser/utils/__init__.py
```diff
@@ -18,9 +18,6 @@ def strip_braces(date_string):
 
 
 def normalize_unicode(string, form='NFKD'):
-    if isinstance(string, bytes):
```
Gallaecio, line 21: Backward-incompatible.

dateparser/utils/__init__.py
```diff
@@ -44,25 +41,6 @@ def combine_dicts(primary_dict, supplementary_dict):
     return combined_dict
 
 
-def convert_to_unicode(info):
```
Gallaecio, line 47: Backward-incompatible.

dateparser/utils/__init__.py
noviluni: As we are going to release a new major version I think it's a good moment to make these backward-incompatible changes. What do you think? 

Apart from that, we could also move some of these functions to the private scope by adding a `_`.

Gallaecio: Either works. We just need to take this into account when writing the release notes, to make sure we cover backward-incompatible changes.

noviluni:
I reverted the change in the `Set` but also the changes introduced by this PR: #509 because I think it's ok to accept `frozenset` and other "set types".

dateparser/date.py
```diff
@@ -1,4 +1,5 @@
 import collections
+from collections import Set
```
Gallaecio, line 2: > DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working

Gallaecio (approved):

Issue #745, opened 2020-07-16
Support for Pypy3?


noviluni:
I tried to add the Pypy3 env to tox and Travis and the tests were failing because the `regex` package doesn't support Pypy/Pypy3 officially.

Should we add support for Pypy3? should we try to remove the dependency on the `regex` package (I tried in the past and it's not as simple as it seems)?

Feedback welcomed

lopuhin:
I think pypy build passes in master https://travis-ci.org/github/scrapinghub/dateparser/jobs/708282459 and from regex bugtracker it seems it works on pypy3 (although there may be some issues with complex unicode handling), so to me it seems that removing regex package is not the best option.

noviluni:
Hi @lopuhin Thanks for your answer

Pypy was passing because when testing we were installing a pinned version of the `regex` package from the `requirements.txt` file, but it didn't work with the last regex version. There was an issue about this but it wasn't investigated so far: https://github.com/scrapinghub/dateparser/issues/565, and after removing the `requirements.txt` file we found the same issue, so I pinned a working `regex` version in the `setup.py` file to allow the pipelines to pass: `'regex==2019.01.24; platform.python_implementation == "PyPy"',` (added in: https://github.com/scrapinghub/dateparser/pull/740)

As we are releasing a new version without Python 2 support we will remove support for Pypy too. I tried adding the Pypy3 env but the pipelines are failing. You can see more in the first comment here: https://github.com/scrapinghub/dateparser/pull/746 As you can see is not just a test failing.

I would love to get a PR fixing those issues, however, I've never seen any issue requesting for pypy3 support, so:
* Do you think is required / really useful to add Pypy3 support?
* It's possible to fix it? 

About removing the regex package, I know it's not trivial (and the performance could be affected), but I found some projects that stoped using `dateparser` because of regex:  
* https://github.com/python-discord/bot/pull/444 (first comment)
* https://github.com/karlicoss/promnesia/commit/1daff43867ffae998e73a79e4025e15b2a4113e7
* etc.

And it seems that it's because it has to be compiled from the sources when installing it and that's something to deal with in some scenarios. This fact and the pypy issues could be enough to at least consider if it's viable to remove the dependency or not.

I don't have too much experience on any of these matters, so if you have any further insight please, share it with me :)

lopuhin:
[...]

Issue #441, opened 2018-08-07
Library is not thread safe


meownoid:
Info:
```
Linux 4.15.0-24-generic #26-Ubuntu SMP Wed Jun 13 08:44:47 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```
```
Python 3.7.0b3
```

```
>>> import dateparser
>>> dateparser.__version__
'0.7.0'
```

Code to reproduce error:
```
import dateparser
from concurrent.futures.thread import ThreadPoolExecutor

fs = []
with ThreadPoolExecutor(16) as executor:
    for _ in range(100):
        fs.append(executor.submit(lambda: dateparser.parse('tomorrow')))

for f in fs:
    print(f.result())
```

Error:
```
Traceback (most recent call last):
  File "test.py", line 10, in <module>
    print(f.result())
  File "/usr/lib/python3.7/concurrent/futures/_base.py", line 425, in result
    return self.__get_result()
  File "/usr/lib/python3.7/concurrent/futures/_base.py", line 384, in __get_result
    raise self._exception
  File "/usr/lib/python3.7/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "test.py", line 7, in <lambda>
    fs.append(executor.submit(lambda: dateparser.parse('tomorrow')))
  File "/home/egor/Projects/rekko/.env/lib/python3.7/site-packages/dateparser/conf.py", line 81, in wrapper
    return f(*args, **kwargs)
  File "/home/egor/Projects/rekko/.env/lib/python3.7/site-packages/dateparser/__init__.py", line 53, in parse
    data = parser.get_date_data(date_string, date_formats)
  File "/home/egor/Projects/rekko/.env/lib/python3.7/site-packages/dateparser/date.py", line 404, in get_date_data
    locale, date_string, date_formats, settings=self._settings)
  File "/home/egor/Projects/rekko/.env/lib/python3.7/site-packages/dateparser/date.py", line 177, in parse
    return instance._parse()
  File "/home/egor/Projects/rekko/.env/lib/python3.7/site-packages/dateparser/date.py", line 187, in _parse
    date_obj = parser()
  File "/home/egor/Projects/rekko/.env/lib/python3.7/site-packages/dateparser/date.py", line 200, in _try_freshness_parser
    return freshness_date_parser.get_date_data(self._get_translated_date(), self._settings)
[...]

PR #755, opened 2020-07-28
Change setup.py

kishan3:
For #727


setup.py
```diff
@@ -44,14 +44,12 @@
     license="BSD",
     zip_safe=False,
     keywords='dateparser',
-    python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*',
+    python_requires='!=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*',
     classifiers=[
         'Development Status :: 2 - Pre-Alpha',
```
noviluni, line 49: I have just realized that we are in "Pre-Alpha" status.... :thinking:  
We should probably update this to be `Development Status :: 4 - Beta` or  `Development Status :: 5 - Production/Stable` @Gallaecio let me know your thoughts.

noviluni: I'm not totally confident in moving to "Production/Stable" because of the "multiple thread" issues: https://github.com/scrapinghub/dateparser/issues/441

Gallaecio: I’m not sure whether or not multithreading support is a requirement for Dateparser to be considered stable, but I think it makes sense to delay that tag until Dateparser 1.0 is released.

Gallaecio:
> Could you also remove this line: `'Programming Language :: Python :: Implementation :: PyPy'`? (as we are deprecating PyPy).

Based on the recent feedback from @lopuhin, I would wait on this. It’s starting to look to me like we may support the latest PyPy.

setup.py
noviluni: @Gallaecio  hmmm.. but these changes are performed to release that 1.0.0 version, so in that case, it makes sense to change it now. right?

noviluni:
@kishan3 , so we will be waiting to clarify the PyPy decision before proceeding.

If you want you can check this list: https://pypi.org/classifiers/ to see if there is something useful that we could add in the setup file.

setup.py
Gallaecio: Oh, I forgot 1.0 was the next version! In that case, I would go for that tag, yes. Stable does not mean feature-complete. It’s production-ready for people that do not need multiprocessing support.

setup.py
```diff
@@ -44,14 +44,12 @@
     license="BSD",
     zip_safe=False,
     keywords='dateparser',
-    python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*',
+    python_requires='!=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*',
```
noviluni, line 47:
```suggestion
    python_requires='>=3.5',
```

However, as the next month Python 3.5 will be deprecated, I think we will be supporting only Python > 3.6. If that's the case, we will need to remove also the `'Programming Language :: Python :: 3.5',` line.

noviluni (changes requested):
Hi @kishan3 !

as mentioned [here](https://github.com/scrapinghub/dateparser/issues/745#issuecomment-680075168), we will drop the support for PyPy, so you can do what I mentioned here: https://github.com/scrapinghub/dateparser/pull/755#pullrequestreview-456502138 (removing the PyPy line and changing the `install_requires`).

kishan3:
Hi @noviluni Thanks for the feedback! I have updated the setup.py file as per your suggestion.

noviluni:
Hi @kishan3, thanks!

but you have to remove this line too: `'Programming Language :: Python :: Implementation :: PyPy'`

kishan3:
Sorry @noviluni  I somehow missed it. It is removed now.

noviluni (approved):

Issue #500, opened 2019-02-01
search_dates not recognizing dd.mm.yyyy


anakin87:
`search_dates('test 13.07.2016 test')`

output:
`[('2016', datetime.datetime(2016, 2, 1, 0, 0))]`

How to make search_dates recognize this format?
****How to pass some date formats to search_dates**?**

Please help.

climberwoodi:
This fix would be very important to me. This format is very common in german texts.

PR #757, opened 2020-07-30
fix dates with period

noviluni:
fixes: https://github.com/scrapinghub/dateparser/issues/500

there is an old test failing, but I think that case is confusing and we should probably remove that test case. 

Let me know what you think about the failing test case @Gallaecio


noviluni:
I updated the test as initially I put it in the wrong place :sweat_smile:

Gallaecio:
Actually, I think we should aim to either not break that test or, at least, break it differently.

The fact that from “DECEMBER 21 19.87 87” we used to extract “DECEMBER 21 19” as a date and instead now we extract “DECEMBER” and “19.87” seems like a slight regression to me.

Issue #714, opened 2020-06-18
Improve the default language used order


noviluni:
When trying to parse a date if we don't pass any locale, language, or region, `dateparser` tries with different locales until it finds something that works.

The used order is defined here: https://github.com/scrapinghub/dateparser/blob/master/dateparser/data/languages_info.py, and that file is generated when running this script: `dateparser_scripts/order_languages.py`, that uses the most known/spoken language coming from [here](https://github.com/unicode-cldr/cldr-core/blob/master/supplemental/territoryInfo.json).

However, I think that this approach is not good enough and it could be improved. Even if this library can be used with a high number of different purposes, the main purpose is to parse dates coming from the internet. If you agree with this, it has more sense to try first with the most used languages on the internet, which are not the same as the most known/spoken languages.

**Here the list of the 10 most common languages on the internet** (font: https://w3techs.com/technologies/overview/content_language):
* en: 59.5%
* ru: 8.6%
* es: 4.0%
* tr: 3.1%
* fa: 2.9%
* de: 2.7% 
* fr: 2.7%
* ja: 2.3%
* pt: 2%
* vi: 1.5%

**Here the list of the 10 first languages we try**:
* en: 59.5%
* zh and zh-Hans: 1.5%
* hi: 0.1%
* es: 4.0%
* ar: 0.9%
* bn: <0.1%
* fr: 2.7%
* ur:  <0.1%
* pt: 2%
* ta: <0.1%

Why are we trying with languages used by less than 0.1% of websites like Urdu, Tamil, etc. before trying with Russian, that it's in the 8.6% of the websites, Turkish (3.1%) or Japanese (2.3%)?

I think that changing this won't have a really big impact for most of the situations (that use English or only one language and the used locale is "cached"), but it could impact other projects that are using `dateparser` to parse some other languages intensively.

Let me know what you think and how we could implement this. :)

noviluni:
Real-world case:

Me and @lopuhin analyzed a project where `dateparser` was used to parse dates coming from different sources in different languages, and I can confirm that the new order would be more efficient.

![image(1)](https://user-images.githubusercontent.com/22377678/85280178-eabc1200-b487-11ea-915d-22eee1f332e0.png)

[...]

PR #764, opened 2020-08-04
Remove encoding from ordering languages script

kishan3:
#727


Gallaecio:
Shouldn’t we also remove https://github.com/scrapinghub/dateparser/blob/master/dateparser/data/languages_info.py#L1 as part of this request, though?

kishan3:
Right @Gallaecio When I tried to run the script it gave me the following error. Am I running it wrong?
```
python dateparser_scripts/order_languages.py

Traceback (most recent call last):
  File "dateparser_scripts/order_languages.py", line 33, in <module>
    language_locale_dict = _get_language_locale_dict()
  File "dateparser_scripts/order_languages.py", line 17, in _get_language_locale_dict
    available_locale_names = os.listdir(cldr_dates_full_dir)
OSError: [Errno 2] No such file or directory: '../raw_data/cldr_dates_full/main/'
```

noviluni:
> Right @Gallaecio When I tried to run the script it gave me the following error. Am I running it wrong?

These scripts are not normally used and should be checked (some parts as the path handlings are really broken), but the last time I used this script, it worked. 

If this error is happening, what's probably happening is that you have an empty/wrong `raw_data` directory. Please, check if you have that directory in the project or even outside (at the same level as the project) and delete it. It should work after that.

However, I don't recommend you to run this script, as it updates the order of the languages (downloading the last version of the repo from Github) and we don't want this to happen until we address this: https://github.com/scrapinghub/dateparser/issues/714, so you can remove that line directly in the `languages_info.py` file.

Let me know if you have any doubt. :slightly_smiling_face: 

@kishan3

kishan3:
Thanks @noviluni for detailed explanation :rocket: I have updated the manually.

PR #778, opened 2020-09-07
create DateData class

noviluni:
This was motivated by this comment: https://github.com/scrapinghub/dateparser/pull/729#issuecomment-652940233

Changing from returning a dict to returning an object seems a better decision, as it gives more control over the results and allows to document the API easier.

This should probably be improved by adding more docs, etc.

This is blocking:
* https://github.com/scrapinghub/dateparser/pull/729
* https://github.com/scrapinghub/dateparser/pull/561


dateparser/calendars/__init__.py
```diff
@@ -19,7 +20,7 @@ def __init__(self, source):
     def get_date(self):
         try:
             date_obj, period = self.parser.parse(self.source, settings)
-            return {'date_obj': date_obj, 'period': period}
+            return DateData(**{'date_obj': date_obj, 'period': period})
```
Gallaecio, line 23: :lipstick: 

```suggestion
            return DateData(date_obj=date_obj, period=period)
```

dateparser/date.py
```diff
@@ -254,6 +252,40 @@ def _is_valid_date_obj(self, date_obj):
         return True
 
 
+class DateData:
+    """
+    Class that represents the parsed data with useful information.
+    It can be accessed like a dict object.
+    """
+
+    def __init__(self, date_obj=None, period=None, locale=None, is_relative=None):
```
Gallaecio, line 261: I think it makes sense to force keyword args for this class.

```suggestion
    def __init__(self, *, date_obj=None, period=None, locale=None, is_relative=None):
```

dateparser/date.py
```diff
@@ -254,6 +252,40 @@ def _is_valid_date_obj(self, date_obj):
...
+            raise KeyError(key)
+        setattr(self, key, value)
+
+    def __repr__(self):
+        return str(self.__dict__)
+
+    def __contains__(self, item):
+        if hasattr(self, item) and getattr(self, item):
+            return True
+        return False
```
Gallaecio, line 283: :lipstick: 

```suggestion
        return hasattr(self, item) and getattr(self, item)
```

Or, even shorter:

```suggestion
        return bool(getattr(self, item, False))
```

dateparser/date.py
```diff
@@ -254,6 +252,40 @@ def _is_valid_date_obj(self, date_obj):
         return True
 
 
+class DateData:
+    """
+    Class that represents the parsed data with useful information.
+    It can be accessed like a dict object.
```
Gallaecio, line 258: I assume this is so that existing code won’t break too much.

If we want to go this route, I think we should consider implementing the whole `MutableMapping` interface, otherwise some dict usages will fail.

However, I would also consider the possibility of _not_ providing a dict-like API, or at least only providing it for setting and getting values, which _may_ help avoid situations were we cannot deprecate fields in a clean way. For example, if user code iterates keys, for example, as we add and remove fields in the future warning about deprecation and keeping a backward-compatible behavior can get messy.

Not a strong opinion, though; just a thought.

dateparser/freshness_date_parser.py
```diff
@@ -144,8 +144,10 @@ def get_kwargs(self, date_string):
         return kwargs
 
     def get_date_data(self, date_string, settings=None):
+        from dateparser.date import DateData
```
Gallaecio, line 147: Why is this here, to avoid a circular dependency?

dateparser/date.py
Gallaecio, line 261: I don’t see `is_relative` used anywhere. Maybe we should delay its inclusion into the class until we have a use for it.

noviluni: I don't understand this. Could you elaborate on it a little? Why should we allow other fields than the ones defined? for avoiding issues with future deprecations? :thinking:

dateparser/freshness_date_parser.py
noviluni: Yes. Maybe we can search for another workaround.

dateparser/date.py
noviluni: At least it's used by `search_dates`, but I didn't deep too much.

dateparser/date.py
noviluni: I've been playing around. It would be really easy to support only setting and getting like a `dict` (I can push the code now), but my intention was to leave at least the __repr__() like a `dict` object (this makes it easier to see the content of the object and would require less effort changing the docs). However, adding the __repr__ like a `dict` could easily make people think that it can be used like a dict: iterating, etc.

What do you think? @Gallaecio

dateparser/date.py
Gallaecio: > Why should we allow other fields than the ones defined?

That’s not what the `*` does, it’s not the same as `*<variable name>` (e.g. `*args`). A single `*` as a parameter simply causes all parameters after it to require to be passed by name. See https://python-3-for-scientists.readthedocs.io/en/latest/python3_advanced.html#keyword-only-arguments

So, before this change, you can create an instance of `DateData` class 2 different ways:

```python
DateData(date_object, 'day', 'gl_ES', False)
DateData(date_object=date_object, period='day', locale='gl_ES', is_relative=True)
```

With my suggestion, only the latter works.

With classes and functions where positional parameter are not obvious, keyword-only arguments are a way to force calling code to remain readable.

I don’t have a strong opinion on this, though. Feel free to ignore and mark as resolved.

dateparser/freshness_date_parser.py
Gallaecio: Well, even though I do prefer to resolve it a different way, I think that can be done on a separate pull request later.

dateparser/date.py
noviluni: Hi @Gallaecio, I changed the __repr__() method and removed some methods, could you review it again?
thanks!

Gallaecio (approved):

noviluni:
I removed the `is_relative` property as it wasn't necessary. It was only used in the search part and it could be avoided.

I changed that part and moved the undocumented `ExactLanguageSearch` to the private scope (`_ExactLanguageSearch`), as we will probably need to change some behaviors in the future and this would avoid deprecating it in the future. Of course, this breaking change will be noted in the release notes.

noviluni:
I improved the coverage @Gallaecio, would you mind to take a look again? :slightly_smiling_face:

dateparser/date.py
noviluni: You were right, I could remove that field after refactoring a part of the `search_dates` part

Issue #723, opened 2020-06-25
Improve the README


noviluni:
This library doesn't have a website (apart from the docs), so the README should be treated as a presentation of the library.

I think we could add a fresher vision over this library and focus it over the use cases instead of including parts from the docs.

- [x] Add new logo ~(blocked by https://github.com/scrapinghub/dateparser/pull/719)~
- [x] Installation
- [x] Supported Python versions
- [x] Basic examples (current content with some simplifications) with basic settings.
- [x] Common use cases (more on this below)
- [ ] Multiple links to docs (how to contribute, etc).
- [x] Recommendation: to use `languages` or `locales` when knowing the language to improve performance and unexpected behaviors.
- [x] Explaining that what you have to pass to the `parse()` method is a date string, not a string containing a date.
- [ ] Explain that `search_dates` is not fully supported and it has a lot of limitations.

If you have any idea don't hesitate to add a comment!

-----
#### When consuming data from different sources:
* Scraping: extract dates from different places with several different formats and languages
* IoT: consuming data coming from different sources with different date formats
* Tooling: consuming dates from different logs / sources
* Format transformations: when transforming dates coming from files (PDF, CSV, etc.) to other formats (database, etc).

#### When offering natural interaction with users
* Tooling and CLI: allow users to write “3 days ago” to retrieve information.
* Search engine: allow people to search by date in an easiest / natural format.
* Bots: allow users to interact with a bot easily

noviluni:
I started working on this.

My idea is to create a new README.md file (which is better supported by Github) and move the current content or part of the content to the docs. You can see a preview of the version I'm working here: https://github.com/noviluni/dateparser/tree/improve_readme

@Gallaecio any feedback will be welcomed :)

noviluni:
Added new PR! #779

noviluni:
As the PR #779 has been merged I will close this issue. 

The missing sections will be omitted from the README as they are in the docs (and the `search_dates` has its own ticket: https://github.com/scrapinghub/dateparser/issues/737).

PR #779, opened 2020-09-08
Improve readme and docs

noviluni:
for: https://github.com/scrapinghub/dateparser/issues/723

### What does this include?
1. New `README.md` (moved from rst to markdown because the support is better in Github). I wrote it from scratch. Ideas are welcomed.
2. Add the "read the docs" theme to the docs to get a result more similar to the real docs when building them locally.
3. Update some configurations in the docs, including the new logo.
4. Add some improvements to the docs: 
    * move old "README.rst" as "introduction"
    * separate the "Supported locales" in a new section
    * improved the TOC tree
    * etc.
5. Update the `setup.py` file and update the `.travis.yml` to use Python 3 to deploy the package.
6. Removed the "Gitter" chat badge (is not maintained) and added the downloads badge.

To preview the `README.md` visit:  https://github.com/scrapinghub/dateparser/blob/a542f25e51b6851d8bfd84e82b4c21c53457a716/README.md

To preview the docs: clone the project, run `tox -e docs` and check the `index.htm` file in `../.tox/docs/tmp/html`.

Note that the settings need an extra cleanup, but I tried to avoid making too many changes in this PR. It should be fixed later.

On the other side, I don't know how to preview the pypi description.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ matrix:
     - python: 3.7
       env: TOXENV=py37
     - python: 3.8
-      env: TOXENV=py38
+      env: TOXENV=py38 PYPI_RELEASE_JOB=true
     - python: 3.9-dev
       env: TOXENV=py39
 
@@ -41,4 +41,4 @@ deploy:
   on:
     tags: true
     repo: scrapinghub/dateparser
-    condition: "$TOXENV == py27"
+    condition: "$PYPI_RELEASE_JOB == true
--- a/README.rst
+++ b/README.rst
@@ -1,486 +1,207 @@
-====================================================
-dateparser -- python parser for human readable dates
-====================================================
+.. raw:: html
 
-.. image:: https://img.shields.io/travis/scrapinghub/dateparser/master.svg?style=flat-square
-   :target: https://travis-ci.org/scrapinghub/dateparser
-   :alt: travis build status
+    <h1 align="center">
+      <br/>
+      <a href="https://github.com/scrapinghub/dateparser">
+          <img src="https://raw.githubusercontent.com/scrapinghub/dateparser/master/artwork/dateparser-logo.png" alt="Dateparser" width="400">
+      </a>
+      <br/>
+    </h1>
 
-.. image:: https://img.shields.io/pypi/v/dateparser.svg?style=flat-square
-   :target: https://pypi.python.org/pypi/dateparser
-   :alt: pypi version
+    <h4 align="center">Python parser for human readable dates</h4>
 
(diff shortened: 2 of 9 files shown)


Gallaecio:
> 1. New `README.md` (moved from rst to markdown because the support is better in Github)

Any example of a reStructuredText syntax that is not properly rendered in https://github.com/scrapinghub/dateparser _and_ has a Markdown equivalent?

I tend to prefer reStructuredText because it allows things that Markdown doesn’t, such as those things you are now implementing in  HTML within Markdown. I’m OK with switching to Markdown if Markdown is enough, but if the price is having HTML within Markdown, maybe it’s better to stay with reStructuredText.

noviluni:
Hi @Gallaecio, I reverted it to use the `rst` format. Could you check it again?

I added the same HTML raw code, as I wasn't able to get the same result without it, any help on this will be appreciated.

On the other sider, opinions about the content, examples, etc. would be also appreciated. 

Link to preview the current `README.rst` version: https://github.com/scrapinghub/dateparser/blob/08905be57af1a55e03f27098ea681d29b8cc98c8/README.rst

Thanks in advance! :smile:

noviluni:
@Gallaecio  I have just pushed a commit removing the HTML, but I can't get the expected result (text, image, etc. are not centered).

You can see here the version after removing the HTML code: https://github.com/scrapinghub/dateparser/blob/107e1b5026d1bd1a17694bfe237aa7b43b2e5b1b/README.rst


the version before (from my last comment): https://github.com/scrapinghub/dateparser/blob/08905be57af1a55e03f27098ea681d29b8cc98c8/README.rst


As you can see the HTML version has a better look and feel. Do you know how to fix it?


I personally think that `rst` is awesome for building documentation, but it's also harder to edit (because it relies on spaces, etc.) and it's harder to learn. The old README was really bad but there weren't too many PRs to improve it and I think it's because of the format. It's ok to me to keep the `rst` format, but I want to get it to look like the HTML version. If not I think I prefer to keep at least the raw HTML.

noviluni:
(FYI: I rechecked it and there is a bug in Github (that's the reason why I mentioned that the Markdown is better supported): https://github.com/github/markup/issues/163). We should probably keep the HTML markup for the logo and the first section).

Gallaecio:
Well, if it’s only the alignment I would seriously consider just letting the images align left. But I’m OK with adding the raw HTML for that if you prefer; just add a comment (line starting with `..`) linking to https://github.com/github/markup/issues/163

Gallaecio (approved):

Issue #549, opened 2019-08-05
Parsing in YMD format


4n4nd:
dateparser.__version__ = '0.7.1'

Problematic code sequence:
```
import dateparser

print(dateparser.__version__)
print("my date = ", dateparser.parse("2019-08-02 11:51:09.138115"))

# This statement does nothing
dateparser.parse("2d")

print("my date = ", dateparser.parse("2019-08-02 11:51:09.138115"))
```
Code output:
```
0.7.1
my date =  2019-08-02 11:51:09.138115
my date =  2019-02-08 11:51:09.138115
```

Problem:
As you can see in the code output, the same date string gets parsed into two different formats, the first one being the correct one. This does not seem to happen when I remove `dateparser.parse("2d")` (line 7).

Current workaround:
To use `settings={"DATE_ORDER": "YMD"}` when parsing any date to force this format.

rennerocha:
The first call uses `en` as locale (as none is specified). The second call (`dateparser.parse("2d")`) uses`es` as locale.

As `try_previous_locales=True`, in `DateDataParser` instance, and `es` is in `previous_locales` list (because of the second call), when we do the first call again, the string will be converted considering `es` as locale.

Lines of interest:
https://github.com/scrapinghub/dateparser/blob/master/dateparser/date.py#L303
https://github.com/scrapinghub/dateparser/blob/master/dateparser/date.py#L446

One option to avoid this side-effect would be set default as `False` for `DateDataParser.try_previous_locales`. This behaviour to try previous locales seems a bit magic to me, so IMO it would be better to explicitly enable if if we want and make clear in the docs that after that enabled, the results may vary between executions.

4n4nd:
Thanks

epicfaace:
@rennerocha would you consider making `try_previous_locales` set to `False` by default in a new release (as a breaking change)? It's not clear that dateparser is nondeterministic and caused a quite difficult bug for me to trace, and it would make more sense for it to be deterministic by default (or clearly advertised as nondeterministic in the README).

noviluni:
Hi @epicfaace 
It's not a problem to add a breaking change as we are working in a new, major, version. However, I think that this behavior (trying with previous locales) it's really important.

Imagine you are parsing a text that it's in Macedonian.

[...]

Issue #780, opened 2020-09-11
Add a "performance" section in the docs


noviluni:

noviluni:
It will be better to start this after merging this: https://github.com/scrapinghub/dateparser/pull/779 to avoid conflicts

Issue #776, opened 2020-09-04
Different parsing results depending on previous calls


tehaef:
Hello,

i hope this issue has not been reported yet (couldn't find anything). Basically i ran into a problem where previous calls to dateparser.parse seem to influence the result of later ones. For example:

```python
In [1]: import dateparser

In [2]: dateparser.parse('2020-09-04T15:47:39Z')
Out[2]: datetime.datetime(2020, 9, 4, 15, 47, 39, tzinfo=<StaticTzInfo 'Z'>)

In [3]: dateparser.parse('Di, Aug 11 2020 15:38:22 +0200')
Out[3]: datetime.datetime(2020, 8, 11, 15, 38, 22, tzinfo=<StaticTzInfo 'UTC\+02:00'>)

In [4]: dateparser.parse('2020-09-04T15:47:39Z')
Out[4]: datetime.datetime(2020, 4, 9, 15, 47, 39, tzinfo=<StaticTzInfo 'Z'>)
```
(For convenice the demonstration is in one session. It happens also when [2] and [3] + [4] are run separately).

As you can see, in the second run month and the day are swapped. The german date in [3] follows the order day month and apparently this is carried over to the next call to dateparser.parse

Any ideas on how to fix this are most welcome.

Thank you.

tehaef:
One addendum:
The behaviour demonstrated above seems to be somewhat random. Occasionally it actually does *not* show up. E.g.:
```python
>>> import dateparser
>>> dateparser.parse('2020-09-04T15:47:39Z')
datetime.datetime(2020, 9, 4, 15, 47, 39, tzinfo=<StaticTzInfo 'Z'>)
>>> dateparser.parse('Di, Aug 11 2020 15:38:22 +0200')
datetime.datetime(2020, 8, 11, 15, 38, 22, tzinfo=<StaticTzInfo 'UTC\+02:00'>)
>>> dateparser.parse('2020-09-04T15:47:39Z')
datetime.datetime(2020, 9, 4, 15, 47, 39, tzinfo=<StaticTzInfo 'Z'>)

(separate run)

>>> import dateparser
>>> dateparser.parse('2020-09-04T15:47:39Z')
datetime.datetime(2020, 9, 4, 15, 47, 39, tzinfo=<StaticTzInfo 'Z'>)
>>> dateparser.parse('Di, Aug 11 2020 15:38:22 +0200')
[...]

PR #781, opened 2020-09-11
don't try previous locales by default

noviluni:
fixes: https://github.com/scrapinghub/dateparser/issues/549, https://github.com/scrapinghub/dateparser/issues/776

Trying previous locales by default improves the performance when trying to parse the less common languages. However, it makes the `dateparser.parser()` nondeterministic, and it produces some side effects that are not easy to detect and track (check referenced issues to see examples).

We have two options:

1. Change the default value in `DateDataParser` for `try_previous_locales` from `True` to `False`.
2. Add `try_previous_locales=False` in the DateDataParser objects created in the `dateparser/__init__.py`.

(Note that it doesn't make sense to move this feature to a new setting because `dateparser.parser()` won't be able to track the used languages as it doesn't have a real internal state and if we pass different parameters we will lose them.)

The second approach has the advantage that it remains `True` by default when using `DateDataParser` directly, but I think it's better to make both APIs (`dateparser.parse()` and `DateDataParser.get_date_data()`) to behave in the same way by default, so I decided to use the first approach.

I added this new issue: https://github.com/scrapinghub/dateparser/issues/780 to create a section in the documentation regarding the performance tweaks.

### DIFF
--- a/dateparser/date.py
+++ b/dateparser/date.py
@@ -298,16 +298,16 @@ class DateDataParser:
     locale_loader = None
 
     @apply_settings
-    def __init__(self, languages=None, locales=None, region=None, try_previous_locales=True,
+    def __init__(self, languages=None, locales=None, region=None, try_previous_locales=False,
                  use_given_order=False, settings=None):
 
-        if not isinstance(languages, (list, tuple, Set)) and languages is not None:
+        if languages is not None and not isinstance(languages, (list, tuple, Set)):
             raise TypeError("languages argument must be a list (%r given)" % type(languages))
 
-        if not isinstance(locales, (list, tuple, Set)) and locales is not None:
+        if locales is not None and not isinstance(locales, (list, tuple, Set)):
             raise TypeError("locales argument must be a list (%r given)" % type(locales))
 
-        if not isinstance(region, str) and region is not None:
+        if region is not None and not isinstance(region, str):
             raise TypeError("region argument must be str or unicode (%r given)" % type(region))
 
         if not isinstance(try_previous_locales, bool):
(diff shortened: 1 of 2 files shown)


dateparser/date.py
```diff
@@ -298,16 +298,16 @@ class DateDataParser:
...
+        if languages is not None and not isinstance(languages, (list, tuple, Set)):
             raise TypeError("languages argument must be a list (%r given)" % type(languages))
 
-        if not isinstance(locales, (list, tuple, Set)) and locales is not None:
+        if locales is not None and not isinstance(locales, (list, tuple, Set)):
             raise TypeError("locales argument must be a list (%r given)" % type(locales))
 
-        if not isinstance(region, str) and region is not None:
+        if region is not None and not isinstance(region, str):
             raise TypeError("region argument must be str or unicode (%r given)" % type(region))
```
noviluni, line 312: I inverted some statements in the `if `s to improve the performance, as checking if something is `None` is cheaper than using `isinstance` and, as the default values are `None`, from now most of the times we won't need to check the `isinstance` condition.

tests/test_date.py
```diff
@@ -395,6 +395,20 @@ def test_returned_detected_locale_should_be(self, date_string, locale):
...
+        self.when_date_string_is_parsed('2020-05-01')
+        self.then_detected_locale('en')
+
+    def test_try_previous_locales_true(self):
+        self.given_parser(try_previous_locales=True)
+        self.when_date_string_is_parsed('Mañana')  # es
+        self.then_detected_locale('es')
+        self.when_date_string_is_parsed('2020-05-01')
+        self.then_detected_locale('es')
+
```
noviluni, line 411: I also added some tests as this part was not covered by the tests.

Gallaecio (approved):

PR #782, opened 2020-09-11
remove numeral translation_data

noviluni:
The `numeral_translation_data` has been never used. Keeping these files makes the code maintainability harder and it's decreasing the coverage.

I think that these files were created and kept with the idea of being able to parse natural written numbers (for example: "twenty-five" for 25), however, this part will be supplied by a separate library called `number-parser`: https://github.com/scrapinghub/number-parser. PR: https://github.com/scrapinghub/dateparser/pull/711


noviluni:
Notice that the total coverage is increased by a tremendous `+2.76%`. :muscle:

Gallaecio (approved):
Nice, 100% is getting closed :tada:

PR #789, opened 2020-09-17
Locale order support

### LINKED ISSUE #770: Wrong prioritization of languages
I think there is something wrong in dateparser prioritization of languages, as introducing 'en' even in the last position hurts extraction of dates that were extracted properly when English was not there.

```py
import dateparser
dateparser.parse("11/12", languages=['en'])
Out[3]: datetime.datetime(2020, 11, 12, 0, 0)
```
This is right
```py
dateparser.parse("11/12", languages=['es'])
Out[4]: datetime.datetime(2020, 12, 11, 0, 0)
```
This is also right, because the standard in Spain is DD/MM
But now if we add English to the languages list in the last position...
```py
dateparser.parse("11/12", languages=['es', 'en'])
Out[5]: datetime.datetime(2020, 11, 12, 0, 0)
```
We got it parsed like in English, even if Spanish is first in the list of languages. This is unexpected to me, I would have expected prioritizing Spanish instead.

dariuschira:
closes #770

### DIFF
--- a/dateparser/date.py
+++ b/dateparser/date.py
@@ -411,7 +411,9 @@ class DateDataParser:
 
     :param use_given_order:
         If True, locales are tried for translation of date string
-        in the order in which they are given.
+        in the order in which they are given. This is equivalent to the
+        ``USE_GIVEN_LANGUAGE_ORDER`` setting; the given order is preserved if
+        either is enabled.
     :type use_given_order: bool
 
     :param settings:
@@ -556,6 +558,15 @@ class DateDataParser:
         return date_tuple(**date_data.__dict__)
 
     def _get_applicable_locales(self, date_string):
+        # The given order is preserved if requested either through the
+        # ``use_given_order`` constructor argument or the
+        # ``USE_GIVEN_LANGUAGE_ORDER`` setting (which also reaches the
+        # top-level :func:`dateparser.parse`). When no languages or locales are
+        # given there is nothing to order, so the default order is used.
+        use_given_order = (
+            self.use_given_order or self._settings.USE_GIVEN_LANGUAGE_ORDER
+        )
+
         pop_tz_cache = []
 
         def date_strings():
@@ -593,7 +604,7 @@ class DateDataParser:
             languages=self.languages,
             locales=self.locales,
             region=self.region,
-            use_given_order=self.use_given_order,
+            use_given_order=use_given_order,
         ):
             for s in date_strings():
                 if self._is_applicable_locale(locale, s):
@@ -604,7 +615,7 @@ class DateDataParser:
                 languages=self._settings.DEFAULT_LANGUAGES,
                 locales=None,
                 region=self.region,
-                use_given_order=self.use_given_order,
+                use_given_order=use_given_order,
             ):
                 yield locale
 
--- a/dateparser_data/settings.py
+++ b/dateparser_data/settings.py
@@ -24,6 +24,7 @@ settings = {
     "SKIP_TOKENS": ["t"],
     "NORMALIZE": True,
     "DEFAULT_LANGUAGES": [],
+    "USE_GIVEN_LANGUAGE_ORDER": False,
     # Optional language detection
     "LANGUAGE_DETECTION_CONFIDENCE_THRESHOLD": 0.5,
     # Time span settings
--- a/dateparser/conf.py
+++ b/dateparser/conf.py
@@ -32,6 +32,7 @@ class Settings:
     * `DEFAULT_DAYS_IN_MONTH`
     * `PARSERS`
     * `DEFAULT_LANGUAGES`
+    * `USE_GIVEN_LANGUAGE_ORDER`
     * `LANGUAGE_DETECTION_CONFIDENCE_THRESHOLD`
     * `CACHE_SIZE_LIMIT`
     """
@@ -227,6 +228,7 @@ def check_settings(settings):
         "FUZZY": {"type": bool},
         "PREFER_LOCALE_DATE_ORDER": {"type": bool},
         "DEFAULT_LANGUAGES": {"type": list, "extra_check": _check_default_languages},
+        "USE_GIVEN_LANGUAGE_ORDER": {"type": bool},
         "LANGUAGE_DETECTION_CONFIDENCE_THRESHOLD": {
             "type": float,
             "extra_check": _check_between_0_and_1,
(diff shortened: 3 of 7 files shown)


docs/usage.rst
```diff
@@ -194,6 +194,15 @@ For example, assuming current date is `June 16, 2019`:
     >>> parse('March 12, 2012', settings={'REQUIRE_PARTS': ['day', 'month', 'year']})
     datetime.datetime(2012, 3, 12, 0, 0)
 
+
+``USE_GIVEN_LANGUAGE_ORDER`` Choose whether to preserve or not the order given in ``languages`` or ``locales``. It is ``False`` by default so dateparser will choose the most common language from the given list for efficiency.
+
+    >>> dateparser.parse("11/12", locales=['es', 'en'])
+    datetime.datetime(2020, 11, 12, 0, 0)
+    >>> dateparser.parse("11/12", locales=['es', 'en'], settings={'USE_GIVEN_LANGUAGE_ORDER': True})
```
ivanprado, line 202: Can you add a case where locales is empty? IMO it shouldn't fail.

dateparser/date.py
```diff
@@ -314,16 +309,11 @@ def __init__(self, languages=None, locales=None, region=None, try_previous_local
...
                             % type(try_previous_locales))
 
-        if not isinstance(use_given_order, bool):
-            raise TypeError("use_given_order argument must be a boolean (%r given)"
-                            % type(use_given_order))
-
-        if not locales and use_given_order:
-            raise ValueError("locales must be given if use_given_order is True")
+        if not (locales or languages) and settings.USE_GIVEN_LANGUAGE_ORDER:
+            raise ValueError("locales or languages must be given if USE_GIVEN_LANGUAGE_ORDER is True")
```
ivanprado, line 313: Not sure of this failure. If the given list of languages is empty in this case then it should use the defaults. 

I'm thinking on the case where this is combined with a language detection model. For some cases, the prediction could fail and not return any language, but still, dateparser must try its best.

docs/usage.rst
mirceachira: I think you're right, I fixed that and mentioned it in the docs here

dateparser/date.py
mirceachira: I do see your point, and I agree. The question is if this is what's expected: to restrict to the given locales (or languages) or to use those as a starting point but try the other locales anyways.
If the second is true, then removing this check should be the right choice.

ivanprado: I can talk about the use case I have in mind: using a page language detector to provide hints to have better date parsing. In this case, what you have is just `hints`:

- In some case, you don't get any language from the language detector. Still, you want dateparser to try its best. 
- In some cases, the list won't be complete. You don't want to remove locales just because the language detector didn't detect them. 

So from my point of view, this list should be just hints: they can change the order of the default locales, but never restrict them.

noviluni (commented):
Hey @mirceachira! Good job!

It seems that there are some errors in the flake8 pipeline: https://travis-ci.org/github/scrapinghub/dateparser/jobs/728909794, could you fix them?

dateparser/date.py
noviluni: Hi @ivanprado, the `languages` and `locales` parameters have been used historically for restricting the used languages. This is sometimes really important and helps to improve the performance, so I think we can't change them to work as you mentioned.

However, what you say sounds like adding something new like a `exclude_languages` parameter, so you could do something like this:

```python3
date = dateparser.parse("<date string>", languages=['ru', 'tr', 'es'])
if not date:
    date = dateparser.parse("<date string>", exclude_languages=['ru', 'tr', 'es'])
```

What do you think? Could this work for you?

Gallaecio: Also, on the argument of not raising an exception because the calling code may provide an empty list, the calling code should also be able not to set that setting when that happens. I fail to see the point of removing the exception.

dateparser_data/settings.py
```diff
@@ -21,4 +21,5 @@
     'RETURN_TIME_AS_PERIOD': False,
     'PARSERS': default_parsers,
     'REQUIRE_PARTS': [],
+    'USE_GIVEN_LANGUAGE_ORDER': False
```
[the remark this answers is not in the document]
mirceachira: I think `False` is right as a default because that way you'd get the added benefit of trying your most likely language first. If you give a list of locales I'd assume you want the date to be extracted correctly *regardless* of which of your locales is the right one. Also, I would argue that expecting it to be as efficient as possible by default is better if you allow the customization.

dateparser/date.py
ivanprado: @noviluni I see your point. It was not clear to me that `locales` and `languages` were used as contraints to improve performance. The documentation (https://github.com/scrapinghub/dateparser/blob/master/dateparser/__init__.py#L23-L31) is ambiguous, and my first thought was that they would be used as "hints" . I think it would be great to update the documentation to make it clear that they restrict the languages used. 

Given that `languages` and `locales` are for restricting I agree with you than then maybe is not the best place for the use case I'm proposing. What I'm proposing is to be able to propose "hints" for the locales/languages. I don't think `exclude_languages` should be way. 

What do you think about a new parameter `language_hints`? It would be used just to alter the languages/locales priorities.

dateparser/date.py
```diff
@@ -314,16 +309,8 @@ def __init__(self, languages=None, locales=None, region=None, try_previous_local
             raise TypeError("try_previous_locales argument must be a boolean (%r given)"
                             % type(try_previous_locales))
 
-        if not isinstance(use_given_order, bool):
-            raise TypeError("use_given_order argument must be a boolean (%r given)"
-                            % type(use_given_order))
-
-        if not locales and use_given_order:
-            raise ValueError("locales must be given if use_given_order is True")
-
```
Gallaecio, line 323: Based on the conversation so far, [my vote is to keep these](https://github.com/scrapinghub/dateparser/pull/789#discussion_r492014507).

Gallaecio (changes requested):

dateparser/date.py
mirceachira: Hey @Gallaecio if this is the way to go the I believe https://github.com/scrapinghub/dateparser/pull/789/commits/94c96fd7585a2b00bbe1ef3670373f673eb82ec7 should cover it.

Let me know what you think

AdrianAtZyte (approved):

Issue #791, opened 2020-09-23
Not compatible with tzlocal 3.0b1


washeck:
`tzlocal 3` switched to from `pytz` to `zoneinfo` from Python 3.9 (or `backports.tzinfo` for older Pythons) and `dateparser` now crashes with

```
.../lib/python3.8/site-packages/dateparser/date_parser.py", line 37, in parse
    date_obj = stz.localize(date_obj)
AttributeError: 'backports.zoneinfo.ZoneInfo' object has no attribute 'localize'

```

dateparse should either restrict the required `tzlocal` version to <3 or fix the compatibility `zoneinfo.ZoneInfo` objects.

noviluni:
Hey @washeck, thank you for raising this up!

I added a PR adding compatibility with both versions. #792

spovich:
I just ran into this with a python 3.8 project using dateparser 0.7.6.  By adding `tzlocal==2.1` as an explicit dependency, I was able to work around the error.  Look forward #792 landing and release! Thanks!

PR #792, opened 2020-09-24
fix compatibility with tzlocal >= 3

noviluni:
fixes: https://github.com/scrapinghub/dateparser/issues/791

As explained in the issue, tzlocal switched from `pytz` to `tzinfo` and now some parsers crash.

I run the test suite with these changes locally with `tzlocal==2.1` and `tzlocal==3.0b1` and both versions worked as expected.

Note that I used `hasattr(tz, 'localize'):` because we can't use directly the `replace` with pytz because it doesn't work for some timezones:

> This library only supports two ways of building a localized time. The first is to use the localize() method provided by the pytz library [...] Unfortunately **using the tzinfo argument of the standard datetime constructors ‘’does not work’’ with pytz for many timezones.**

(and when I tried I got some errors)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,4 +41,4 @@ deploy:
   on:
     tags: true
     repo: scrapinghub/dateparser
-    condition: "$PYPI_RELEASE_JOB == true
+    condition: "$PYPI_RELEASE_JOB == true"
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ matrix:
     - python: 3.7
       env: TOXENV=py37
     - python: 3.8
-      env: TOXENV=py38
+      env: TOXENV=py38 PYPI_RELEASE_JOB=true
     - python: 3.9-dev
       env: TOXENV=py39
 
@@ -41,4 +41,4 @@ deploy:
   on:
     tags: true
     repo: scrapinghub/dateparser
-    condition: "$TOXENV == py27"
+    condition: "$PYPI_RELEASE_JOB == true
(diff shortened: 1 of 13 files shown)


Gallaecio (commented):
Looks good to me.

However, I don’t see coverage and test results in this PR (maybe we broke something?). I was looking forward to the coverage data, as I expect we may need additional jobs to cover all scenarios here.

Gallaecio (approved):

Gallaecio (approved):
While I think a job that covers the other scenario would be nice to have, I think it’s OK to merge this _as is_.

Gallaecio:
Oh, I’ve just realized you’ve gone ahead with the adding of a job.

Gallaecio:
I’ve not tested it, but [this](https://github.com/noviluni/dateparser/pull/1) might work.

I think it makes sense to modify the existing Python 3.9 job to also use the latest `tzlocal`.

Although the opposite may also be a good choice, possibly better: make the job with the lowest Python version also install pinned versions of dependencies, with the lowest supported versions of dependencies, and let all other jobs use the latest versions of everything.

Gallaecio:
The coverage data seems still incomplete. But some of the if-else statements do have 100% coverage now, so I’m a bit confused.

noviluni:
> The coverage data seems still incomplete. But some of the if-else statements do have 100% coverage now, so I’m a bit confused.

I'm also confused. However, without checking the coverage, I think that we only added tests for **tzlocal >= 3 with Python 3.9** and the older python versions (3.5, 3.6, 3.7 and **3.8) with tzlocal < 3**, but I think we missed the case Python <3.9 with tzlocal >= 3. We could add an extra **Python 3.8 with tzlocal >= 3**, what do you think?

Gallaecio:
I had completely forgotten the Python version also made a difference, now I get why you had added 2 additional jobs :facepalm:

PR #794, opened 2020-09-24
remove FUZZY setting

noviluni:
As mentioned here: https://github.com/scrapinghub/dateparser/pull/722

I removed the `FUZZY` setting as:
* It has been never documented
* There isn't any known use case where a user needs to change it.
* There are 0 references in issues

### DIFF
--- a/dateparser/parser.py
+++ b/dateparser/parser.py
@@ -293,7 +293,7 @@ class _parser:
                     self.time = lambda: time_parser(self._token_time)
                     continue
 
-            results = self._parse(type, token, settings.FUZZY, skip_component=skip_component)
+            results = self._parse(type, token, skip_component=skip_component)
             for res in results:
                 if len(token) == 4 and res[0] == 'year':
                     skip_component = 'year'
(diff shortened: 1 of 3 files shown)


dateparser/parser.py
```diff
@@ -395,15 +395,6 @@ def _results(self):
...
 
         time = self.time() if self.time is not None else None
-
-        if self.settings.FUZZY:
-            attr_truth_values = []
-            for attr in ['day', 'month', 'year', 'time']:
-                attr_truth_values.append(getattr(self, attr, False))
-
-            if not any(attr_truth_values):
-                raise ValueError('Nothing date like found')
```
Gallaecio, line 405: It looks like the purpose of fuzzy is to only return a date if all those elements can be extracted for that date. Is that not a scenario that makes sense? Or is it that it does not work? (i.e. does the test also pass if it is set to `False`?)

noviluni: Hi @Gallaecio, there are several reasons to remove this setting:

1. Is not documented.
2. Only applies to `absolute-parser`.
3. Has a confusing name ("when Fuzzy is True then raise an error if the date is not complete" does this make sense for you? The semanthics are really bad.
4. We have the `STRICT_PARSING` setting which is documented and works better as well as `REQUIRE_PARTS`, that can be used with the same purpose.

Gallaecio (approved):

PR #795, opened 2020-09-25
fix .travis.yml

noviluni:


Gallaecio (approved):

setup.py
```diff
@@ -1,7 +1,7 @@
 import re
 from setuptools import setup, find_packages
 
-__version__ = re.match(r"__version__.*\s*=\s*[']([^']+)[']", open('dateparser/__init__.py').read())[1]
+__version__ = re.findall(r"__version__.*\s*=\s*[']([^']+)[']", open('dateparser/__init__.py').read())[0]
```
Gallaecio, line 4: Untested, but this might also work and be closer to the original code:

```suggestion
__version__ = re.search(r"__version__.*\s*=\s*[']([^']+)[']", open('dateparser/__init__.py').read())[1]
```

`re.match` is basically `re.search` with an implicit `^` in the pattern.

Gallaecio: Nope. The problem is indeed the syntax. Today I learn that is specific to 3.6+.

What about using the old syntax equivalent, `.group(1)`?

Issue #796, opened 2020-09-25
Improve `settings`


noviluni:
There are some bugs related to the `settings` that are hard to debug and some `settings` that aren't documented, so I think we should address this.

This is my plan:
- [ ] Validate settings and raise an error if they are not valid (breaking change)
- [ ] Possibly remove some settings? (breaking change)
- [ ] Document missed settings and improve others.
- [ ] Check that all the settings are applying for each parser.

mnikolop:
I have been trying to use the settings with this 
` from dateparser.settings import default_parsers`   
`parsers = [parser for parser in default_parsers if parser != 'relative-time']`   
`parse('today', settings={'PARSERS': parsers})`

and it throws with `ModuleNotFoundError: No module named 'dateparser.settings'`. 
Is there something I am missing? I have installed dateparser with pip as admin and have access to everything as admin as well.
Thanks for any help.

noviluni:
Hi @mnikolop , there is an error in the docs that will be fixed in the next version. You can import `default_parsers` from `dateparser_data.settings`:


```pythin3
from dateparser_data.settings import default_parsers
```

DeepSpace2:
Maybe the settings should be an enum? this way they can be auto-completed (and auto documented, by using tools such as `sphinx.autodoc`) so there will be no need to "guess" what settings are available.

PR #797, opened 2020-09-25
add settings validation

noviluni:
Adding settings validation in one single place. In that way, we can raise an error if there is misconfigurations and we ensure that the submitted settings are valid.

for: https://github.com/scrapinghub/dateparser/issues/796

~It's not fully finished, as we need a way to get a list of the existing parsers.~

I would prefer to get the "existing parsers" from another place (https://github.com/scrapinghub/dateparser/issues/798), but I think this can be refactored in the future.


dateparser/conf.py
```diff
@@ -79,8 +81,135 @@ def wrapper(*args, **kwargs):
...
     return wrapper
+
+
+class SettingValidationError(ValueError):
+    pass
+
+
+def _check_require_part(provided_list):
+    """Returns `True` if the provided list of parts is valid"""
+    return not set(provided_list) - {'day', 'month', 'year'}
```
noviluni, line 96: should I check the length to avoid duplicated parsers?

Gallaecio: I think that would be great.

dateparser/conf.py
```diff
@@ -79,8 +81,133 @@ def wrapper(*args, **kwargs):
...
+
+
+def _check_parsers(provided_list):
+    """Returns `True` if the provided list of parsers is valid"""
+    existing_parsers = ['timestamp', 'relative-time', 'custom-formats', 'absolute-time', 'no-spaces-time']
```
Gallaecio, line 101: I feel like `existing_parsers` should be imported from somewhere else. And I think it would be useful for users to be able to import it as well, e.g. to write something like:

```python
undesired_parsers = {'foo', 'bar'}
parsers_to_use = [parser for parser in EXISTING_PARSER if parser not in undesired_parsers]
```

dateparser/conf.py
```diff
@@ -79,8 +81,133 @@ def wrapper(*args, **kwargs):
...
+            'type': str,
+        },
+        'TO_TIMEZONE': {
+            # It defaults to None, but it's not allowed to use it directly
+            'values': (),
+            'type': str
+        },
+        'RETURN_AS_TIMEZONE_AWARE': {
+            # It defaults to 'default', but it's not allowed to use it directly
+            'values': (True, False),
```
Gallaecio, line 127: I think we can skip this for booleans.

dateparser/conf.py
```diff
@@ -79,8 +81,133 @@ def wrapper(*args, **kwargs):
...
+        # check type:
+        if not setting_type == setting_props['type']:
+            raise SettingValidationError(
+                '"{}" must be "{}", not "{}".'.format(
+                    setting_name, setting_props['type'].__name__, setting_type.__name__
+                )
+            )
+
+        # check values:
+        if setting_props['values'] and setting_value not in setting_props['values']:
```
Gallaecio, line 198: What about `setting_props.get('values', None)`, to allow to remove `values` keys above for settings that do not need value validation?

dateparser/conf.py
```diff
@@ -79,8 +81,133 @@ def wrapper(*args, **kwargs):
...
+                    setting_value,
+                    setting_name,
+                    '", "'.join(setting_props['values'][:-1]),
+                    setting_props['values'][-1],
+                )
+            )
+
+        # specific checks
+        extra_check = setting_props.get('extra_check')
+        if extra_check and not extra_check(setting_value):
```
Gallaecio, line 210: What if we let the check function raise the exception? That way, the message can be made more useful when possible.

dateparser/conf.py
noviluni: I think this too: https://github.com/scrapinghub/dateparser/issues/798

However, I don't want to leave this without merging after fixing that issue, so I prefer to include it as-is and change it once we fix #798

noviluni:
Hi @Gallaecio , I addressed all from your feedback, could you review it again?

Thanks you!

noviluni:
Note that this new behaviour will be documented in this PR: https://github.com/scrapinghub/dateparser/pull/722

Gallaecio (approved):

PR #799, opened 2020-09-28
fix flake8 W rules

noviluni:


dateparser/languages/dictionary.py
```diff
@@ -167,16 +167,12 @@ def _split_by_numerals(self, string, keep_formatting):
                 if self._should_capture(token, keep_formatting)]
 
     def _should_capture(self, token, keep_formatting):
-        return (
-            keep_formatting or
-            (token in ALWAYS_KEEP_TOKENS) or
-            KEEP_TOKEN_PATTERN.match(token)
-        )
+        return keep_formatting or (token in ALWAYS_KEEP_TOKENS) or KEEP_TOKEN_PATTERN.match(token)
```
Gallaecio, line 170: :lipstick:  Maybe remove the internal parenthesis as well?

dateparser_scripts/write_complete_data.py
```diff
@@ -98,10 +98,10 @@ def write_complete_data(in_memory=False):
         out_text = ('info = ' + translation_data + '\n').encode('utf-8')
         _write_file(date_translation_directory + language + '.py', out_text, 'wb', in_memory, in_memory_result)
 
-    init_text = '\n'.join(
-            ["from dateparser.data import date_translation_data",
-             "from .languages_info import language_order, language_locale_dict"]
-             )
+    init_text = (
+        "from dateparser.data import date_translation_data\n"
+        "from .languages_info import language_order, language_locale_dict\n"
```
Gallaecio, line 103: This is different (extra line break at the end). I’m not sure if it is intended or not, so I’m asking just in case.

noviluni: Yes, it is. It's because the generated file was missing the line break at the end of file.

Gallaecio (approved):

PR #800, opened 2020-09-28
Fix flake8 rules

noviluni:
~(This includes part of this: https://github.com/scrapinghub/dateparser/pull/799, so it should be reviewed/merged after it).~


dateparser/calendars/jalali_parser.py
```diff
@@ -147,8 +147,8 @@ def comp_key(tup):
         day_pairs[-14] = thirty
         day_pairs[1] = thirteen
 
-        for persian, number in reduce(
+        for pers, number in reduce(
```
Gallaecio, line 150: :lipstick: This looks like a change for the worse to me. Any approach that does not require shortening the variable name?

dateparser/parser.py
```diff
@@ -281,7 +281,7 @@ def __init__(self, tokens, settings):
 
                 try:
                     microsecond = MICROSECOND.search(self.filtered_tokens[index + 1][0]).group()
-                    _is_after_time_token = token.index(":")
+                    token.index(":")  # raise ValueError if ':' can't be found
                     _is_after_period = self.tokens[self.tokens.index((token, 0)) + 1][0].index('.')
```
Gallaecio, line 285: `_is_after_period` seems unused as well.

dateparser/calendars/jalali_parser.py
noviluni: Yeah, that's right. What about  `from convertdate import persian as persian_calendar` (and the `persian, number` here) or leave it the import as-is and then change to `for persian_number, number in ...` ?

What approach do you like the most?

dateparser/parser.py
noviluni: You are right, I pushed a new commit amending this.

dateparser/calendars/jalali_parser.py
noviluni: I pushed a commit with a new version, let me know what you think :+1:

Gallaecio: Perfect, now looks much better!

Gallaecio (approved):
Nice!

PR #803, opened 2020-10-01
added different names of months in Arabic language

MostafaAhmed456:
I am using dateparser in my work to parse dates extracted from web pages. we are scraping around 10K web pages every day. and using your package to parse these different date formats. we were missing some month names, we added it to ar.py and tested it


Gallaecio:
Thank you!

However, please make sure you follow https://dateparser.readthedocs.io/en/latest/contributing.html#guidelines-for-editing-translation-data

PR #807, opened 2020-10-09
Consider RETURN_TIME_AS_PERIOD for relative times

### LINKED ISSUE #806: RETURN_TIME_AS_PERIOD doesn't always work
For example I'd expect the following to have a period of time, but it's day:

```
>>> import dateparser
>>> dp = dateparser.date.DateDataParser(languages=['en'], settings={'RETURN_TIME_AS_PERIOD': True})
>>> dp.get_date_data('yesterday at 11:34')
{'date_obj': datetime.datetime(2020, 10, 7, 11, 34), 'period': 'day', 'locale': 'en'}
```

Whereas the following has a period of time as expected:

```
>>> dp.get_date_data('11:34')
{'date_obj': datetime.datetime(2020, 10, 8, 11, 34), 'period': 'time', 'locale': 'en'}
```

onlynone:
This should set the period to time when RETURN_TIME_AS_PERIOD is set, and you parse a date like 'yesterday at 1:30'. This should fix #806 .

### DIFF
--- a/dateparser/freshness_date_parser.py
+++ b/dateparser/freshness_date_parser.py
@@ -49,17 +49,14 @@ class FreshnessDateDataParser:
 
         _settings_tz = settings.TIMEZONE.lower()
 
-        def apply_time(dateobj, timeobj, period):
+        def apply_time(dateobj, timeobj):
             if not isinstance(_time, time):
-                return dateobj, period
-
-            if settings.RETURN_TIME_AS_PERIOD:
-                period = 'time'
+                return dateobj
 
             return dateobj.replace(
                 hour=timeobj.hour, minute=timeobj.minute,
                 second=timeobj.second, microsecond=timeobj.microsecond
-            ), period
+            )
 
         if settings.RELATIVE_BASE:
             self.now = settings.RELATIVE_BASE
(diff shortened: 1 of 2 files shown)


dateparser/freshness_date_parser.py
```diff
@@ -91,7 +94,7 @@ def apply_time(dateobj, timeobj):
         date, period = self._parse_date(date_string, settings.PREFER_DATES_FROM)
 
         if date:
-            date = apply_time(date, _time)
+            date, period = apply_time(date, _time, period)
```
Gallaecio, line 97: :lipstick: I personally think that checking if the return value is the same as the passed `date` object after this line, and changing `period` based on that, would look better than the proposed changes to the implementation of `apply_time`.

Gallaecio (approved):

dateparser/freshness_date_parser.py
onlynone: Made this change.

onlynone:
@Gallaecio I made the changes you requested, but I don't know why that made the test coverage go down.

noviluni (approved):
Nice!

PR #820, opened 2020-10-27
Add architecture ppc64le to travis build

ddeka2910:
Add support for architecture ppc64le.  
This is part of the Ubuntu distribution for ppc64le. This helps us simplify testing later when distributions are re-building and re-releasing. For more info tag @gerrith3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,28 @@ matrix:
       env: TOXENV=latest
     - python: 3.9-dev
       env: TOXENV=latest
+    #powerjobs
+    - python: 3.8
+      arch: ppc64le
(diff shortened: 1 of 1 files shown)


Gallaecio:
@ddeka2910 This is a pure-Python package. Wouldn’t it make more sense for this kind of change to be applied to non-pure-Python dependencies instead?

ddeka2910:
We want to test the compatability of repository for the ppc64le irrespective of language.

Gallaecio:
I don’t think that justifies duplicating our CI time.

noviluni:
We are not using Travis anymore, so I'm going to close this issue.

PR #822, opened 2020-10-28
add Python 3.10 support

noviluni:


noviluni:
It seems that Python 3.10 is not mature enough :smile: 

I will close this until the version is more stable.

PR #805, opened 2020-10-07
make use of most common content locales

mirceachira:
closes #714

PR #824, opened 2020-10-28
Locale order fix invalid languages

mirceachira:
@noviluni regarding your comment here https://github.com/scrapinghub/dateparser/pull/805#issuecomment-705368883

I created this pull-request but there are still some tests failing, I'm not sure why some dates don't get parsed when there should be language data for them and I only changed the locale order.

Is this the approach you had in mind for updating the language data?


noviluni:
Hi @mirceachira, yes, this is the approach I thought, but I didn't expect the data to change so much. Most of the tests are failing because they removed specific words for some languages and the old tests now don't pass. This could be fixed by adding the words to the `supplementary_language_data` files, checking new languages... but this is not definitely the way to go. 

We can't update the CLDR data in this way (if I'm not wrong, the data we have is from 2017), we should definitely go version by version and review every update carefully. I will open a new issue/PR with the idea I have in mind and you will be able to finish the #805 PR. In the meanwhile, you can keep this PR open if you want.

noviluni:
Hi @mirceachira, I created this PR: https://github.com/scrapinghub/dateparser/pull/825 that fixes the errors you are finding, as it sets a version for the CLDR data. I think we can close this PR in favour of the other PR.

PR #825, opened 2020-10-29
Use versioned CLDR data

noviluni:
We have been using an old version of CLDR data. It has been tried to update the data from CLDR in the past but it's something really hard to do. 

I changed the `dateparser_scripts.utils.get_raw_data` function to point to a specific version and updated the files accordingly.

That version was released in April 2017 and it is the most similar I found to the current data (I would say "exactly the same version"). Most of the changes you will see are related to the order, as it's important to keep all ordered to update the data easily in the future.

Once I fixed that I downloaded that version data (JSON files, script: `python dateparser_scripts/get_cldr_data.py`), I updated the resources (py files, script: `dateparser_scripts/write_complete_data.py`) and I run the tests and all is working as expected, so I think this is a good starting point. Once we merge this we can gradually update the CLDR versions to get the most recent version.

(you can review this easier by going commit by commit)

closes: https://github.com/scrapinghub/dateparser/pull/824


noviluni:
BTW, I removed the `orderedset` dependency for the scripts.

noviluni:
related to this: https://github.com/scrapinghub/dateparser/pull/487/files

noviluni:
I created this ticket with the step by step guide to update the CLDR in the future once we merge this: https://github.com/scrapinghub/dateparser/issues/826

Gallaecio (approved):
Awesome! :rocket:

PR #814, opened 2020-10-16
Remove self.now from freshness_date_parser in favor of a local variable (#813)

abkonsta:
In web server environment, since freshness_date_parser is a shared instance, it looks like self.now was set to None by some threads while still being used by others.

This change removes self.none from FreshnessDateDataParser and instead introduces a local variable insude FreshnessDateDataParser.parse().

I could've also just removed the statement at the end of pase() setting self.now to None, but I feel that this way is cleaner.

Thanks!

Fixes #813

Issue #813, opened 2020-10-16
Concurrency issues resulting in TypeError: unsupported operand type(s) for -: 'NoneType' and 'relativedelta'


andrek-data-axle:
I am assuming its a concurrency issue because the only way I can get it to fail is running multiple calls to dateparser.parse('14 days ago UTC') on separate threads.

```
  File "/var/app/current/credit/users/services/profile.py", line 67, in add_profile
    date_viewed__gt=dateparser.parse('14 days ago UTC')
  File "/var/app/venv/staging-LQM1lest/lib/python3.7/site-packages/dateparser/conf.py", line 85, in wrapper
    return f(*args, **kwargs)
  File "/var/app/venv/staging-LQM1lest/lib/python3.7/site-packages/dateparser/__init__.py", line 53, in parse
    data = parser.get_date_data(date_string, date_formats)
  File "/var/app/venv/staging-LQM1lest/lib/python3.7/site-packages/dateparser/date.py", line 418, in get_date_data
    locale, date_string, date_formats, settings=self._settings)
  File "/var/app/venv/staging-LQM1lest/lib/python3.7/site-packages/dateparser/date.py", line 194, in parse
    return instance._parse()
  File "/var/app/venv/staging-LQM1lest/lib/python3.7/site-packages/dateparser/date.py", line 198, in _parse
    date_obj = self._parsers[parser_name]()
  File "/var/app/venv/staging-LQM1lest/lib/python3.7/site-packages/dateparser/date.py", line 212, in _try_freshness_parser
    return freshness_date_parser.get_date_data(self._get_translated_date(), self._settings)
  File "/var/app/venv/staging-LQM1lest/lib/python3.7/site-packages/dateparser/freshness_date_parser.py", line 149, in get_date_data
    date, period = self.parse(date_string, settings)
  File "/var/app/venv/staging-LQM1lest/lib/python3.7/site-packages/dateparser/freshness_date_parser.py", line 94, in parse
    date, period = self._parse_date(date_string, settings.PREFER_DATES_FROM)
  File "/var/app/venv/staging-LQM1lest/lib/python3.7/site-packages/dateparser/freshness_date_parser.py", line 134, in _parse_date
    date = self.now - td
TypeError: unsupported operand type(s) for -: 'NoneType' and 'relativedelta'
```

abkonsta:
I monkey patched in my project and was able to stress-test it with no more concurrency issues in freshness_date_parser
Pull request submitted.

[...]

PR #834, opened 2020-11-12
Resolve concurrency issues

alex-triphonov:
I’ve found two main concurrent issues with dateparser:
### I) TypeError: unsupported operand type(s) for -: 'NoneType' and 'relativedelta' #441 #813
**_Reason:_** Variable **freshness_date_parser** is a shared instance and its **self.now** was set to **None** by some threads while still being used by others.
**_Solution:_** Make a local variable now instead of self.now.  There’s already an PR #814 with this solution, but unfortunately it stuck, so I had to add it to my code in order to pass tests.

### II) Threads mixed up date orders
**_Reason:_** Settings instances are shared between locales and while one thread in
**_DateLocaleParser._try_parser()** was parsing a date with _date order_ picked for specific locale, other threads with the same **Settings** instances and different locales used shared **_settings.DATE_ORDER** before previous thread set default _settings.DATE_ORDER back,  which is wrong for their locales and thus returned wrong dates.
**_Solution:_**  Create separate **Settings** instances not only for different incoming settings, but for all other _kwargs_ e.g. different locales, languages, etc. 

**_other notes:_**
I noticed that the **replace()** method changed the original settings dict as it's a mutable object, so I’ve modified it to work with a _copy()_ of it. As settings can be passed to _kwargs_ as a **None**- added a check for this case.

### DIFF
--- a/dateparser/conf.py
+++ b/dateparser/conf.py
@@ -31,19 +31,21 @@ class Settings:
     _pyfile_data = None
     _mod_settings = dict()
 
-    def __init__(self, settings=None):
-        if settings:
-            self._updateall(settings.items())
-        else:
+    def __init__(self, **kwargs):
+        if not kwargs.get('settings'):
             self._updateall(self._get_settings_from_pyfile().items())
 
-    @classmethod
-    def get_key(cls, settings=None):
-        if not settings:
-            return 'default'
+        elif len(self.__dict__) == 1:
+            self._updateall(kwargs['settings'].items())
 
-        keys = sorted(['%s-%s' % (key, str(settings[key])) for key in settings])
-        return hashlib.md5(''.join(keys).encode('utf-8')).hexdigest()
+    @classmethod
+    def get_key(cls, **kwargs):
+        if kwargs:
+            keys = [f'{key}-{val}' for key, val in kwargs.pop('settings').items()]
+            keys.extend([f'{key}-{val}' for key, val in kwargs.items() if val])
+            return hashlib.md5(''.join(sorted(keys)).encode('utf-8')).hexdigest()
+
+        return 'default'
 
     @classmethod
     def _get_settings_from_pyfile(cls):
@@ -57,18 +59,21 @@ class Settings:
             setattr(self, key, value)
 
     def replace(self, mod_settings=None, **kwds):
-        for k, v in kwds.items():
+        _settings = kwds.get('settings', {}).copy()
+        for k, v in _settings.items():
             if v is None:
                 raise TypeError('Invalid {{"{}": {}}}'.format(k, v))
 
-        for x in self._get_settings_from_pyfile().keys():
-            kwds.setdefault(x, getattr(self, x))
+        z = self._get_settings_from_pyfile().keys()
+        for x in z:
+            _settings.setdefault(x, getattr(self, x))
 
-        kwds['_default'] = False
+        _settings['_default'] = False
         if mod_settings:
-            kwds['_mod_settings'] = mod_settings
+            _settings['_mod_settings'] = mod_settings
 
-        return self.__class__(settings=kwds)
+        kwds['settings'] = _settings
+        return self.__class__(**kwds)
 
 
 settings = Settings()
@@ -77,11 +82,15 @@ settings = Settings()
 def apply_settings(f):
     @wraps(f)
     def wrapper(*args, **kwargs):
-        mod_settings = kwargs.get('settings')
-        kwargs['settings'] = mod_settings or settings
+        mod_settings = kwargs.get('settings', {})
+        if mod_settings is None:
+            kwargs['settings'] = mod_settings = {}
 
-        if isinstance(kwargs['settings'], dict):
-            kwargs['settings'] = settings.replace(mod_settings=mod_settings, **kwargs['settings'])
+        if kwargs:
+            if isinstance(mod_settings, dict):
+                kwargs['settings'] = settings.replace(mod_settings=mod_settings.copy(), **kwargs)
+        else:
+            kwargs['settings'] = settings
 
         if not isinstance(kwargs['settings'], Settings):
             raise TypeError("settings can only be either dict or instance of Settings class")
--- a/dateparser/freshness_date_parser.py
+++ b/dateparser/freshness_date_parser.py
@@ -16,8 +16,6 @@ PATTERN = re.compile(r'(\d+)\s*(%s)\b' % _UNITS, re.I | re.S | re.U)
 
 class FreshnessDateDataParser:
     """ Parses date string like "1 year, 2 months ago" and "3 hours, 50 minutes ago" """
-    def __init__(self):
-        self.now = None
 
     def _are_all_words_units(self, date_string):
         skip = [_UNITS,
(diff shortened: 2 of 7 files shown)


dateparser/conf.py
```diff
@@ -31,19 +31,21 @@ class Settings:
     _pyfile_data = None
     _mod_settings = dict()
 
-    def __init__(self, settings=None):
-        if settings:
-            self._updateall(settings.items())
-        else:
+    def __init__(self, **kwargs):
```
Gallaecio, line 34: Why the switch to `**kwargs`?

dateparser/conf.py
```diff
@@ -31,19 +31,21 @@ class Settings:
...
+        if not kwargs.get('settings'):
             self._updateall(self._get_settings_from_pyfile().items())
 
+        elif len(self.__dict__) == 1:
```
Gallaecio, line 38: I find this hard to read, as I have no idea what `self.__dict__` is meant to contain at this point. What is this checking for? Any way we can make this check more readable?

dateparser/conf.py
```diff
@@ -57,18 +59,21 @@ def _updateall(self, iterable):
             setattr(self, key, value)
 
     def replace(self, mod_settings=None, **kwds):
```
Gallaecio, line 61: Why is `kwds` now a `dict` that contains a `settings` key?

dateparser/conf.py
alex-triphonov: To pass other keyword arguments to Settings class which is required to get different `registry_key` for settings with different locale/language arguments.

dateparser/conf.py
alex-triphonov: On each init Settings checks incoming params and create new instance or take one from `__registry_dict`.\
New ones on init have only one attribute: `registry_key` and others are set by `self._updateall`.\
This check is checking if it's a new instance to avoid updating existing existing instances as this does not make any good and leads to concurrency issues.\
Not sure we can make this check any nicer..

dateparser/conf.py
alex-triphonov: To pass other keyword arguments to Settings class which is required to get different registry_key for settings with different locale/language arguments.

Gallaecio:
After ~1h playing around with the changes, I _think_ I’m starting to get a grasp of some of it.

At the moment there are 2 questions in my mind:
- Would it me possible to split the change in 2, “now” and “settings”? Or does it need to be in the same pull request due to the tests?
- I wonder if we should not consider a bigger rewrite of how settings work. I feel like the logic we have there to prevent duplicate setting objects is basically being voided by these changes. And even if that is not the case, I wonder if the optimization of avoiding duplicate objects is worth it. Maybe we need to rethink `Settings` as a whole.
- Even with big changes, I would strive for backward compatibility. In their current form, these changes make some API changes which I would like to avoid, if possible. Unless we want to post-pone these changes to dateparser 2.0.

alex-triphonov:
@Gallaecio
1) Yes, tests will fail without `now` PR.
2) Absolutely agree, I guess settings in existing form were made to use less instances of settings and as this doesn't work i.e. leads to concurrency issues, whole settings conception should be reconsidered, but this is's an epic..
3)  What backward compatibility exactly concerns you? I thought I made minimum possible changes, that most users shouldn't feel.

dateparser/conf.py
noviluni, line 34: @Gallaecio @alex-triphonov 
would it be possible to keep the `settings=None` and, in case it's used, raise a deprecation warning and preserve the old behavior? 

Something like:
```
def __init__(self, settings=None, **kwargs):
    if settings:
        # raise exception (deprecation warning, not concurrent) + old behavior
    else:
        # new behavior using kwargs
```

I know the code won't be nice, but I think in that way we could keep the backward compatibility until the next major version.

(You should do the same for `get_key()` and maybe something similar in the other two methods you touched).

Maybe I'm saying something stupid, sorry guys, I didn't have too much time to play with the code.

alex-triphonov: No this is not possible- settings are still being used, but in kwargs for more convenience, the only way to separate  behaviours is to add some new _kwarg_ like `new_settings` but is old behaviour with fewer `Settings` instances that much essential?

Gallaecio: To me it’s mostly a matter of API. The current change would require to postpone this to dateparser 2.x, as it is backward compatible. If we can figure a backward compatible implementation, this can go on the next dateparser version.

dateparser/conf.py
```diff
@@ -77,11 +81,15 @@ def replace(self, mod_settings=None, **kwds):
 def apply_settings(f):
     @wraps(f)
     def wrapper(*args, **kwargs):
-        mod_settings = kwargs.get('settings')
-        kwargs['settings'] = mod_settings or settings
+        mod_settings = kwargs.get('settings', {})
+        if mod_settings is None:
+            kwargs['settings'] = mod_settings = {}
```
noviluni, line 86: I would avoid this double assignment. You could do something like this:

```suggestion
            kwargs['settings'], mod_settings = {}, {}
```

There are multiple reasons, but basically to ease the code review. You can find more context here: https://stackoverflow.com/questions/4172223/double-assignment-should-it-be-avoided

noviluni:
@alex-triphonov you did a good job here :muscle:, but as mentioned before we can't merge it as-is.

The point is not "how essential it is", the issue here is that when declaring a public class or function (in this case `Settings` and `replace()`) in a library used by other people you can’t change the behavior without increasing the “major” version because some users could be using the `Settings` class in their way and this would be a breaking change for them.

However, fixing multithreading is a must and a priority, so I will check it further to see if I find a way to support both "settings" versions to avoid a really premature `dateparser==2.0.0`.

thanks!

dateparser/conf.py
alex-triphonov: Sure, done.

alex-triphonov:
Ok I get it. So in order to add backward incompatibility, we can add a new flag to settings, like `SAFE_CONCURRENCY`
and then separate behaviours like:
`def __init__(self, settings=None, **kwargs):
    if kwargs['SAFE_CONCURRENCY']:
        # new behavior using kwargs
    else:
        # raise warning (deprecation warning, not concurrent) + old behavior`
What do you think?

noviluni:
Hi @alex-triphonov

> Ok I get it. So in order to add backward incompatibility, we can add a new flag to settings, like `SAFE_CONCURRENCY`
> and then separate behaviours like:
> 
> ```
> def __init__(self, **kwargs):
>     if kwargs.get('SAFE_CONCURRENCY'):
>         # new behavior using kwargs
>     else:
>         # raise warning (deprecation warning, not concurrent) + old behavior
> ```
> 
> This is relatively easy to implement, but then everyone who doesn't care about concurrency
> will have to add this flag in order to get rid of deprecation warning..
> So maybe warning should be avoided then. What do you think?


Sorry, I think I would need to add some tests to fully understand if these changes related to settings creation are breaking changes or not. 

But for the `replace` method it seems obvious, as we needed to change the tests, I would like this: `settings = settings.replace(NORMALIZE=False)` and this `settings = settings.replace(settings={'NORMALIZE': False})` to work in the same way.

Couldn't be possible to do something like this?:

```python
 def replace(self, mod_settings=None, **kwds):
     if any(kwarg for kwarg in kwds if kwarg not in ('settings')):
         # warning + old behavior
     else:
         # new behavior
```

alex-triphonov:
> But for the `replace` method it seems obvious, as we needed to change the tests, I would like this: `settings = settings.replace(NORMALIZE=False)` and this `settings = settings.replace(settings={'NORMALIZE': False})` to work in the same way.
> 
> Couldn't be possible to do something like this?:
> 
> ```python
>  def replace(self, mod_settings=None, **kwds):
>      if any(kwarg for kwarg in kwds if kwarg not in ('settings')):
>          # warning + old behavior
>      else:
>          # new behavior
> ```
No I don't think we can do it this way because we have kwargs besides `settings`: `locales`, `date_formats` etc..
I can add some additional tests if you describe your concerns.

alex-triphonov:
>@noviluni  added this to the 2.0.0 milestone 2 days ago
Does it mean adding different behaviour for `Settings` when using special key in it is not backwards compatible enough and we'll not see concurrency safety in near future?

noviluni:
Hey @alex-triphonov 

no, that milestone means that, as of now, the PR couldn't be merged as-is before a `dateparser==2.0.0` version. Of course, if we fix it, we can merge it before.

It's ok to use the `SAFE_CONCURRENCY` as a temporary workaround, you can go ahead and implement it :)

alex-triphonov:
@noviluni 
So I returned to the topic and checked the possibility to add `SAFE_CONCURRENCY` flag to settings, and I need to clarifu something.
1) The only change to `Settings` behaviour is creating separate `Settings` instances not only for different incoming settings, but for all other kwargs e.g. different locales, languages, etc. Do we want to separate it from old behaviour? because then we just should check this flag in `get_key` method and don't make extra instances.
2) 
`But for the replace method it seems obvious, as we needed to change the tests, I would like this: settings = settings.replace(NORMALIZE=False) and this settings = settings.replace(settings={'NORMALIZE': False}) to work in the same way.`\
I can do it by moving `settings_values` from `check_settings` function and check kwargs in `replace` against it's keys. is it ok?

PR #847, opened 2020-11-26
remove convert_to_local_tz

noviluni:
This function is not used, is not documented, it's not tested and there isn't any reference anywhere (https://github.com/scrapinghub/dateparser/search?q=convert_to_local_tz).

However removing it directly could be considered as a breaking change.


Gallaecio:
Time for a 2.0 milestone?  :slightly_smiling_face:

PR #851, opened 2020-12-03
Make order of previous locales deterministic

surkova:
When using `set` for storing previously seen locales we eventually end up in the land of non-determinism when looping through them:
```
Python 3.8.5 (default, Jul 21 2020, 10:48:26)
[Clang 11.0.3 (clang-1103.0.32.62)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> a = set()
>>> a.add("ru")
>>> a.add("es")
>>> a
{'es', 'ru'}
>>> a.add("en")
>>> a
{'es', 'en', 'ru'}
...
```

Since python 3.7 insertion order is [preserved](https://mail.python.org/pipermail/python-dev/2017-December/151283.html) in dictionaries, so we can take advantage of this. 

I'm not sure what the policy is as for support of python < 3.7, on travis we seem to be running even 3.5, then my test will fail there. However, when looking at [this](https://github.com/scrapinghub/dateparser/issues/549) discussion, I see there's an appetite for making this logic deterministic. Approach with `dict` should not be slowing parsing either:
```
>>> b = {}
>>> b.update({"es": "es"})  # first string parsed
>>> b
{'es': 'es'}
>>> b.update({"ru": "ru"})  # second string parsed
>>> b
{'es': 'es', 'ru': 'ru'}
>>> b.update({"es": "es"})  # third string parsed
>>> b
{'es': 'es', 'ru': 'ru'}
...
```

### DIFF
--- a/dateparser/date.py
+++ b/dateparser/date.py
@@ -363,7 +363,7 @@ class DateDataParser:
         self.languages = languages
         self.locales = locales
         self.region = region
-        self.previous_locales = set()
+        self.previous_locales = collections.OrderedDict()
 
     def get_date_data(self, date_string, date_formats=None):
         """
@@ -423,7 +423,7 @@ class DateDataParser:
             if parsed_date:
                 parsed_date['locale'] = locale.shortname
                 if self.try_previous_locales:
-                    self.previous_locales.add(locale)
+                    self.previous_locales[locale] = None
                 return parsed_date
         else:
             return DateData(date_obj=None, period='day', locale=None)
@@ -457,7 +457,7 @@ class DateDataParser:
                 yield stripped_date_string
 
         if self.try_previous_locales:
-            for locale in self.previous_locales:
+            for locale in self.previous_locales.keys():
                 for s in date_strings():
                     if self._is_applicable_locale(locale, s):
                         yield locale
--- a/tests/test_date.py
+++ b/tests/test_date.py
@@ -410,6 +410,14 @@ class TestDateDataParser(BaseTestCase):
         self.when_date_string_is_parsed('2020-05-01')
         self.then_detected_locale('es')
 
+    def test_try_previous_locales_order_deterministic(self):
+        self.given_parser(try_previous_locales=True)
+        self.when_date_string_is_parsed('Mañana')  # es
+        self.then_detected_locale('es')
+        self.when_date_string_is_parsed('Понедельник')  # ru
+        self.then_detected_locale('ru')
+        self.then_previous_locales_is(["es", "ru"])
+
     @parameterized.expand([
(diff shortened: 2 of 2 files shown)


surkova:
I can change to use `collections.OrderedDict` if we want to keep this change and need to support < 3.7.

dateparser/date.py
```diff
@@ -423,7 +423,7 @@ def get_date_data(self, date_string, date_formats=None):
             if parsed_date:
                 parsed_date['locale'] = locale.shortname
                 if self.try_previous_locales:
-                    self.previous_locales.add(locale)
+                    self.previous_locales.update({locale: locale})
```
noviluni, line 426:
```suggestion
                    self.previous_locales[locale] = None
```

I think adding `None` as value makes sense here as we only need the keys. On the other hand, assigning the value with `[]` is about three times faster than using `.update`, so I would prefer to use it. :)

dateparser/date.py
```diff
@@ -363,7 +363,7 @@ def __init__(self, languages=None, locales=None, region=None, try_previous_local
         self.languages = languages
         self.locales = locales
         self.region = region
-        self.previous_locales = set()
+        self.previous_locales = dict()
```
noviluni, line 366: We will drop the Python 3.5 support, so it could be ok to leave `dict`, as it's ordered since Python 3.6.

**However**, as explained here: https://stackoverflow.com/questions/39980323/are-dictionaries-ordered-in-python-3-6#answer-39980744 it's just an implementation detail and it's since Python 3.7 that we can depend on this.

So I would move from `dict` to `collections.OrderedDict`. We can always change this in the future.

noviluni (commented):
Good catch @surkova!

Could you kindly address what I added in the comments?

Thanks!

noviluni (approved):

dateparser/date.py
```diff
@@ -457,7 +457,7 @@ def date_strings():
                 yield stripped_date_string
 
         if self.try_previous_locales:
-            for locale in self.previous_locales:
+            for locale in self.previous_locales.keys():
```
Gallaecio, line 460: :lipstick: 

```suggestion
            for locale in self.previous_locales:
```

I’m just mentioning this in case `.keys()` was not added on purpose. But given `.keys()` is more explicit, maybe the current code is more readable. :heavy_check_mark: either way.

Gallaecio (approved):

dateparser/date.py
noviluni: I think we can leave it as-is. For me, it seems more accurate to explicitly indicate that we iterate through the keys. :+1:

noviluni:
Thanks again @surkova! :rocket:

PR #876, opened 2021-02-07
Added support for fractional units

### LINKED ISSUE #753: Parsing of fractional hours is not supported
`dateparser.parse('in 2 hours')` yields a correct result, but parsing `'in 1.5 hours'` is taken to mean `'in 5 hours'`.

Of course, supporting things like `'in 1.2 hours'` would be weird (that's most likely `'in 1 hour and 12 minutes'`), but `'in 1.5 hours'` sounds like a common-enough use-case that people would expect to work.

Initially reported here: https://github.com/anoadragon453/matrix-reminder-bot/issues/37

DenverCoder1:
Fixes #753

Added support for fractional units by making the regex pattern support digits before a decimal point and converting `num` to a `float` instead of an `int` on line 150.

### What this changes

Before this fix, `parse("in 0.5 hours")` would return the time **5 hours** in the future instead of **0.5 hours** in the future.

With this fix, parsing fractional units are no longer parsed incorrectly.

### Tests

```py
>>> parse("now")
datetime.datetime(2021, 2, 7, 18, 50, 10, 107501)
>>> parse("in 0.5 hours")
datetime.datetime(2021, 2, 7, 19, 20, 15, 773800)
>>> parse("in 3.75 hours") 
datetime.datetime(2021, 2, 7, 22, 35, 20, 463238)
>>> parse("2.5 hours ago")  
datetime.datetime(2021, 2, 7, 16, 20, 29, 614937)
```


dateparser/freshness_date_parser.py
```diff
@@ -11,7 +11,7 @@
 
 
 _UNITS = r'decade|year|month|week|day|hour|minute|second'
-PATTERN = re.compile(r'(\d+)\s*(%s)\b' % _UNITS, re.I | re.S | re.U)
+PATTERN = re.compile(r'(\d*\.?\d+)\s*(%s)\b' % _UNITS, re.I | re.S | re.U)
```
Gallaecio, line 14: We should probably support commas as well, for better multi-language support.

I think 1000+ values are unlikely in this scenario, so I’m personally OK with not taking into account the possibility of thousand separators messing things up.

Gallaecio (commented):
Could you also extend tests to cover this new feature? I’m think we should at least test 1 such value, and assuming we include comma support the same value with a comma instead of a period.

noviluni:
HI @DenverCoder1, thank you for this PR.

Could you add those tests you mention in the description in the project tests? Thanks!

DenverCoder1:
@Gallaecio Added some tests for fractional units. Also added some support for numbers with commas instead of periods. Adding comma support was not as simple because of the way the `Dictionary` splits the strings.

Gallaecio:
I see you’ve had to modify the per-language Python files. Those are not meant to be modified manually, and are instead generated from other files in the repository. See https://dateparser.readthedocs.io/en/latest/contributing.html#guidelines-for-editing-translation-data . On the bright side, it may mean that you just need to edit the script that generates them.

DenverCoder1:
@Gallaecio

I see. So I should update `write_complete_data.py` and the `yaml` files then rerun `write_complete_data.py`?

Gallaecio:
I am not yet familiar with that process, but I would assume you would have to edit _either_ `write_complete_data.py` (hopefully) _or_ the yaml files (if they include the `\d` pattern), and then rerun `write_complete_data.py`.

This is also assuming that you can get the desired result editing one of those two files. If the patterns came directly from CLDR (unlikely but possible), it may get messier.

Please have a look at `en.yml` and `write_complete_data.py`, see if you can figure out where the change should go. Otherwise, @noviluni is the person most familiar with the process, so he probably knows.

DenverCoder1:
It seems to me that some of the data in the .py files is coming from the date_translation_data JSON files and some parts (such as decades) are coming from the supplementary YAML files. I made changes in both places to get the .py files to fully update.

Gallaecio:
The JSON files are not meant to be modified (they come from http://cldr.unicode.org/), so this may be not as trivial as I had hoped.

DenverCoder1:
> The JSON files are not meant to be modified (they come from [http://cldr.unicode.org](http://cldr.unicode.org/)), so this may be not as trivial as I had hoped.

I didn't need to modify the JSON, only had to change the way it is converted in write_complete_data.py.

Gallaecio (approved):
:heavy_check_mark: 

@noviluni Mind that tests did not run in CI for some reason.

Gallaecio:
Looking at https://github.com/scrapinghub/dateparser/pull/859 maybe it’s just a matter or reopening.

noviluni:
Hi @DenverCoder1 @Gallaecio  Thank you for the PR.

I like the approach, however, I'm don't want to merge this as-is, because not all the locale/regions accept both (period and comma) as decimal separator. This could lead to some errors.

For example, if we have `1.500 s` we don't know if it's 1 point 5 seconds or 1 thousand 500 seconds. In UK and US, they normally use comma (`,`) as a thousand separator, while in Spanish we usually use the period (`.`) as a thousand separator. 

I tried with this code, and this is what I got:

```python
>>> datetime.now() - dateparser.parse('1,5000 s', languages=['en'])
datetime.timedelta(0, 4999, 995131)

>>> datetime.now() - dateparser.parse('1.5000 s', languages=['en'])
datetime.timedelta(0, 1, 496419)
```

```python
>>> datetime.now() - dateparser.parse('1,5000 s', languages=['es'])
datetime.timedelta(0, 4999, 996924)

>>> datetime.now() - dateparser.parse('1.5000 s', languages=['es'])
datetime.timedelta(0, 1, 496970)
```

I personally think that the decimal/comma separator should be configured by default by locale and we should probably be able to configure it manually using the settings, so this approach wouldn't probably work.

Let me know what you think / your ideas :)

Gallaecio:
I think you have a point, but I also imagine that those examples are interpreted as `5000` with `master`. If so, I imagine this change would not break anything that is parsed fine at the moment, but only extend support for fractional numbers without thousand separators.

Given that, I would not mind merging as is and creating 2 separate tickets (1) for restricting supported decimal separators based on locale (we can hopefully use data from CLDR to choose the right ones) and (2) for supporting thousand separators.

But yes, we could also aim to solve all of that in this patch. I’m just worried about the extra time that could require.

DenverCoder1:
@Gallaecio @noviluni What's the status on this? I don't think I have the time now to try making it locale-based as suggested. I agree that it could be made a separate PR.

As of now, it will make things such as "2.5 hours", etc. parse correctly when right now they don't ("2.5" gets treated as "5").

To make it work with periods only required a couple simple changes. Making it work with commas as well was a bit more complicated due to the way the parser ignores commas, but it should also be working properly with these changes.

If there's a worry that parsing commas as decimal points could break things, that part could be removed for now until the locale settings are in place, or if not, it could be merged as is.

Let me know if there's anything else needed from me.

repl-samuel:
Could I ask the status of this PR?
It's approved almost a year ago but still not get merged?

Gallaecio:
We usually follow a 2-maintainer-approval policy before merging. The next step is for another maintainer to have a look. However, most (all?) dateparser maintainers are also maintainers of many other open source libraries, and our time is limited.

dateparser/freshness_date_parser.py
```diff
@@ -11,7 +11,7 @@
 
 
 _UNITS = r'decade|year|month|week|day|hour|minute|second'
-PATTERN = re.compile(r'(\d+)\s*(%s)\b' % _UNITS, re.I | re.S | re.U)
+PATTERN = re.compile(r'(\d+[.,]?\d*)\s*(%s)\b' % _UNITS, re.I | re.S | re.U)
```
DenverCoder1, line 14: 1. Added decimal point to unit parser regex

dateparser/freshness_date_parser.py
```diff
@@ -147,7 +147,7 @@ def get_kwargs(self, date_string):
 
         kwargs = {}
         for num, unit in m:
-            kwargs[unit + 's'] = int(num)
+            kwargs[unit + 's'] = float(num.replace(",", "."))
```
DenverCoder1, line 150: 2. Convert unit type to float instead of int, (replacing comma with period for [multi-language support](https://github.com/scrapinghub/dateparser/pull/876#discussion_r571985940))

dateparser_scripts/write_complete_data.py
```diff
@@ -31,7 +31,7 @@ def _modify_relative_data(relative_data):
     modified_relative_data = OrderedDict()
     for key, value in relative_data.items():
         for i, string in enumerate(value):
-            string = RELATIVE_PATTERN.sub(r'(\\d+)', string)
+            string = RELATIVE_PATTERN.sub(r'(\\d+[.,]?\\d*)', string)
```
DenverCoder1, line 34: 3. Update the data converter script to include decimal places in regexes

tests/test_freshness_date_parser.py
```diff
@@ -96,6 +96,10 @@ def test_relative_past_dates_with_time_as_period(self, date_string, ago, period)
         param('eight months ago', ago={'months': 8}, period='month'),
         param('six days ago', ago={'days': 6}, period='day'),
         param('five years ago', ago={'years': 5}, period='year'),
+        # Fractional units
+        param('2.5 hours', ago={'hours': 2.5}, period='day'),
+        param('10.75 minutes', ago={'minutes': 10.75}, period='day'),
+        param('1.5 days', ago={'days': 1.5}, period='day'),
```
DenverCoder1, line 102: 5. Add test cases for fractional units (tests added to 4 places within this file including with commas)

DenverCoder1 (commented):
To make it easier for future reviewers, the change is much smaller than it appears to be. The reason there are many files changed is because the regex in `write_complete_data.py` was updated and ran.

See below for shortcuts to the code changes.

jgayfer:
Would love to get some more visibility on this! ♥️

DenverCoder1:
Updating the base branch will take some extra work due to some changes in the language data.

#1079 specifically has caused some issues since they changed the CLDR data instead of the supplementary yaml. I can open a separate PR to fix that.

DenverCoder1:
All languages should be working again now.

Note: The fix in #1090 is also incorporated in this PR.

dateparser_scripts/write_complete_data.py
Gallaecio: While we wait for a 2nd maintainer approval, I wonder if we should go for a more strict regular expression here. Something like `(\d+(?:[.,]\d+)?` maybe?

DenverCoder1: @Gallaecio I've spent hours trying more specific regexes but for some reason, any regex I try besides the current one ends up with hundreds of failed tests and strings not translating. Feel free to try some things out and see if you can get it to work, but I think sticking to this regex that works is easiest.

PR #881, opened 2021-02-13
Reducing cyclomatic complexity of search and addressing offset awareness issue

SimplicityGuy:
Includes 2 contributions:
1. Since the `search` functionality is complex, the first change addresses cyclomatic complexity and deep indentation in the search code. This should make the code easier to read and maintain.
2. Addresses #679 by handling timezone offset awareness.

### DIFF
--- a/dateparser/parser.py
+++ b/dateparser/parser.py
@@ -450,7 +450,8 @@ class _parser:
 
         # NOTE: If this assert fires, self.now needs to be made offset-aware in a similar
         # way that dateobj is temporarily made offset-aware.
-        assert not (self.now.tzinfo is None and dateobj.tzinfo is not None), "Review comment for details."
+        assert not (self.now.tzinfo is None and dateobj.tzinfo is not None),\
+            "`self.now` doesn't have `tzinfo`. Review comment in code for details."
 
         # Store the original dateobj values so that upon subsequent parsing everything is not
         # treated as offset-aware if offset awareness is changed.
--- a/dateparser/parser.py
+++ b/dateparser/parser.py
@@ -448,17 +448,18 @@ class _parser:
 
             dateobj = dateobj + delta
 
-        # Store the original self.now and dateobj values so that upon subsequent parsing
-        # everything is not treated as offset-aware if offset awareness is changed.
-        self.original_now = self.now
+        # NOTE: If this assert fires, self.now needs to be made offset-aware in a similar
+        # way that dateobj is temporarily made offset-aware.
+        assert not (self.now.tzinfo is None and dateobj.tzinfo is not None), "Review comment for details."
+
+        # Store the original dateobj values so that upon subsequent parsing everything is not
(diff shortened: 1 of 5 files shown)


SimplicityGuy:
Oooo, ok. I'll add an additional test here.

SimplicityGuy:
hey @noviluni So seems like the CC is complaining because I haven't hit [this line](https://codecov.io/gh/scrapinghub/dateparser/pull/881/diff?src=pr&el=tree#D1-461).

Totally understand. Here's the problem... after code tracing, it doesn't seem that `dateobj` will ever have `tzinfo` as the code is currently written, so in practice this line of code is dead. My concern is that if the code changes to now start adding at `tzinfo`, this bug will be reintroduced. While I can easily remove this path, I do not feel that for the longer term this is the right thing to do.

Thoughts?

Gallaecio:
What about adding an assert for `dateobj`’s `tzinfo` being `None`, with a comment that explains that if it ever becomes possible for that to be something other than `None`, the code needs to change, and describes exactly how it needs to change?

SimplicityGuy:
> What about adding an assert for `dateobj`’s `tzinfo` being `None`, with a comment that explains that if it ever becomes possible for that to be something other than `None`, the code needs to change, and describes exactly how it needs to change?

Great idea! Totally forgot about using an assert here. Changing it now.

Gallaecio (approved):

dateparser/parser.py
```diff
@@ -447,6 +448,19 @@ def _correct_for_time_frame(self, dateobj):
 
             dateobj = dateobj + delta
 
+        # NOTE: If this assert fires, self.now needs to be made offset-aware in a similar
+        # way that dateobj is temporarily made offset-aware.
+        assert not (self.now.tzinfo is None and dateobj.tzinfo is not None), "Review comment for details."
```
noviluni, line 453: @SimplicityGuy 
I know that it doesn't seem to happen, but Imagine we introduce the bug without noticing and some user gets this error. I think we should do the error message a bit more self-explanatory. What about something like...?:
```suggestion
        assert not (self.now.tzinfo is None and dateobj.tzinfo is not None), "`self.now` shouldn't have tzinfo. Review comment in code for details."
```

or
 
```suggestion
        assert not (self.now.tzinfo is None and dateobj.tzinfo is not None), "Timezone info is wrong. Review comment in code for details."
```

In that way, people affected by this (either developers or users) will be able to provide more context if they find this error.

@Gallaecio let me know what you think

noviluni (commented):
Hey @SimplicityGuy! Good job here!

The `search` functionality has to be refactored and checked because it is hard to read and there are a bunch of known bugs: https://github.com/scrapinghub/dateparser/issues?q=is%3Aissue+is%3Aopen+label%3Asearch_dates

Thank you very much for addressing this! I left one minor comment.

noviluni (approved):

noviluni:
Thank you!

PR #885, opened 2021-02-18
remove self.now from freshness_date_parser

noviluni:
Removes the `self.now` from the `freshness_date_parser` to make it more compatible with concurrency/threads (partial fix).

closes: https://github.com/scrapinghub/dateparser/pull/814
simplifies: https://github.com/scrapinghub/dateparser/pull/834 (as it can't be merged as-is)

fixes partially: https://github.com/scrapinghub/dateparser/issues/813, https://github.com/scrapinghub/dateparser/issues/441


Gallaecio (approved):

ageorge95:
That fixed my problem of using threading and dateparser.
Thanks @noviluni 👍

PR #893, opened 2021-02-28
Make command for venv and dependencies

### LINKED ISSUE #892: Make virtual environment for dateparser
Hi, I propose a command to make a virtual environment in dateparser, thus making it easier for developers to contribute to the project.  This is my plan for the same:

1. Add command in the `Makefile` for creating a virtual environment, which can be invoked like:
```
make venv
```
This command will also automatically install all the dependencies from `requirements.txt` to the venv

2. Add the thus created venv to `.gitignore` file, making it further easier for developers to not include this while adding to git

What do you guys think of this feature? I would like to work on it if found useful! 😄

jay24rajput:
- A command `make venv` for making a virtual environment and installing all the required dependencies in the virtual environment
- The virtual environment(`venv`) and `.vscode`  folders added to `.gitignore`
- Dependencies added for running tests and other commands smoothly in virtual environment

Fixes #892

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,6 +47,10 @@ docs/_build
 # Editors
 *.swp
 .idea
+.vscode
 
 # Other
 raw_data
+
+# Virtual environment
+venv/
--- a/Makefile
+++ b/Makefile
@@ -59,3 +59,11 @@ dist: clean
 	python setup.py sdist
 	python setup.py bdist_wheel
 	ls -l dist
+
+venv:
+	# Create a virtual environment which can be used to run the build package.
+	python3 -m venv venv && \
+	source venv/bin/activate && \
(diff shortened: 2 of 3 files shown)


noviluni:
Hi @jay24rajput, thank you for the effort of adding this to the Makefile.

However, after an internal discussion, we decided to remove the `Makefile` as we usually don't use it: https://github.com/scrapinghub/dateparser/pull/907

It's also true that creating a virtualenv is something that we usually do only once, so I'm not sure if it makes sense to add it to a Makefile, it would be better to document it well in the docs.

This PR and the related issue will be closed after merging the PR (https://github.com/scrapinghub/dateparser/pull/907). If you have any other question or concern let me know :slightly_smiling_face:

PR #900, opened 2021-03-30
Fixing Issue  #894

gavishpoddar:
It seems removing this line resolves issue #894

Please check and merge the pull request

Before 

`[('March 3, 2021, 12:47 am', datetime.datetime(2021, 3, 3, 12, 47))]
`
After

`[('March 3, 2021, 00:47', datetime.datetime(2021, 3, 3, 0, 47))]
`

PR #901, opened 2021-03-30
Fixing #894 and correcting tests

### LINKED ISSUE #894: search_dates() wrong result for 12:xx am
search_dates() gives wrong results for times between 12:00 am and 12:59 am (hour is off by 12), but parse() is correct.

```
>>> import dateparser
>>> from dateparser.search import search_dates 
>>> dateparser.__version__
'1.0.0'
>>> 
>>> search_dates('March 3, 2021, 12:47 am')
[('March 3, 2021, 12:47 am', datetime.datetime(2021, 3, 3, 12, 47))]
>>>
>>> dateparser.parse('March 3, 2021, 12:47 am')
datetime.datetime(2021, 3, 3, 0, 47)
```
Python version is 3.8.9, OS is Ubuntu 20.10

gavishpoddar:
Hi, @Gallaecio ,

I have fixed #894 and submitted PR #900 but it failed because the tests didn't supported `am` so tests failed.

In this PR the tests and #894 is fixed.


dateparser/search/search.py
```diff
@@ -86,10 +86,15 @@ def split_if_not_parsed(self, item, original):
...
         return possible_splits
 
-    def parse_item(self, parser, item, translated_item, parsed, need_relative_base):
+    def parse_item(self, parser, item, translated_item, parsed, need_relative_base, language):
         relative_base = None
-        item = item.replace('ngày', '')
-        item = item.replace('am', '')
+
+        if language != "en":
+            item = item.replace('ngày', '')
```
noviluni, line 93: "ngày" is not English, but Vietnamese ("vi")

dateparser/search/search.py
```diff
@@ -86,10 +86,15 @@ def split_if_not_parsed(self, item, original):
...
+            item = item.replace('am', '')
```
noviluni, line 94: why should we apply this only in English? There's a lot of languages that use `am` and `pm`.

dateparser/search/search.py
```diff
@@ -86,10 +86,15 @@ def split_if_not_parsed(self, item, original):
...
+
+
+        #CHANGE HERE
```
noviluni, line 97: I don't think this comment should be here.

dateparser/search/search.py
```diff
@@ -153,9 +158,12 @@ def search_parse(self, shortname, text, settings):
             languages = [shortname]
             to_parse = original
 
+        
+
+
```
noviluni, line 163: why did you add these three lines?

noviluni (commented):
I added some doubts I have. I'm not also sure if we can just modify the `parse_item` and the `parse_found_objects` signature directly...

dateparser/search/search.py
gavishpoddar:
```suggestion
        #CHANGE HERE
```


Thanks, I have removed the comments

dateparser/search/search.py
gavishpoddar: Surething, even Vietnamese is not causing the issue it's actually German.

ie,

`am 8 ` means `on the 8th`

but 

`8 am` actually means `8 am`

so however it is done it presents problem one way or other.

dateparser/search/search.py
```diff
@@ -86,10 +86,12 @@ def split_if_not_parsed(self, item, original):
                 possible_splits.extend(self.split_by(item, original, splitter))
         return possible_splits
 
-    def parse_item(self, parser, item, translated_item, parsed, need_relative_base):
+    def parse_item(self, parser, item, translated_item, parsed, need_relative_base, language):
         relative_base = None
-        item = item.replace('ngày', '')
-        item = item.replace('am', '')
+
+        if language == "de":
```
gavishpoddar, line 92: For the German language making an edge case

gavishpoddar:
Put together 

The German language date parsing needs some correction but this PR fixes `AM PM` for other languages including English.

tests/test_search.py
```diff
@@ -461,15 +461,16 @@ def test_search_and_parse(self, shortname, string, expected, settings=None):
                     January UTC
                     June 5 am utc
                     June 23th 5 pm EST
-                    May 31, 8am UTC""",
+                    May 31
+                    8am UTC""",
```
Gallaecio, line 465: Why this change in the test input?

gavishpoddar: These changes just look clear the output is the same. Should I change it to how it was.

From
```
May 31, 8am UTC
```

To 

```
May 31
8am UTC
```

Gallaecio: Yes, I think that would be best. If we stopped supporting that comma in there, this change would prevent the test from failing accordingly. Also, I think `May 31, 8am UTC` is more readable, as that is meant to be parsed as a single date (regardless of whether or not `search_dates` does so at the moment).

tests/test_search.py
```diff
@@ -461,15 +461,16 @@ def test_search_and_parse(self, shortname, string, expected, settings=None):
...
               [('May 2020', datetime.datetime(2020, 5, datetime.datetime.utcnow().day, 0, 0)),
-               ('June 2020', datetime.datetime(2020, 6, datetime.datetime.utcnow().day, 0, 0)),
-               ('2023', datetime.datetime(2023, 6, datetime.datetime.utcnow().day, 0, 0)),
-               ('January UTC', datetime.datetime(2023, 1, datetime.datetime.utcnow().day, 0, 0, tzinfo=pytz.utc)),
-               ('June 5 am utc', datetime.datetime(2023, 6, 5, 0, 0, tzinfo=pytz.utc)),
+               ('June 2020', datetime.datetime(2020, 6, 30, 0, 0)),
```
Gallaecio, line 467: Won’t this cause these tests to fail whenever the tests are run on a day that is not the 30th of the month? If these tests failed for you, I wonder if maybe we should get here the current date (by local timezone) instead of the UTC day.

gavishpoddar: Corrected test's from the `30` to `datetime.datetime.utcnow().day` 

Regarding the UTC specific time zones are only specified if the input specifically states it

tests/test_search.py
noviluni: Hi @gavishpoddar, could you revert this as @Gallaecio requested? :)

dateparser/search/search.py
```diff
@@ -101,7 +103,7 @@ def parse_item(self, parser, item, translated_item, parsed, need_relative_base):
             parsed_item = parser.get_date_data(item)
         return parsed_item, is_relative
 
-    def parse_found_objects(self, parser, to_parse, original, translated, settings):
+    def parse_found_objects(self, parser, to_parse, original, translated, settings, language):
```
noviluni, line 106:
```suggestion
    def parse_found_objects(self, parser, to_parse, original, translated, settings, language=None):
```

This is to avoid a breaking change (as this function is declared as public, even if it should be private).

dateparser/search/search.py
```diff
@@ -86,10 +86,12 @@ def split_if_not_parsed(self, item, original):
                 possible_splits.extend(self.split_by(item, original, splitter))
         return possible_splits
 
-    def parse_item(self, parser, item, translated_item, parsed, need_relative_base):
+    def parse_item(self, parser, item, translated_item, parsed, need_relative_base, language):
```
noviluni, line 89:
```suggestion
    def parse_item(self, parser, item, translated_item, parsed, need_relative_base, language=None):
```

Again, this is to avoid a breaking change (as this function is declared as public, even if it should be private).

dateparser/search/search.py
noviluni: This seems like a good fix. In case of Vietnamese, I would add:

```python
if language == "vi":
    item = item.replace('ngày', '')
```

I don't know the reasons, but I would prefer to keep it there as it was probably because of some reason, so I would prefer to keep it.

When it was introduced is here: https://github.com/scrapinghub/dateparser/commit/7e7e22417e71e103225a2fdf05e68a6e67db041c#diff-4e4813a9f7a6e85270fb6054ada722c2a230066dc9fc9d442250b0da38acbfd0R26

dateparser/search/search.py
noviluni: @gavishpoddar, it makes totally sense, however, the next time someone would arrive here wouldn't know the reason, so I think you could explain this in a comment :)

dateparser/search/search.py
```diff
@@ -153,9 +155,12 @@ def search_parse(self, shortname, text, settings):
             languages = [shortname]
             to_parse = original
 
+        
+
+
```
noviluni, line 160: Why did you resolve this? I think you should remove these three new newlines.

gavishpoddar: The new lines are removed now

tests/test_search.py
gavishpoddar: I have made the changes

gavishpoddar:
Hi, @noviluni I have made the changes, and thanks for the review

tests/test_search.py
Gallaecio: As far as I can tell you have not, the change still appears in the pull request.

dateparser/search/search.py
```diff
@@ -86,10 +86,15 @@ def split_if_not_parsed(self, item, original):
...
 
-    def parse_item(self, parser, item, translated_item, parsed, need_relative_base):
+    def parse_item(self, parser, item, translated_item, parsed, need_relative_base, language=None):
         relative_base = None
-        item = item.replace('ngày', '')
-        item = item.replace('am', '')
+
+        # Replacing language specific words with special meaning
+        if language == "de":
+            item = item.replace('am', '')
```
noviluni, line 94:
```suggestion
        if language == "de":
            # Replace "am" because "am 8" means "on the 8th" but "8 am" actually means "8 am"
            item = item.replace('am', '')
```

noviluni: Hmmmm... Maybe we can use regex (`re.sub`) to remove it only when preceding a number? :thinking:

tests/test_search.py
Gallaecio: I’m thinking that I may not have been clear. The line used to have a comma, you changed it to a line break, and I’m saying it should be a comma again.

gavishpoddar: Thanks, @Gallaecio, now I understood it.

tests/test_search.py
```diff
@@ -466,10 +466,9 @@ def test_search_and_parse(self, shortname, string, expected, settings=None):
                ('June 2020', datetime.datetime(2020, 6, datetime.datetime.utcnow().day, 0, 0)),
                ('2023', datetime.datetime(2023, 6, datetime.datetime.utcnow().day, 0, 0)),
                ('January UTC', datetime.datetime(2023, 1, datetime.datetime.utcnow().day, 0, 0, tzinfo=pytz.utc)),
-               ('June 5 am utc', datetime.datetime(2023, 6, 5, 0, 0, tzinfo=pytz.utc)),
+               ('June 5 am utc', datetime.datetime(2023, 6, datetime.datetime.utcnow().day, 5, 0, tzinfo=pytz.utc)),
                ('June 23th 5 pm EST', datetime.datetime(2023, 6, 23, 17, 0, tzinfo=pytz.timezone("EST"))),
-               ('May 31', datetime.datetime(2023, 5, 31, 0, 0)),
-               ('8am UTC', datetime.datetime(2023, 8, 31, 0, 0, tzinfo=pytz.utc))]),
+               ('May 31, 8am UTC', datetime.datetime(2023, 5, 31, 8, 0, tzinfo=pytz.utc))]),
```
Gallaecio, line 471: Wait, so it was divided before? Sorry, I was convinced that it was `'May 31, 8am UTC'` before your changes for some reason. Could you revert again? I’m really sorry :sweat_smile:

gavishpoddar: Thanks, @Gallaecio, Even I got confused 😂.

I have fixed it now

dateparser/search/search.py
gavishpoddar: Hi @noviluni, I have made some can you please check and recommend

tests/test_search.py
```diff
@@ -461,15 +461,16 @@ def test_search_and_parse(self, shortname, string, expected, settings=None):
...
               [('May 2020', datetime.datetime(2020, 5, datetime.datetime.utcnow().day, 0, 0)),
                ('June 2020', datetime.datetime(2020, 6, datetime.datetime.utcnow().day, 0, 0)),
                ('2023', datetime.datetime(2023, 6, datetime.datetime.utcnow().day, 0, 0)),
                ('January UTC', datetime.datetime(2023, 1, datetime.datetime.utcnow().day, 0, 0, tzinfo=pytz.utc)),
-               ('June 5 am utc', datetime.datetime(2023, 6, 5, 0, 0, tzinfo=pytz.utc)),
+               ('June 5 am utc', datetime.datetime(2023, 6, datetime.datetime.utcnow().day, 5, 0, tzinfo=pytz.utc)),
                ('June 23th 5 pm EST', datetime.datetime(2023, 6, 23, 17, 0, tzinfo=pytz.timezone("EST"))),
                ('May 31', datetime.datetime(2023, 5, 31, 0, 0)),
-               ('8am UTC', datetime.datetime(2023, 8, 31, 0, 0, tzinfo=pytz.utc))]),
+               ('8am UTC', datetime.datetime(2023, 5, 31, 8, 0, tzinfo=pytz.utc))]),
```
Gallaecio, line 473: I think the parsing of `8am UTC` is a change for the better, but I’m not sure about `June 5 am utc`. The previous interpretation seems OK to me, “June 5th in the morning UTC”. In fact, I think it may even be better, because I suspect leaving a unit missing in the middle (i.e. having month and hour, but missing day) may be less likely than the other interpretation (month and day, am meaning sometime in the morning).

I checked and that test string was added as part of a performance improvement, https://github.com/scrapinghub/dateparser/pull/881, and I imagine it does not come from an actual case found in the internet, which makes it hard to argue either way.

gavishpoddar: Right it `“June 5th in the morning UTC”.` is a better interpretation. I am fixing that but I think it's beyond the scope of #894. 

So should I open a new issue or maybe directly create a new pull request or fix it in the current PR but that's beyond not directly related to issue #894.

Gallaecio: Well, I guess you are right that fixing it is out of the scope of this change and should be a separate issue. It was just bad luck that your change broke it, or rather it was just dumb luck that it was working in the first place.

However, what I would do here is move that scenario into a separate test, with the previous value, and mark the test as an expected failure (xfail), to make it clear that the expected outcome is still the old one, only that Dateparser does not support it yet.

gavishpoddar: Great after this PR is merged I will create an issue and create a supporting PR

Gallaecio: Don’t forget the xfail part, though, I do think that needs to be addressed in this pull request, instead of changing the test expectations.

gavishpoddar: I am working on it now and by the time you create separate test, with the previous value, and mark the test as an expected failure (xfail) it should be done

Gallaecio: > by the time you create separate test

I was hoping for you to do that, as it needs to be done as part of this pull request (which is the one that causes the test to fail).

gavishpoddar: Ok on it

gavishpoddar: It seems it is broken throughout `search_dates` and `parse` function as with other cases like 


`June 5th pm EST ---> [('June 5th pm EST', datetime.datetime(2021, 6, 24, 17, 0, tzinfo=<StaticTzInfo 'EST'>))]` 

or 

`June 5th pm EST ---> ('June 9 pm', datetime.datetime(2021, 6, 24, 21, 0))]`

Which is incorrect. I am working on the patch but it will require some time and significant change in other parts of the library and many tests will be broken because they are also incorrect. Or the current interpretation is correct.

gavishpoddar:
Thanks, for your support and suggestions these improvements are now made on PR #945.

Issue #854, opened 2020-12-08
There are two time styles that cannot be resolved (original: 有两种时间样式无法解析)


xinshen0819:
x = ['刚刚', '2020-11-0 13:36', '14分钟前']
    for i in x:
        temp_d = dateparser.parse(i)
        print("date_str:", i, "result:", temp_d)

date_str: 刚刚 result: None
date_str: 2020-11-0 13:36 result: None
date_str: 14分钟前 result: 2020-12-08 10:45:19.074193

noviluni:
Hi @xinshen0819, thanks for opening the issue.

1. For this: "刚刚" it should be the same as "now", right?

But, 

2. what did you expect to get with "2020-11-0 13:36"? I have never seen a date with the day "0" (zero) :thinking:, could you explain what did you expect to get?

xinshen0819:
非常感谢回复！
关于"2020-11-0 13:36"是我在尝试解析新闻发布时间时获取到的错误数据，因为我的过度猜测，没有考虑到数据的合理性，非常抱歉，通过测试，证明了我的过度猜测是错误的。
t_list = ['2020-11-0 13:36', '2020-11-1 13:36', '2020-1-1 13:36', '2020-1-1 13:36', '2020-01-01 13:36']
for i in t_list:
    temp_d = dateparser.parse(i)
    print("date_str:", i, "result:", temp_d)

('date_str:', '2020-11-0 13:36', 'result:', None)
('date_str:', '2020-11-1 13:36', 'result:', datetime.datetime(2020, 11, 1, 13, 36))
('date_str:', '2020-1-1 13:36', 'result:', datetime.datetime(2020, 1, 1, 13, 36))
('date_str:', '2020-1-1 13:36', 'result:', datetime.datetime(2020, 1, 1, 13, 36))
('date_str:', '2020-01-01 13:36', 'result:', datetime.datetime(2020, 1, 1, 13, 36))

关于“刚刚”，这个问题确实是存在的，也如您所说，就是“now”的意思。

PR #910, opened 2021-04-12
add support for 刚刚

noviluni:
Fixes: https://github.com/scrapinghub/dateparser/issues/854

Now `dateparser.parse("刚刚")` returns "now" instead of `None`.


peonone (approved):
LGTM @noviluni , thanks for adding this!

PR #911, opened 2021-04-22
Create workflow to publish project on PyPI when released

rennerocha:
When Travis was removed, the steps to deploy new versions on PyPI was removed too. This PR includes a new workflow that will publish to PyPI every time a new release is made.

@Gallaecio , I don't have access, but probably we already have these secrets configured for this repository, am I right?

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,27 @@
+name: Upload Python Package
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python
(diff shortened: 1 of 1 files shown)


Gallaecio:
> @Gallaecio , I don't have access, but probably we already have these secrets configured for this repository, am I right?

I’ve added it now :)

Gallaecio (approved):
@noviluni Just for your information: with this change, the package is deployed to PyPI when a release entry is created in GitHub, in https://github.com/scrapinghub/dateparser/releases, not when a version tag is pushed.

noviluni:
Nice! I think it's time to release a new version, so I will merge it just before doing it and will see if it works :)

thanks both @Gallaecio @rennerocha

Gallaecio (approved):

noviluni (approved):

Issue #916, opened 2021-04-28
Handling whitespace when matching relative dates


kss149:
This is for version `dateparser==1.0.0`.
I've noticed that for some of the dates I tried to parse I need to add spaces to make parsing work and for some I need to remove them.
It seems a bit random and I am guessing it depends on how the string was added to date_translation_data.
That doesn't seem very flexible or reliable, and I wonder if it would make sense to either remove all whitespace from both input data and date_translation_data before matching, or make them optional.

**Examples**

- 2 minutes ago in Mandarin

```
>>> dateparser.parse("2分鐘前")
>>> dateparser.parse("2 分鐘前")                  -> works only if you add space between the number and symbol
datetime.datetime(2021, 4, 28, 20, 45, 19, 266428)
```

- 2 months ago in Thai
```
>>> dateparser.parse("2 เดือน ที่ผ่านมา")            -> doesn't work with space between years and ago
>>> dateparser.parse("2 เดือนที่ผ่านมา")             -> works when you remove space between years and ago
datetime.datetime(2021, 2, 28, 22, 3, 40, 264496)
>>> dateparser.parse("2เดือนที่ผ่านมา")              -> doesn't work when you remove all spaces
```

Issue #913, opened 2021-04-28
Parsing relative dates error with Taiwanese Mandarin


kss149:
I have trouble parsing the following string "2分鐘前". It should translate to "2 minutes ago" in Mandarin (checked with DeepL, GoogleTranslate, etc), but parsing returns None.
```
>>> dateparser.parse("2分鐘前")
>>> dateparser.parse("2 分鐘前")
datetime.datetime(2021, 4, 28, 20, 45, 19, 266428)
```
I added a space in between the number and the symbols and it works, but ideally I would like it to work without adding it.
I think it may be because I got it from a Taiwanese version of a website, but I am no expert.

I translated back to Mandarin "2 minutes ago" and got "2分钟前" (notice the symbol in the middle is different), which parses fine with and without a space, but it'd be great if the ?Taiwanese? version will work as well.
```
>>> dateparser.parse("2分钟前")
datetime.datetime(2021, 4, 28, 20, 45, 32, 129967)
>>> dateparser.parse("2 分钟前")
datetime.datetime(2021, 4, 28, 20, 45, 35, 659334)
```

PR #917, opened 2021-04-29
Fix spaces in langs with no word spacing

noviluni:
Some regexes come from the CLDR corpus and contain spaces that should be optional. I modified the `write_complete_data.py` script to override them with optional spaces in those languages where the spaces are optional (indicated with `no_word_spacing`).

Apart from that, I added the `no_word_spacing` flag to Thai, as the spaces are optional, and in the case of Thai I also added a new word for "ago".

Finally, I added some tests from the examples provided in the issues below.

* fixes: https://github.com/scrapinghub/dateparser/issues/915
* fixes: https://github.com/scrapinghub/dateparser/issues/916
* fixes: https://github.com/scrapinghub/dateparser/issues/913


dateparser_scripts/write_complete_data.py
```diff
@@ -24,23 +24,27 @@
...
 
 
-def _modify_relative_data(relative_data):
+def _modify_relative_data(relative_data, replace_spaces=False):
     modified_relative_data = OrderedDict()
     for key, value in relative_data.items():
         for i, string in enumerate(value):
             string = RELATIVE_PATTERN.sub(r'(\\d+)', string)
+            if replace_spaces:
+                string = re.sub(r'\s+', '\\s?', string)
```
Gallaecio, line 33: So that we still allow more than 1 space character:

```suggestion
                string = re.sub(r'\s+', '\\s*', string)
```

Gallaecio (commented):
The change looks good to me, but the test error seems legit. I wonder if it has anything to do with `no_word_spacing`.

dateparser_scripts/write_complete_data.py
noviluni: hmmmm... I thought it when coding, but the string is "sanitized" before, and multiple spaces are converted to just one, so the string shouldn't come with more than one space: https://github.com/scrapinghub/dateparser/blob/21c629f067a40c2becb19c7e88ec8170a8da3fcb/dateparser/date.py#L44-L48

You can try this in this branch and you will see that it's working:

```python
>>> dateparser.parse('2 分鐘前')
datetime.datetime(2021, 4, 29, 11, 9, 25, 525814)

>>> dateparser.parse('2  分鐘前')
datetime.datetime(2021, 4, 29, 11, 9, 27, 740355)

>>> dateparser.parse('2      分鐘前')
datetime.datetime(2021, 4, 29, 11, 9, 30, 363743)
```

I could change it, but maybe the intention is not so clear. What do you think?

noviluni:
Yes... the Thai test failing in  `search_dates` seems legit... I will check it.

PR #922, opened 2021-05-13
Consider RETURN_TIME_AS_PERIOD for timestamp times

### LINKED ISSUE #920: consider RETURN_TIME_AS_PERIOD for datetimes parsed as seconds since the epoch
The following:

```python
import dateparser
DATE_PARSER = dateparser.date.DateDataParser(
    languages = ['en'],
    settings = {
        'RETURN_TIME_AS_PERIOD': True,
        'TIMEZONE': 'UTC',
        'TO_TIMEZONE': 'UTC'
    }
)
DATE_PARSER.get_date_data('1620145800')
```

Results in: `DateData(date_obj=datetime.datetime(2021, 5, 4, 12, 30), period='day', locale='en')`

I believe the period should be `time` here.

onlynone:
This should fix #920

### DIFF
--- a/dateparser/date.py
+++ b/dateparser/date.py
@@ -187,7 +187,7 @@ class _DateLocaleParser:
     def _try_timestamp(self):
         return DateData(
             date_obj=get_date_from_timestamp(self.date_string, self._settings),
-            period='day',
+            period='time' if self._settings.RETURN_TIME_AS_PERIOD else 'day',
         )
 
     def _try_freshness_parser(self):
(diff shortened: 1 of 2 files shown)


noviluni:
Hi @onlynone! Thanks for opening this PR.

This definitely should work. Could you add a test that covers this change?

noviluni:
I think it's not necessary to add the 4 examples to the test, but it's ok for me.

noviluni (approved):

noviluni:
cc: @Gallaecio

Gallaecio (approved):

PR #954, opened 2021-07-27
test coverage for how unix timestamps are parsed

onlynone:


tests/test_date.py
```diff
@@ -710,13 +713,74 @@ def then_date_object_is_invalid(self):
...
+            ['EST5EDT4', 'UTC0', 'PST8PDT7', None],
+        )
+    )
+    def test_timestamp_with_different_timestr(self, timestamp, timezone, to_timezone, tzstr):
+        settings = {
+            'RETURN_AS_TIMEZONE_AWARE': True,
+            'TIMEZONE': timezone,
+        }
+
+        # is TO_TIMEZONE supposed to be allowed to be False, or None ???
```
Gallaecio, line 769: It’s `None` by default, so that value should be allowed, but `False` would be bad user input, I believe: https://dateparser.readthedocs.io/en/latest/settings.html#timezone-related-configurations

In order to test both passing `None` specifically and not defining the setting at all, you could define `DONT_SET = object()` in this module, add it to the `to_timezone` parameter list above, and use it in the `if` here instead of `None`.

onlynone: I got validation errors when I tried both `None` and `False`.

onlynone: The code seems to indicate that `False` is the default value? https://github.com/scrapinghub/dateparser/blob/21c629f067a40c2becb19c7e88ec8170a8da3fcb/dateparser_data/settings.py#L15 But I don't know exactly how `dateparser_data.settings` is used.

onlynone:
By the way, when I was adding tests, I tried to follow the pattern of the other tests in `TestTimestampParser` by calling `date.get_date_from_timestamp` directly, and passing in a custom value for `settings`. I also saw a pattern for creating a new `Settings` object, in another test in `tests/test_date.py` of doing something like:

```
        settings_mod = copy(settings)
        settings_mod.RETURN_AS_TIMEZONE_AWARE = True
        settings_mod.TIMEZONE = timezone
        settings_mod.TO_TIMEZONE = to_timezone

        date.get_date_from_timestamp(timestamp, settings_mod)
```

But, when I do that, it seems like I end up modifying the global `settings` object imported from `dateparser.conf` and that caused other tests to fail because they weren't expecting `RETURN_AS_TIMEZONE_AWARE` to be `True`. Is `dateparser.conf.Settings` a singleton? Should the other test that tries to do a `copy` be fixed? Should there be a more standard way to create a custom/one-off settings object for use in tests?

tests/test_date.py
Gallaecio: OK, given your results and https://github.com/scrapinghub/dateparser/blob/21c629f067a40c2becb19c7e88ec8170a8da3fcb/dateparser/conf.py#L146-L150 I guess your current test code covers all it needs to cover.

onlynone:
@Gallaecio Is this good to merge?

dateparser/date.py
```diff
@@ -120,11 +122,18 @@ def get_date_from_timestamp(date_string, settings, negative=False):
          match = RE_SEARCH_TIMESTAMP.search(date_string)
 
     if match:
+        if settings is not None and settings.TIMEZONE is not None and 'local' not in settings.TIMEZONE.lower():
```
Gallaecio, line 125:
```suggestion
        if settings is not None and settings.TIMEZONE is not None and 'local' not in settings.TIMEZONE.lower():
```

I am unsure about `'local' not in settings.TIMEZONE.lower()`. According to the documentation, the local timezone is used by default (i.e. when the value is None), and otherwise a valid timezone has to be specified. So this statement does not seem necessary, and it can be troublesome (imagine a city containing the letters “local” in it).

onlynone: > I am unsure about 'local' not in settings.TIMEZONE.lower(). According to the documentation, the local timezone is used by default (i.e. when the value is None), and otherwise a valid timezone has to be specified. 

RIght, when `settings` is `None` (maybe this is impossible :shrug:), or the `TIMEZONE` is `None` or `'local'` is in the timezone, then it'll use `get_localzone()`. It's only when `settings.TIMEZONE` exists and does not contain `'local'`, that the code will call `get_timezone_from_tz_string(settings.TIMEZONE)`. Would it be better to flip the boolean logic and use:

```python
        if settings is None or settings.TIMEZONE is None or 'local' in settings.TIMEZONE.lower():
            local_timezone = get_localzone()
        else:
            local_timezone = get_timezone_from_tz_string(settings.TIMEZONE)
```

> So this statement does not seem necessary, and it can be troublesome (imagine a city containing the letters “local” in it).

I got that idiom from other parts of the code that check if the timezone is local:

```
dateparser/date_parser.py:        _settings_tz = settings.TIMEZONE.lower()
dateparser/date_parser.py:            if 'local' not in _settings_tz:
dateparser/date_parser.py:            if 'local' in _settings_tz:
dateparser/freshness_date_parser.py:        _settings_tz = settings.TIMEZONE.lower()
dateparser/freshness_date_parser.py:            if 'local' not in _settings_tz:
dateparser/freshness_date_parser.py:            if 'local' in _settings_tz:
dateparser/freshness_date_parser.py:            if 'local' not in _settings_tz:
dateparser/utils/__init__.py:    if 'local' in settings.TIMEZONE.lower():
```

Is there a better one to use?

Gallaecio: Hmm…

Well, if it is done elsewhere, I see no reason to block your pull request for that reason. Also, it does not seem like any [current timezone](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones) contains the work “local” in it.

Gallaecio (approved):
Great work!

gutsytechster:
TBH, I am not able to understand the issue here. Maybe an example may help.

But if we are applying the timezone and stripping it off, to eventually apply the timezone from settings through `apply_timezone_from_settings `, how does the behavior change?

Also the statement in the initial explanation
> The `dateparser` code will try to interpret that `datetime` object as if it was in the `settings.TIMEZONE` timezone, which is incorrect

Where do we see that behavior, since the `apply_tiemzone_from_settings` would apply the timezone from the settings, I believe?

onlynone:
The examples in the added tests should show the issue. You could try running those tests on the main line code without my changes.

> But if we are applying the timezone and stripping it off, to eventually apply the timezone from settings through apply_timezone_from_settings , how does the behavior change?

The problem is that there's not just one time zone that's being stripped and reapplied. The date is initially constructed with the time zone of the system (regardless of any dateparser settings), but then *that* time zone is not kept on the datetime. Then later in the code, the datetime is interpreted as if it was in dateparser's settings.TIMEZONE. If the system's time zone happens to be the same as the one used for dateparser's settings, then I think everything is fine. But if my system is set to 'America/New_York', and I set the dateparser settings time zone to 'UTC' and ask it to parse '1661996156', I believe dateparser will currently give me 'August 31, 2022 9:35:56 PM UTC' when it should give me 'September 1, 2022 1:35:56 AM UTC'.

Sorry if there are any errors in the above comment, I'm on mobile and it's also been a while since I first made this branch. Please test it the examples and see if what I'm saying makes sense.

onlynone:
> Where do we see that behavior, since the apply_tiemzone_from_settings would apply the timezone from the settings, I believe?

Yes, but with the new code, the naive datetime object is created in the same time zone as settings.TIMEZONE. so that later, when it's interpreted as if it was in settings.TIMEZONE, that's now a correct assumption. With the old code, the naive datetime is created relative to the system's time zone.

onlynone:
Here's an example. With the current code:

```
>>> dateparser.parse('1661996156', settings={'TIMEZONE': 'UTC', 'TO_TIMEZONE': 'UTC', 'RETURN_AS_TIMEZONE_AWARE': True}).timestamp()
1661981756.0
```

The returned unix timestamp doesn't match the original timestamp. And with my branch:

```
>>> dateparser.parse('1661996156', settings={'TIMEZONE': 'UTC', 'TO_TIMEZONE': 'UTC', 'RETURN_AS_TIMEZONE_AWARE': True}).timestamp()
1661996156.0
```

It does match.

gutsytechster:
Thanks @onlynone!

It's clear now. I understand where the difference comes from. 

As far as I understood, previous to, when the date goes to the `apply_timezone_from_settings`, it got converted into a date object with a local timezone(without timezone included), and that date is further used for applying timezone from the mentioned method.

dateparser/date.py
```diff
@@ -120,11 +122,18 @@ def get_date_from_timestamp(date_string, settings, negative=False):
...
+            local_timezone = get_timezone_from_tz_string(settings.TIMEZONE)
+        else:
+            local_timezone = get_localzone()
```
gutsytechster, line 128: Will it make more sense to use the variable as `timezone` or `current_timezone` instead of `local_timezone` as the timezone coming from `get_timezone_from_tz_string(settings.TIMEZONE)` may not be the local timezone?

dateparser/date.py
gutsytechster, line 125:
```suggestion
        if (
            settings is not None 
            and settings.TIMEZONE is not None 
            and 'local' not in settings.TIMEZONE.lower()
        ):
```

This way it seems more readable. I had trouble initially understanding the whole statement. So I believe this can be restructured a bit.

dateparser/date.py
onlynone: yeah that makes sense

dateparser/date.py
onlynone: how about flipping the boolean logic, like I mentioned in this comment: https://github.com/scrapinghub/dateparser/pull/954#discussion_r958668351 ? I feel like that might be clearer. I could change the formatting and logic.

onlynone: The latest push has the change in formatting you suggested, clarifying comments, and flips the boolean logic so that I think it's clearer.

gutsytechster (approved):

ghazel:
Using an old version of `tzlocal`, 2.1 made `dateparser` return Local Mean Time instead of the correct timezone as the result of this change. After they switched away from `pytz` to `zoneinfo` that problem went away. You might want to bump your minimum `tzlocal` version dependency.

PR #955, opened 2021-07-29
Fixing: flake8 E402 and windows setup.py not working

gavishpoddar:
This PR Fixes,

- Fixing E402 module-level import, not at top of file in dateparser/docs/conf.py
- Fixing UnicodeDecodeError: 'charmap' codec can't decode byte 0x8f in position 1442: character maps to <undefined> in Windows

Note: This change makes it possible to run `pip install . `. But tox is not entirely fixed.

Please suggest.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -13,6 +13,7 @@
 import sys
 import os
 
+import dateparser
 
 # Get the project root dir, which is the parent dir of this
 cwd = os.getcwd()
@@ -22,8 +23,6 @@ project_root = os.path.dirname(cwd)
 # This lets us ensure that the source package is imported, and that its
 # version is used.
 sys.path.insert(0, project_root)
-
-import dateparser
 
 # -- General configuration ---------------------------------------------
 
--- a/setup.py
+++ b/setup.py
@@ -3,8 +3,8 @@ from setuptools import setup, find_packages
 
 __version__ = re.search(r"__version__.*\s*=\s*[']([^']+)[']", open('dateparser/__init__.py').read()).group(1)
 
-introduction = re.sub(r':members:.+|..\sautomodule::.+|:class:|:func:|:ref:', '', open('docs/introduction.rst').read())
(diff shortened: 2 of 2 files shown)


Gallaecio (approved):
The fixes seem safe to me.

docs/conf.py
```diff
@@ -23,8 +24,6 @@
 # version is used.
 sys.path.insert(0, project_root)
 
-import dateparser
-
```
noviluni, line 27: From what I read in the comments:

> Insert the project root dir as the first element in the PYTHONPATH.
> This lets us ensure that the source package is imported, and that its
> version is used.

 I think that this import is here intentionally. I'm not sure if moving it from here would have any adversarial consequence. Wouldn't be better to add change it to...
 
  ```python3
  import dateparser  # noqa: E402
  ```

just in case? @Gallaecio

noviluni (changes requested):
Could you remove the ` dateparser/docs/conf.py E402` line from the `pytest.ini`?

thanks!

gavishpoddar:
Hi, I have made the changes

docs/conf.py
noviluni: BTW, I've seen that in `parsel` has the import above, so maybe I'm worrying unnecessarily:

https://github.com/scrapy/parsel/blob/master/docs/conf.py

Gallaecio: I had not thought of that. Indeed, having it after will prevent a system-wide-installed version of dateparser being used instead.

It should not be a problem when using `tox` though, I believe, since `tox` installs the project into its virtual environment every time.

noviluni: I prefer to keep it where it was to avoid needing to modify the comments, etc.

@gavishpoddar could you move it again where it was and add the "noqa"?:

`import dateparser  # noqa: E402`

Thanks!

gavishpoddar: Done 👍🏻

Gallaecio (approved):

PR #957, opened 2021-08-07
Homogenize flake8 line length

noviluni:
We set a [max length of 119](https://github.com/scrapinghub/dateparser/pull/735) characters and the `flake8` on the pipelines was using that value from the `pytest.ini` (`flake8-max-line-length`). However, there were some references to 100 characters. I updated the docs (`CONTRIBUTING.rst`) and the `setup.cfg`.

On the other hand, when using the `flake8` tool directly (instead of using it along with `tox`), the `flake8` file wasn't working. I checked the [flake8 documentation](https://flake8.pycqa.org/en/latest/user/configuration.html), and that file is not accepted, in any case, it should start with a period (`.flake8`):

> Flake8 supports storing its configuration in the following places:
>
>   * Your top-level user directory
>   * In your project in one of setup.cfg, tox.ini, or .flake8.

So I removed it. Then I also realized that it would be a good idea to have a "single point of truth", and as `pytest-flake8` will take the values from the `setup.cfg`, I removed the `flake8-max-line-length` from the `pytest.ini`.

Finally, and following with the last idea, I moved the `flake8` excusion rules to the `setup.cfg`.


CONTRIBUTING.rst
```diff
@@ -89,7 +89,7 @@ tests, including testing other Python versions with tox::
 
     $ tox
 
-   To get flake8 and tox, just pip install them into your virtualenv. (Note that we use ``max-line-length = 100`` for flake8, this is configured in ``setup.cfg`` file.)
+   To get tox, just pip install them into your virtualenv. (Note that we use ``max-line-length = 119`` for flake8, this is configured in ``setup.cfg`` file.)
```
Gallaecio, line 92: What about not mentioning flake8 or any configuration file at all at this point?

```suggestion
   To get ``tox``, just ``pip install`` it into your virtualenv. In addition to tests, ``tox`` checks for code style and maximum line length (119 characters).
```

Gallaecio (approved):

PR #966, opened 2021-08-18
Fix github actions (language pack fr)

noviluni:


lopuhin (approved):
Makes sense, thanks @noviluni :+1:

lopuhin:
The build still failed for py35 but it's a different reason: https://github.com/scrapinghub/dateparser/runs/3370394856?check_suite_focus=true

Btw I wonder if it would be fine to drop support for python 3.5? I think many popular libraries dropped it a long time ago.

noviluni:
> The build still failed for py35

Hi @lopuhin, that error is really weird, it says that it couldn't find the convertdate package. I re-runed the pipelines and it worked. Seems more like an error in PyPI, not something related to Python 3.5.

> it would be fine to drop support for python 3.5?

Yeah, I was thinking about deprecating Python 3.5, but I thought it could be better to wait one version more, as this new version (that I plan to release right after the GSoC) will include a lot of fixes and all is compatible with Python 3.5, so it would be nice to keep the support at least one more version. For me it's totally fine to update to Python 3.6 the next of the next version :)

Issue #973, opened 2021-08-27
Python segfaults when parsing specific date through pytest


La0:
Got a weird bug on a Python project with unit tests today.

Some specific dates trigger a Python segfault (not an exception) when running `dateparser.parse("14 july 1789")` in a unit test run by [pytest](https://docs.pytest.org/en/6.2.x/)

I was able to reproduce in an isolated Docker container (here is the [full gist](https://gist.github.com/La0/b47ffd259809be0dc08cae1315ccf2b2) with source code)

Running these two unit tests will trigger a segfault on the second one:

```python
import dateparser


def test_ok():
    dateparser.parse('1 July 1789')


def test_crash():
    dateparser.parse('14 July 1789')
```

Here is the stack trace:

```
============================= test session starts ==============================
platform linux -- Python 3.8.11, pytest-6.2.4, py-1.10.0, pluggy-0.13.1 -- /usr/local/bin/python
cachedir: .pytest_cache
rootdir: /
collecting ... collected 2 items

test.py::test_ok PASSED                                                  [ 50%]
test.py::test_crash Fatal Python error: Segmentation fault

Current thread 0x00007fb8ba7a0740 (most recent call first):
  File "/usr/local/lib/python3.8/site-packages/dateparser/languages/dictionary.py", line 136 in split
  File "/usr/local/lib/python3.8/site-packages/dateparser/languages/locale.py", line 72 in is_applicable
  File "/usr/local/lib/python3.8/site-packages/dateparser/date.py", line 473 in _is_applicable_locale
  File "/usr/local/lib/python3.8/site-packages/dateparser/date.py", line 469 in _get_applicable_locales
  File "/usr/local/lib/python3.8/site-packages/dateparser/date.py", line 420 in get_date_data
  File "/usr/local/lib/python3.8/site-packages/dateparser/__init__.py", line 54 in parse
  File "/usr/local/lib/python3.8/site-packages/dateparser/conf.py", line 89 in wrapper
  File "/test.py", line 9 in test_crash
  File "/usr/local/lib/python3.8/site-packages/_pytest/python.py", line 183 in pytest_pyfunc_call
  File "/usr/local/lib/python3.8/site-packages/pluggy/callers.py", line 187 in _multicall
  File "/usr/local/lib/python3.8/site-packages/pluggy/manager.py", line 84 in <lambda>
  File "/usr/local/lib/python3.8/site-packages/pluggy/manager.py", line 93 in _hookexec
[...]

Issue #972, opened 2021-08-27
New Regex Release: regex==2021.8.27 breaks dateparser (SegFault)


AndyXan:
Hello Dateparser Team, 

Long story short - did take some debugging today. But the newly bread regex release breaks dateparser utility. 

`virtualenv -p /usr/bin/python3 ~/.virtualenvs/fresh`
`workon fresh`
`pip install dateparser`

```
pip freeze
backports.zoneinfo==0.2.1
dateparser==1.0.0
python-dateutil==2.8.2
pytz==2021.1
regex==2021.8.27
six==1.16.0
tzlocal==3.0
```

```
Type "help", "copyright", "credits" or "license" for more information.
import dateparser                                                       
dateparser.parse("2020-09-10 14:32:00.100").replace(tzinfo=None) 
Segmentation fault (core dumped)
```

`pip install regex==2020.10.28`

```
import dateparser                                                       
dateparser.parse("2020-09-10 14:32:00.100").replace(tzinfo=None) 
datetime.datetime(2020, 9, 10, 14, 32, 0, 100000)
```

So quickfix: pin regex-version in dateparser dependencies. Long term.. I have no idea what causes this :)

athossampayo:
@noviluni This problem was affecting some deployments were I work, so I made a quick hotfix doing exactly what you said, excluding the problematic version on setup.py. 

Mind to check the PR? Any suggestions are very well received as this is my first PR on regex.

PR #974, opened 2021-08-27
fix(regex-version): removed version 2021.8.27 that breaks dateparser with SegFault

athossampayo:
fix(regex-version): removed version 2021.8.27 that breaks dateparser with SegFault

Related to  #972

Tests result:
image

-----

Fixes: #972. 
Fixes: #973

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@ setup(
         'python-dateutil',
         'pytz',
         # https://bitbucket.org/mrabarnett/mrab-regex/issues/314/import-error-no-module-named
-        'regex !=2019.02.19',
+        'regex !=2019.02.19,!=2021.8.27',
         'tzlocal',
     ],
     extras_require={


noviluni (changes requested):
Hi @athossampayo, thank you for opening this PR. I think the correct way to do it is separating the versions with commas, as set in [PEP-440](https://www.python.org/dev/peps/pep-0440/#version-specifiers): `'regex !=2019.02.19,!=2021.8.27',`

Could you try with these semantics?

Thanks! :)

athossampayo:
Hello @noviluni, thanks for the tip! Did not find a pattern to follow on this case. 

Already updated the code.

If there's anything more to fix, just say.

Thanks!

noviluni (approved):
:muscle:

lopuhin (approved):
Thanks @athossampayo :+1:

PR #997, opened 2021-10-07
Allow to use_given_order for languages too

whalebot-helmsman:


dateparser/date.py
```diff
@@ -358,8 +358,8 @@ def __init__(self, languages=None, locales=None, region=None, try_previous_local
             raise TypeError("use_given_order argument must be a boolean (%r given)"
                             % type(use_given_order))
 
-        if not locales and use_given_order:
-            raise ValueError("locales must be given if use_given_order is True")
+        if not locales and not languages and use_given_order:
+            raise ValueError("locales or locales must be given if use_given_order is True")
```
Gallaecio, line 362:
```suggestion
            raise ValueError("locales or languages must be given if use_given_order is True")
```

Gallaecio:
Is that all that needs changing? (i.e. was the feature there but this exception was getting in the way?)

Could you add a test?

Gallaecio:
> There is a test of loading languages with `use_given_order`

But it was not failing before. We need a test that would have caught the issue before your change, i.e. a test that fails if we revert your change, and passes after we apply it. And I don’t mean https://github.com/scrapinghub/dateparser/pull/997/files#diff-fd61b87100377a706498644e5ff42b8baecf18918169f6d4082af10d5a053477R658.

PR #1022, opened 2021-11-25
Fix directories creation error

### LINKED ISSUE #1021: create_data_model_home() fails without ~/.cache
`dateparser_cli.utils.create_data_model_home()` calls `os.mkdir('~/.cache/dateparser_models`)` which fails when ~/.cache doesn't exist (see https://stackoverflow.com/questions/600268/mkdir-p-functionality-in-python for suggestions).

It would also be nice to use `xdg.xdg_cache_home` instead of hardcoding `.cache`, though that adds a dependency.

Laerte:
This fix #1021

### DIFF
--- a/dateparser_cli/utils.py
+++ b/dateparser_cli/utils.py
@@ -25,7 +25,7 @@ dateparser_model_home = os.path.expanduser(
 
 def create_data_model_home():
     if not os.path.isdir(dateparser_model_home):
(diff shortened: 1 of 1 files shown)


Laerte:
Hey @lopuhin can you review this? Thank you.

lopuhin (approved):
Looks good, thanks for the fix @Laerte

PR #1060, opened 2022-06-06
Add support for negative timestamp

### LINKED ISSUE #546: Timestamp in milliseconds with minus (-) - parsing failes/returns None
Hi,

**Version**:
dateparser==0.7.1

parsing a timestamp in milliseconds with an minus in front of it, returns None.

**Example**:
>> dateparser.parse("-186454800000")
>> None

The correct date would be:
Feb 03 1964 23:00:00

If I remove the minus or pass any other positive timestamp in milliseconds, everything works as expected, so it´s definetly the minus. 
I obviously can´t remove it, because it will return a wrong datetime value.

Converter for timestamp in milliseconds:
https://currentmillis.com/

Thanks and best regards

gutsytechster:
This closes #546

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -144,7 +144,7 @@ Language Detection
 Default Languages
 +++++++++++++++++
 
-``DEFAULT_LANGUAGES``: It is a ``list`` of language codes in ISO 639 that will be used as default 
+``DEFAULT_LANGUAGES``: It is a ``list`` of language codes in ISO 639 that will be used as default
 languages for parsing when language detection fails. eg. ["en", "fr"]:
 
     >>> from dateparser import parse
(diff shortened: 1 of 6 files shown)


docs/settings.rst
```diff
@@ -181,6 +181,9 @@ The following parsers exist:
     followed by additional digits or a period (``.``), those first 10 digits
     are interpreted as `Unix time <https://en.wikipedia.org/wiki/Unix_time>`_.
 
+-    ``'negative-timestamp'``: It is similar to ``'timestamp'``, but parses the
+    negative timestamp.
```
Gallaecio, line 185:
```suggestion
-   ``'negative-timestamp'``: ``'timestamp'`` for negative timestamps. For 
    example, parses ``-186454800000`` as ``1964-02-03T23:00:00``.
```

tests/test_date.py
```diff
@@ -726,6 +726,34 @@ def test_timestamp_in_microseconds(self):
             datetime.fromtimestamp(1570308760).replace(microsecond=263111)
         )
 
+    @parameterized.expand([
+        param(
+            input_timestamp='-1570308760',
+            negative_allowed=True,
```
Gallaecio, line 732: This seems to indicate that negative and positive are both parsed. Maybe:

```suggestion
            negative=True,
```

Gallaecio (approved):

Gallaecio (approved):

Issue #1013, opened 2021-10-28
Python 3.9 PytzUsageWarning


kkoehncke:
When upgrading to Python 3.9, getting the following warning: 
```
Python 3.9.5 (default, Jun  4 2021, 12:28:51)
[GCC 7.5.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import dateparser
>>> dateparser.parse("2021-08-02")
/opt/conda/lib/python3.9/site-packages/dateparser/date_parser.py:35: PytzUsageWarning: The localize method is no longer necessary, as this time zone supports the fold attribute (PEP 495). For more details on migrating to a PEP 495-compliant implementation, see https://pytz-deprecation-shim.readthedocs.io/en/latest/migration.html
  date_obj = stz.localize(date_obj)
datetime.datetime(2021, 8, 2, 0, 0)
```
Is there an anticipated fix for this or a way to suppress this?

Thanks!

viseshrp:
is there a way to silence these warnings temporarily?

FinnWoelm:
Hi @viseshrp,

I was able to temporarily silence these warnings with:

```python
import warnings
import dateparser

# Ignore dateparser warnings regarding pytz
warnings.filterwarnings(
    "ignore",
    message="The localize method is no longer necessary, as this time zone supports the fold attribute",
)

# Your code ...
```

SO thread on filtering Python warnings: https://stackoverflow.com/a/56142903/6451879

----

For `pytest`, you can modify `pytest.ini` like this:

```
[pytest]
filterwarnings = 
    ignore:The localize method is no longer necessary, as this time zone supports the fold attribute
```

viseshrp:
> Hi @viseshrp,
> 
> I was able to temporarily silence these warnings with:
> 
> ```python
> import warnings
> import dateparser
> 
> # Ignore dateparser warnings regarding pytz
> warnings.filterwarnings(
>     "ignore",
>     message="The localize method is no longer necessary, as this time zone supports the fold attribute",
> )
> 
> # Your code ...
> ```
> 
> SO thread on filtering Python warnings: https://stackoverflow.com/a/56142903/6451879
> 
> For `pytest`, you can modify `pytest.ini` like this:
> 
> ```
> [pytest]
> filterwarnings = 
>     ignore:The localize method is no longer necessary, as this time zone supports the fold attribute
> ```

Thanks!

ElijahSink:
[...]

PR #1062, opened 2022-06-10
Fix PytzUsageWarning for Python versions >= 3.6

ismaildawoodjee:
This fix makes the `localize` method to be used only for Python versions that are lower than 3.6, so that the warning doesn't appear for versions 3.6+. This also allows compatibility with Python 3.5.

Referencing Issue #1013 and implemented suggestions by @[DavidMStraub](https://github.com/DavidMStraub) and @[bsekiewicz](https://github.com/bsekiewicz) 

- [ ] Codecov report pending
- [x] Test cases passed with `tox` and `tox -e py`
image

### DIFF
--- a/dateparser/date_parser.py
+++ b/dateparser/date_parser.py
@@ -1,3 +1,5 @@
+import sys
+
 from tzlocal import get_localzone
 
 from .timezone_parser import pop_tz_offset_from_string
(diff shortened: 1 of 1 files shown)


gutsytechster (approved):

Gallaecio:
Thanks!

Gallaecio:
There is no dateparser release with the fix yet. If you install from the main Git branch, you should not get it.

andydoc:
utils/__init__.py has the same problem at line 129
it is also possible that line 27 in date_parser.py needs similar amendment, but my current use case dose not get there

Pugio:
[freshness_date_parser.py:76](https://github.com/scrapinghub/dateparser/blob/master/dateparser/freshness_date_parser.py#L76) also needs the same kind of change.

PR #1064, opened 2022-07-12
Slovak suplementary data

janik-martin:
The main aim is to cover shorthand relative dates when relating to "one". Similar to English, where 'a' is used, Slovak leaves out the slot entirely. So relative date 'o hodinu' means 'in an hour', which was not working. But 'o 1 hodinu' was working, this is rarely used in Slovak however.
Also added forms of words used with count 2,3,4 - e.g. in 2 weeks -> o 2 týždne, but in 5 weeks -> o 5 týždňov. This is sometimes referred as "one, few, many" grouping, where English only uses one and many.


janik-martin:
Hi I merged the latest changes and updated mine according to them. It would be great if somebody take take a look at this PR. Thanks.

Gallaecio (approved):

Issue #1052, opened 2022-03-28
dateparser-1.1.1 is incompatible with regex-2022.3.15


8lurry:
Upgraded pip packages;

```
(env) blurry:>$ pip freeze | grep dateparser
dateparser==1.1.1
(env) blurry:>$ pip freeze | grep regex
regex==2022.3.15
```

dateparser.parse raises the following exception:

```
Python 3.8.10 (default, Nov 26 2021, 20:14:08) 
Type 'copyright', 'credits' or 'license' for more information
IPython 8.1.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import dateparser

In [2]: dateparser.parse('01.JUN.1968')
---------------------------------------------------------------------------
error                                     Traceback (most recent call last)
Input In [2], in <cell line: 1>()
----> 1 dateparser.parse('01.JUN.1968')

File ~/lino/env/lib/python3.8/site-packages/dateparser/conf.py:92, in apply_settings.<locals>.wrapper(*args, **kwargs)
     89 if not isinstance(kwargs['settings'], Settings):
     90     raise TypeError("settings can only be either dict or instance of Settings class")
---> 92 return f(*args, **kwargs)

File ~/lino/env/lib/python3.8/site-packages/dateparser/__init__.py:61, in parse(date_string, date_formats, languages, locales, region, settings, detect_languages_function)
     57 if languages or locales or region or detect_languages_function or not settings._default:
     58     parser = DateDataParser(languages=languages, locales=locales,
     59                             region=region, settings=settings, detect_languages_function=detect_languages_function)
---> 61 data = parser.get_date_data(date_string, date_formats)
     63 if data:
     64     return data['date_obj']

File ~/lino/env/lib/python3.8/site-packages/dateparser/date.py:428, in DateDataParser.get_date_data(self, date_string, date_formats)
    425 date_string = sanitize_date(date_string)
    427 for locale in self._get_applicable_locales(date_string):
--> 428     parsed_date = _DateLocaleParser.parse(
    429         locale, date_string, date_formats, settings=self._settings)
    430     if parsed_date:
    431         parsed_date['locale'] = locale.shortname

File ~/lino/env/lib/python3.8/site-packages/dateparser/date.py:178, in _DateLocaleParser.parse(cls, locale, date_string, date_formats, settings)
    175 @classmethod
    176 def parse(cls, locale, date_string, date_formats=None, settings=None):
[...]

Issue #1045, opened 2022-03-16
Can't parse anything


Etirf:
Hello everyone,

Tried parsing under python 3.7.5 and 3.9

`dateparser.parse('12/12/12')`

It also gives the same output for any "valid" input shown in the doc:
```
dateparser.parse('Fri, 12 Dec 2014 10:55:50')
dateparser.parse('22 Décembre 2010', date_formats=['%d %B %Y'])
...
```

Here's the error:

```

---------------------------------------------------------------------------
error                                     Traceback (most recent call last)
Input In [46], in <cell line: 1>()
----> 1 dateparser.parse("12/12/12")

File c:\users\strey\appdata\local\programs\python\python39\lib\site-packages\dateparser\conf.py:92, in apply_settings.<locals>.wrapper(*args, **kwargs)
     89 if not isinstance(kwargs['settings'], Settings):
     90     raise TypeError("settings can only be either dict or instance of Settings class")
---> 92 return f(*args, **kwargs)

File c:\users\strey\appdata\local\programs\python\python39\lib\site-packages\dateparser\__init__.py:61, in parse(date_string, date_formats, languages, locales, region, settings, detect_languages_function)
     57 if languages or locales or region or detect_languages_function or not settings._default:
     58     parser = DateDataParser(languages=languages, locales=locales,
     59                             region=region, settings=settings, detect_languages_function=detect_languages_function)
---> 61 data = parser.get_date_data(date_string, date_formats)
     63 if data:
     64     return data['date_obj']

File c:\users\strey\appdata\local\programs\python\python39\lib\site-packages\dateparser\date.py:428, in DateDataParser.get_date_data(self, date_string, date_formats)
    425 date_string = sanitize_date(date_string)
    427 for locale in self._get_applicable_locales(date_string):
--> 428     parsed_date = _DateLocaleParser.parse(
    429         locale, date_string, date_formats, settings=self._settings)
    430     if parsed_date:
    431         parsed_date['locale'] = locale.shortname

File c:\users\strey\appdata\local\programs\python\python39\lib\site-packages\dateparser\date.py:178, in _DateLocaleParser.parse(cls, locale, date_string, date_formats, settings)
    175 @classmethod
    176 def parse(cls, locale, date_string, date_formats=None, settings=None):
    177     instance = cls(locale, date_string, date_formats, settings)
--> 178     return instance._parse()

[...]

PR #1067, opened 2022-08-11
Try fixing regex incompatibility

goodspark:
Fixes https://github.com/scrapinghub/dateparser/issues/1052

Based on https://github.com/scrapinghub/dateparser/issues/1045#issuecomment-1129846022

### DIFF
--- a/dateparser/languages/locale.py
+++ b/dateparser/languages/locale.py
@@ -169,7 +169,7 @@ class Locale:
             if normalize:
                 value = list(map(normalize_unicode, value))
             pattern = '|'.join(sorted(value, key=len, reverse=True))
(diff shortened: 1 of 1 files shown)


Gallaecio:
Why is `DIGIT_GROUP_PATTERN` replaced by `pattern`? Isn’t that a change in behavior?

Gallaecio:
Makes sense.

However:
- `DIGIT_GROUP_PATTERN` does not seem used anymore, so we should remove it.
- `pattern.replace('(\\d+', '(?P<n>\\d+')` indeed sounds better. However, seeing how the old code did it without it, I am OK with not improving on this.

DJRHails:
Made those changes https://github.com/scrapinghub/dateparser/pull/1095 here :+1:

Issue #849, opened 2020-12-02
DateParser throws error when imported on embedded Python.


sudeepjd:
The dateparser library uses a exec_module on the _strptime. This causes the import to fail when using the the embedded version of python as the import happens via zipimport and zipimport does not have exec_module() so the following error gets thrown on import dateparser

```
D:\DateParser Test\python-3.9.0-embed-amd64>python
Python 3.9.0 (tags/v3.9.0:9cf6752, Oct  5 2020, 15:34:40) [MSC v.1927 64 bit (AM
D64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import dateparser
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "D:\DateParser Test\python-3.9.0-embed-amd64\lib\site-packages\dateparser\__init__.py", line 3, in <module>
    from .date import DateDataParser
  File "D:\DateParser Test\python-3.9.0-embed-amd64\lib\site-packages\dateparser\date.py", line 9, in <module>
    from dateparser.date_parser import date_parser
  File "D:\DateParser Test\python-3.9.0-embed-amd64\lib\site-packages\dateparser\date_parser.py", line 5, in <module>
    from .conf import apply_settings
  File "D:\DateParser Test\python-3.9.0-embed-amd64\lib\site-packages\dateparser\conf.py", line 5, in <module>
    from .parser import date_order_chart
  File "D:\DateParser Test\python-3.9.0-embed-amd64\lib\site-packages\dateparser\parser.py", line 12, in <module>
    from dateparser.utils.strptime import strptime
  File "D:\DateParser Test\python-3.9.0-embed-amd64\lib\site-packages\dateparser\utils\strptime.py", line 57, in <module>
    __strptime = patch_strptime()
  File "D:\DateParser Test\python-3.9.0-embed-amd64\lib\site-packages\dateparser\utils\strptime.py", line 28, in patch_strptime
    _strptime_spec.loader.exec_module(_strptime)
AttributeError: 'zipimporter' object has no attribute 'exec_module'
>>>
```

**Steps to Reproduce**

1. Download the python.zip file the link below 
https://www.python.org/ftp/python/3.9.0/python-3.9.0-embed-amd64.zip

2. Uncomment _import site_ in the **python39._pth**

3. Download pip from [https://bootstrap.pypa.io/get-pip.py](https://bootstrap.pypa.io/get-pip.py) and run **python get-pip.py** to install pip

[...]

PR #1069, opened 2022-08-24
Workaround for zipimporter not having exec_module before Python 3.10

georgevreilly:
Fixes https://github.com/scrapinghub/dateparser/issues/849

Tested with Python 3.8 and 3.9

### DIFF
--- a/dateparser/utils/strptime.py
+++ b/dateparser/utils/strptime.py
@@ -14,6 +14,14 @@ TIME_MATCHER = re.compile(
 
 MS_SEARCHER = re.compile(r'\.(?P<microsecond>[0-9]{1,6})')
 
+def _exec_module(spec, module):
+    if hasattr(spec.loader, "exec_module"):
+        spec.loader.exec_module(module)
+    else:
+        # This can happen before Python 3.10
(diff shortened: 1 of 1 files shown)


dateparser/utils/strptime.py
```diff
@@ -14,6 +14,14 @@
 
 MS_SEARCHER = re.compile(r'\.(?P<microsecond>[0-9]{1,6})')
 
+def _exec_module(spec, module):
+    if hasattr(spec.loader, "exec_module"):
+        spec.loader.exec_module(module)
+    else:
+        # This can happen before Python 3.10
+        # if spec.loader is a zipimporter and the Python runtime is in a zipfile
+        code = spec.loader.get_code(module.__name__)
```
georgevreilly, line 23: This, in essence, is what `loader.exec_module` does.

georgevreilly-stripe:
@serhii73, can I get this reviewed and merged, please?

Gallaecio (approved):
Looks good to me, though I wonder if we should go with a try-except approach instead.

wRAR (approved):

Issue #940, opened 2021-07-03
Outdated : CLDR Data


gavishpoddar:
CLDR data doesn't seem to be updated and languages like `as` are not translated properly.

Additionally,  CLDR data URL is archived: `https://github.com/unicode-cldr/cldr-dates-full` (archived) 

Thanks

PR #1071, opened 2022-08-25
Add 'Sept' for 'September'

wulmer:
Due to an error with my Google Sheets table, I found that Google (and [DIN 5008](https://www.din-5008-richtlinien.de/startseite/datum/)) use `Sept` as abbreviation for `September`, but `dateparser` only understood `Sep`. It looks like both `Sept` and `Sep` are valid.

This PR adds `Sept` as variant of `September`.

### DIFF
--- a/dateparser/data/date_translation_data/de.py
+++ b/dateparser/data/date_translation_data/de.py
@@ -37,6 +37,7 @@ info = {
     ],
     "september": [
         "sep",
+        "sept",
         "september"
     ],
     "october": [
--- a/dateparser_data/cldr_language_data/date_translation_data/de.json
+++ b/dateparser_data/cldr_language_data/date_translation_data/de.json
@@ -34,6 +34,7 @@
     ],
     "september": [
         "sep",
+        "sept",
         "september"
     ],
     "october": [
--- a/tests/test_date_parser.py
+++ b/tests/test_date_parser.py
@@ -117,6 +117,7 @@ class TestDateParser(BaseTestCase):
         param('21. Dezember 2013', datetime(2013, 12, 21)),
         param('19. Februar 2012', datetime(2012, 2, 19)),
         param('26. Juli 2014', datetime(2014, 7, 26)),
+        param('1. Sept 2000', datetime(2000, 9, 1)),
         param('18.10.14 um 22:56 Uhr', datetime(2014, 10, 18, 22, 56)),
         param('12-Mär-2014', datetime(2014, 3, 12)),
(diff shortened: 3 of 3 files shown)


wulmer:
@serhii73 : thanks a lot for your response. I have updated the JSON data now and also included a test.

Gallaecio:
I believe the CLDR data JSON file is not meant to be updated manually, that you should be updating the YAML file instead.

wulmer:
> I believe the CLDR data JSON file is not meant to be updated manually, that you should be updating the YAML file instead.

I don't think so: the script in `dateparser_scripts/write_complete_data.py` ([lines 49-54](https://github.com/scrapinghub/dateparser/blob/master/dateparser_scripts/write_complete_data.py#L49-L54)) first reads all data from the two directories (with the JSON and YAML files) and then combines the data before writing it to the Python files:

```python
    if language in cldr_languages:
        with open(cldr_date_directory + language + '.json') as f:
            cldr_data = json.load(f, object_pairs_hook=OrderedDict)
    if language in supplementary_languages:
        with open(supplementary_date_directory + language + '.yaml') as g:
            supplementary_data = OrderedDict(RoundTripLoader(g).get_data())
    complete_data = combine_dicts(cldr_data, supplementary_data)
```

At least, when running the script, the Python files get updated based on the changes in the JSON _or_ YAML files.

wulmer:
@Gallaecio : the upstream files in the given repos will not change as the repos are currently archived, see the referenced repos from [`get_raw_data`](https://github.com/scrapinghub/dateparser/blob/master/dateparser_scripts/utils.py#L8) at https://github.com/unicode-cldr/cldr-dates-full, https://github.com/unicode-cldr/cldr-core, and https://github.com/unicode-cldr/cldr-rbnf.

There is however a newer still maintained repo at https://github.com/unicode-org/cldr-json/ , and they seem to have `Sept` as abbreviation for `September`.

How should I proceed here now? Next year, in September, my scripts will crash again... :-(

wRAR (approved):

Gallaecio:
@wulmer Abotu CLDR data being outdated, that is it seems a [known issue](https://github.com/scrapinghub/dateparser/issues/940), and definitely one we would like to address.

RadhiFadlillah:
Apologize for necroing this PR, but I just want to notify that it seems the fix by @serhii73 is not merged in this PR.

I've checked the master branch, and the `sept` keyword is still applied directly in `de.json` file instead of in `de.yaml`.

PR #1075, opened 2022-09-12
Replace md5 with sha512

vvalouch:
MD5 is not FIPS compliant thus I am proposing replacement with SHA512.

### DIFF
--- a/dateparser/conf.py
+++ b/dateparser/conf.py
@@ -46,7 +46,7 @@ class Settings:
             return 'default'
 
         keys = sorted(['%s-%s' % (key, str(settings[key])) for key in settings])
-        return hashlib.md5(''.join(keys).encode('utf-8')).hexdigest()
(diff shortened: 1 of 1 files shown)


Gallaecio:
This is not a cryptographic use of md5, we are not handling unsafe input here, so I do not see the point: a slower algorithm with no added security benefit.

florin-szilagyi:
@Gallaecio we should allow some sort of switch
md5 is not FIPS compliant, and will crash on FIPS enabled machines. unless the `usedforsecurity=false` which is only avaialable on python >3.9

Gallaecio:
Sounds like setting `usedforsecurity=False` on Python 3.9+ is the way forward, then.

AdrianAtZyte:
Fixed in #1267, which keeps md5 but passes `usedforsecurity=False`, so it works on FIPS-enabled machines.

PR #1095, opened 2022-11-11
Replace regex sub with simple string replace

DJRHails:
Fixes https://github.com/scrapinghub/dateparser/issues/1045

Updates from: #1067


Gallaecio (approved):
Thanks!

PR #1113, opened 2022-12-28
Remove requirements-build file

serhii73:
Close #1112

requirements-build.txt is used only in this file. Do you know if we need this file and the wheel in it?


Gallaecio (approved):
If that’s the only reference to the file, I think we can go ahead.

In fact, I suspect we do not need `requirements.txt` either, but I don’t want to extend the scope of this change.

PR #1117, opened 2023-01-08
Support datetime.timezone

Vioshim:
Settings.TIMEZONE and Settings.TO_TIMEZONE are now compatible with user defined timezones.


Vioshim:
Closed as it has some performance issues.

PR #1121, opened 2023-01-10
Parse Monday

### LINKED ISSUE #1116: Is not parsing properly the Last-Modified header value
Hi,

Just in case, in the previous version we were able to parse the following 'Mon, 05 Dec 2022 19:43:03 GMT' without any problem and now is returning None.

This is happening only when is Monday.

It could be related to this commit: https://github.com/scrapinghub/dateparser/commit/89a5b29517a358e2784cc818687b7b6bed343159 ?

Thank you!

Best,
Nicolas.

serhii73:
Close #1116

### DIFF
--- a/tests/test_freshness_date_parser.py
+++ b/tests/test_freshness_date_parser.py
@@ -1518,6 +1518,16 @@ class TestFreshnessDateDataParser(BaseTestCase):
         self.assertEqual(None, self.result['date_obj'])
 
     @parameterized.expand([
+            param('1mon ago'), #1116
+      ])
+    def test_known_issues(self, date_string):
+        self.given_parser()
+        self.given_date_string(date_string)
+        self.when_date_is_parsed()
+        self.then_error_was_not_raised()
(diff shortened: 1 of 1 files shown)


tests/test_freshness_date_parser.py
```diff
@@ -97,7 +97,6 @@ def test_relative_past_dates_with_time_as_period(self, date_string, ago, period)
         param('nine hours ago', ago={'hours': 9}, period='day'),
         param('three week ago', ago={'weeks': 3}, period='week'),
         param('eight months ago', ago={'months': 8}, period='month'),
-        param('1mon ago', ago={'months': 1}, period='month'),
```
Gallaecio, line 100: Similar to how there is `test_insane_dates` and `test_dates_not_supported_by_date_time`, we could add a `test_known_issues` that keeps this test case around but expects it to fail, and include a comment pointing to the issue that caused us to break it.

(I thought we were using pytest’s parametrize when I suggested marking this line as an expected failure)

Gallaecio (approved):

PR #1129, opened 2023-01-24
Fix ago bug

serhii73:
Close #340


Gallaecio:
`"(\\d+)\s*год\\b": "\\1"`, as [originally suggested](https://github.com/scrapinghub/dateparser/pull/340#issuecomment-628660191) by @noviluni, works for the new test cases, but fails for existing ones.

Because I do not speak Russian, I am not sure how to best address those other issues.

For example, according to tests, `1 год 2 месяца` means 1 year and 2 months ago. Is that correct? If so, does `1 год` mean 1 year ago or year 1? How can you tell?

Gallaecio:
How can you tell `1 год` means year 1 but `1 год 2 месяца` means 1 year and 2 months ago? How would you say 1 year ago in line with `1 год 2 месяца`?

Gallaecio:
OK, so I have gone with 2 regular expressions:
- Given a number of 4+ digits ending in 1 and followed by `год` and *only* by `год`, remove the “years” suffix so that it is treated as an absolute year rather than a number of years. I went with this after I wound that `parse("1000")` interprets 1000 as the year, but `parse("999")` gives no result.
- Given any number not ending in 1 and followed by `год`, remove the “years” suffix so that it is treated as an absolute year.

I have added a few more test cases that I hope are accurate. Please take a good look.

wRAR (approved):

PR #1137, opened 2023-02-06
English data added

ArunSinghJio:
English data aliases added


ArunSinghJio:
Adding data to make suitable for my testcases

PR #1141, opened 2023-02-11
Added check for date-formats

### LINKED ISSUE #1020: `parse` parsing strings not matching `date_formats`
Is this a bug, an undocumented feature, or maybe I didn't ready the docs thoroughly enough ?=/

```
In: dp.parse('Time', date_formats=['%d.%m.%Y %H:%M'])
Out: datetime.datetime(2021, 11, 19, 0, 0)

In: dp.parse('(m)', date_formats=['%d.%m.%Y %H:%M'])
Out[58]: datetime.datetime(2021, 11, 19, 0, 0)
```

I'd expect this to return None to be [consistent ](https://dateparser.readthedocs.io/en/latest/dateparser.html?highlight=parse#dateparser.parse)when `date_string` doesn't match any `date_formats`.

```
In : dp.parse('something', date_formats=['%d.%m.%Y %H:%M'])
Out: None

```

RyanWalker277:
Fixes #1020 

## Code changes
Added a function to check if an input `date_string` belongs to specefied `date_formats` or not 

## Help Required 
Even though I did an extremely small code change , It is failing some tests. Can any of the available maintainers help me determine the cause?

cc - @Gallaecio

### DIFF
--- a/dateparser/date.py
+++ b/dateparser/date.py
@@ -438,6 +438,9 @@ class DateDataParser:
             period='day', locale='en')
 
         """
+        if any(datetime.strptime(date_string, format) for format in date_formats):
+            raise TypeError('Input is of unknown format')
+        
         if not isinstance(date_string, str):
             raise TypeError('Input type must be str')
 


RyanWalker277:
Hey @Gallaecio , I have fixed the test cases issue. It is now passing all the test cases. Kindly review it.

Gallaecio (commented):
Looking at the code, it feels to me like this change will have no effect whatsoever.

In any case, could you add a test based on the original issue that does not pass before your change and starts passing thanks to your change?

PR #1143, opened 2023-02-21
Set "usedforsecurity' to Flase for md5 checksum.

sgonsal:
There is no cryptographic use of md5 done. So setting usedforsecurity to False. Doing so will make this FIPS complaint.


Gallaecio (approved):
Today I learned!

PR #1148, opened 2023-03-03
Update zh data

serhii73:


wRAR (approved):

Gallaecio (approved):

RadhiFadlillah:
If you don't mind, I'm wondering why does the modification in this PR are done directly in CLDR JSON file?

From what I understand, those CLDR files are synced periodically with the upstream repo,so it shouldn't be modified. Instead, we should modify the supplemental `*.yaml` file which later will be merged with the CLDR data.

Is my understanding correct? Or have there been recent changes to the way translations are contributed?

Gallaecio:
You are right, we should only be modifying YAML files.

@serhii73 It is probably OK for now, but whenever we update the CLDR JSON files from upstream next, we may need to check if what we added here exists upstream, and if not, we would need to move the data to the YAML files to make sure it is included going forward.

PR #1157, opened 2023-03-22
Release 1.1.8

serhii73:


Gallaecio (approved):

HISTORY.rst
```diff
@@ -3,6 +3,19 @@
...
+------------------
+
+Improvements:
+- Improved date parsing for Chinese (#1148)
+- Improved date parsing for Czech (#1151)
+- Reorder language by popularity (#1152)
+- Fix leak of memory in cache (#1140)
+- Add support for "\d units later" (#1154)
+- Move modification in CLDR data to yaml (#1153)
+- Use timezone via settings to get PREFER_DATES_FROM result (#1155)
```
gutsytechster, line 16: This may look like we are only going to use timezone via settings. Can we restructure it like this - `Add support to use timezone via settings to get PREFER_DATES_FROM result`? Does it sound better? Or do we have some guidelines to use the PR titles as the changelog headers I may be unaware of?

gutsytechster (approved):

PR #1172, opened 2023-07-04
Updated Czech localization

TheXer:
In some cases, there's some inconsistencies with Czech grammar. For example, in English:

`In a month` will output a correct datetime object - 1 month from now. But in Czech `za měsíc` won't print anything, but `příští měsíc` will. Both are correct Czech sentences.

 The same proposed changes applies to `za týden` (in a week), `za hodinu` (in a hour), `příští rok` (next year).

Added new words:
In two days - `pozítří`
Two days ago - `předevčírem`
Last year - `loni` or `vloni`

And finally, added some new declensions of initial words.

### DIFF
--- a/dateparser/data/date_translation_data/cs.py
+++ b/dateparser/data/date_translation_data/cs.py
@@ -4,45 +4,54 @@ info = {
     "january": [
         "led",
         "leden",
-        "ledna"
+        "ledna",
+        "lednu"
     ],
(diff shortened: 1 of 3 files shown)


Gallaecio:

wRAR (approved):

serhii73:
Thanks!

Issue #533, opened 2019-06-28
Import very slow


skjerns:
I figured that the import of this library is very slow (500-1000ms on my system).
Given that the library probably less complex than e.g. `numpy` or similarly large packages: Do you think the import times can be optimized in some way?

noviluni:
Hi @skjerns , could you confirm that this is happening with the last version?

noviluni:
On the other side, the import is a little slower than other libraries because it initializes some parser to be faster when executing it, so I think there isn't too much we can do.

skjerns:
@noviluni it's faster now (280ms), but I've also upgraded my PC, so I can't really tell ;-) 

using `python`: 100ms
using `ipython`: 280ms

Maybe you could implement some lazy loading to the parsers? Or are they all needed?

noviluni:
I tried with the last version of `dateparser` (1.0.0) and it takes less time than `pandas`, so I think it shouldn't be considered an issue.

I will close it. Feel free to comment or reopen it if you thing there is something that should be improved :slightly_smiling_face:

skjerns:
Yes! Seems to be much faster now :)

Natureshadow:
dateparser 1.1.8 takes around 600ms to import here.

skjerns:
> dateparser 1.1.8 takes around 600ms to import here.

cannot replicate: `Elapsed context: 275 ms`

Natureshadow:
> cannot replicate: Elapsed context: 275 ms

So, your system is faster than mine.

275ms import time is still unacceptable (basically everything >10ms on an average system is a bug)

beda42:
I have to agree that hundreds of milliseconds are a problem. For me the import is about 250 ms and about 200 ms from that is the generation of `_search_regex_ignorecase` and related intermediate values at the end of `timezone_parser.py`:

```python
_search_regex_parts = []
_tz_offsets = list(build_tz_offsets(_search_regex_parts))
_search_regex = re.compile('|'.join(_search_regex_parts))
_search_regex_ignorecase = re.compile(
    '|'.join(_search_regex_parts), re.IGNORECASE)
```

(from that 140 ms is just `_tz_offsets = list(build_tz_offsets(_search_regex_parts))`, but that does not really matter).

[...]

PR #1181, opened 2023-09-07
postpone timezone regex evaluation until first use - shaves off time from package import

beda42:
This MR is related to issue #533 which is caused mainly by a time intensive parsing of regular expressions for timezone matching. This MR introduces a global object `TzRegexCache` and moves preparation of the regexps into it. The regexps are no longer parsed on startup, but during first use instead. On my machine it shaves off about 200 ms from the import time of `dateparser` thus reducing it to less than 20 % of the original import time.

### DIFF
--- a/dateparser/timezone_parser.py
+++ b/dateparser/timezone_parser.py
@@ -32,8 +32,8 @@ class StaticTzInfo(tzinfo):
 
 
 def pop_tz_offset_from_string(date_string, as_offset=True):
-    if _search_regex_ignorecase.search(date_string):
-        for name, info in _tz_offsets:
+    if TzRegexCache.search_regex_ignorecase().search(date_string):
+        for name, info in TzRegexCache.tz_offsets():
             timezone_re = info["regex"]
             timezone_match = timezone_re.search(date_string)
             if timezone_match:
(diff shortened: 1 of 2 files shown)


tobymao:
is this library still actively maintained? i ran into this issue and would like to fix this. my idea is instead to cache the regexes so that overall time gets faster. this solution will still incur regex parsing time, although it does speed up parsing.

but i only want to do a PR if it actually has a shot at making it in.

Gallaecio:
We are reviewing PRs, and may even have a release “soon” (see the most recent PRs).

tobymao:
> We are reviewing PRs, and may even have a release “soon” (see the most recent PRs).

amazing! would you accept my idea? is there somewhere to discuss with maintainers this approach before i start to make iterations faster?

Gallaecio:
You can either start a draft PR with little code and start the discussion there, or discuss it here.

tobymao:
i've taken another stab at this @beda42 https://github.com/scrapinghub/dateparser/pull/1250

instead of lazy evaluation, i do caching so that we only pay the regex compile time cost once. it will break the cache if there's a version bump.

PR #1183, opened 2023-09-14
Fix skipping day when parsing times if UTC and TIMEZONE day are different

### LINKED ISSUE #1169: Returned datetime skips a day
I reported in #1092 that the datetime is returned correctly only when supplying the timezone in the argument but not via settings, which is still the behaviour as of 1.1.7. As of 1.1.8, both options give the same result again, however the date is wrong. See an example below.

Note this is a different example than the one provided in #1092, the original example now works correctly on 1.1.8. This bug seems to occur when the time is already the next day in UTC.

**Versions:**

- Ubuntu 22.04.1 LTS
- Python 3.10.6
- dateparser (1.1.8)
  - python-dateutil (2.8.2)
  - pytz (2022.6)
  - regex (2022.3.2)
  - tzlocal (4.2)

**Minimum working example**
```
#!/usr/bin/env python3
import dateparser
import datetime

parse_settings={
        'PREFER_DATES_FROM': 'future',
        'TO_TIMEZONE': 'etc/utc',
        'RETURN_AS_TIMEZONE_AWARE': False,
        'RELATIVE_BASE': datetime.datetime(2023, 5, 21, 15, 0)
        }

time = dateparser.parse('7pm EDT', settings=parse_settings)
print(time) # 2023-05-21 23:00:00
time = dateparser.parse('9pm EDT', settings=parse_settings)
print(time) # 2023-05-23 01:00:00

parse_settings['TIMEZONE'] = 'america/new_york'
time = dateparser.parse('7pm', settings=parse_settings)
print(time) # 2023-05-21 23:00:00
time = dateparser.parse('9pm', settings=parse_settings)
print(time) # 2023-05-23 01:00:00
```

MxMarx:
This should fix #1169 

If only a time is provided such as "8pm", the previous logic determines if the time is in the "future" or "past" by comparing the base time of day and the parsed time.
This works if the timezone is UTC, but to make this respect the TIMEZONE setting, #1002 subtracts the timezone's offset from the parsed time. If this causes the day to roll over, then comparing times will incorrectly report the parsed date as being in the past. For example. `datetime(2023, 2, 3, 4).time() == datetime.time(4, 0)` is reported as earlier tham ` datetime(2023, 2, 2, 18).time() == datetime.time(18, 0)`, and erroneously adding day. To correct this, we could compare dates rather than times.

```
from datetime import datetime
import dateparser
import pytz

now = datetime(2023, 2, 2, 10).astimezone(pytz.timezone("America/Los_Angeles"))
print(now)      # 2023-02-02 10:00:00-08:00
now_utc = now.astimezone(pytz.timezone("UTC")).replace(tzinfo=None)
print(now_utc)  # 2023-02-02 18:00:00

settings = {
    'PREFER_DATES_FROM': 'future',
    'RETURN_AS_TIMEZONE_AWARE': True,
    'RELATIVE_BASE': now_utc,
    'TIMEZONE': 'America/Los_Angeles',
    "TO_TIMEZONE": "etc/utc",
}
time = dateparser.parse('2pm', settings=settings)
print(time) # 2023-02-02 22:00:00+00:00 - correct

time = dateparser.parse('8pm', settings=settings)
print(time) # 2023-02-04 04:00:00+00:00 - incorrect

# datetime(2023, 2, 3, 4).time() = datetime.time(4, 0)
# datetime(2023, 2, 2, 18).time() = datetime.time(18, 0)
# datetime.time(4, 0) < datetime.time(18, 0) which causes date to be added even though the date is in the future

```

### DIFF
--- a/dateparser/parser.py
+++ b/dateparser/parser.py
@@ -563,18 +563,15 @@ class _parser:
             # Convert dateobj to utc time to compare with self.now
             try:
                 tz = tz or get_timezone_from_tz_string(self.settings.TIMEZONE)
+                tz_offset = tz.utcoffset(dateobj)
             except pytz.UnknownTimeZoneError:
-                tz = None
+                tz_offset = timedelta(hours=0)
 
-            if tz:
-                dateobj_time = (dateobj - tz.utcoffset(dateobj)).time()
-            else:
-                dateobj_time = dateobj.time()
             if "past" in self.settings.PREFER_DATES_FROM:
-                if self.now.time() < dateobj_time:
+                if self.now < dateobj - tz_offset:
                     dateobj = dateobj + timedelta(days=-1)
             if "future" in self.settings.PREFER_DATES_FROM:
-                if self.now.time() > dateobj_time:
+                if self.now > dateobj - tz_offset:
                     dateobj = dateobj + timedelta(days=1)
 
         # Reset dateobj to the original value, thus removing any offset awareness that may
--- a/tests/test_date_parser.py
+++ b/tests/test_date_parser.py
@@ -1213,6 +1213,16 @@ class TestDateParser(BaseTestCase):
                 {"PREFER_DATES_FROM": "future"},
             ),
             param(
+                "10pm EDT",
+                datetime(2021, 10, 20, 2, 0),
+                {"PREFER_DATES_FROM": "future"},
+            ),
+            param(
+                "8am AEDT",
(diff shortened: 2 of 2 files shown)


Baviaan (approved):
Thanks for making these changes, looks good!

Gallaecio:
Closing and reopening to trigger CI (hopefully).

Gallaecio:
Perfect, thank you!

PR #1187, opened 2023-09-25
support two-digit year in non-gregorian calendars

sm-ghavami:
We had an issue with parsing two digit year in non-gregorian calendars. This PR will fix that issue and provide a fix for both jalali and hijri calendars.

### DIFF
--- a/dateparser/calendars/__init__.py
+++ b/dateparser/calendars/__init__.py
@@ -76,6 +76,9 @@ class non_gregorian_parser(_parser):
 
         return result
 
+    def handle_two_digit_year(self, year):
+        raise ValueError
+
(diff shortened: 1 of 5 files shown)


Gallaecio (approved):
Nice.

I wonder if we should do something similar for Gregorian, but no need to do it here.

sm-ghavami:
The primary advantage of this approach over the Gregorian one is its customizability. Everyone can easily extend the parsers and override the `handle_two_digit_year` method to tailor it to their specific use cases.

sm-ghavami:
@Gallaecio This PR has been approved but not merged yet. Could you check this please?

Issue #683, opened 2020-05-16
Timezones support is deficient


noviluni:
I think we should rethink how the timezones are applied, as we have more than one case of abbreviations with different meanings.

I could found at least these cases:

* **ACT:** 
    * Acre Time: UTC−05
    * ASEAN Common Time (unofficial): UTC+06:30 – UTC+09

* **AMT:**
    * Amazon Time (Brazil): UTC−04
    * Armenia Time: UTC+04

* **AST:**
    * Arabia Standard Time: UTC+03
    * Atlantic Standard Time: UTC−04

* **BST:** 
    * Bangladesh Standard Time: UTC+06
    * British Summer Time (British Standard Time from Feb 1968 to Oct 1971): UTC+01
    * Bougainville Standard Time: UTC+11
    * issues / PRs: https://github.com/scrapinghub/dateparser/pull/682

* **CDT:**
    * Central Daylight Time (North America): UTC−05
    * Cuba Daylight Time: UTC−04

* **CST:**
    * Central Standard Time (North America): UTC−06
    * Cuba Standard Time: UTC−05
    * China Standard Time: UTC+08

* **ECT:**
    * Ecuador Time: UTC−05
    * Eastern Caribbean Time: UTC−04

* **GST:**
    * Gulf Standard Time: UTC+04
    * South Georgia and the South Sandwich Islands Time: UTC−02

* **IST:** 
    * Irish Standard Time: UTC+01
    * Israel Standard Time: UTC+02
    * Indian Standard Time: UTC+05:30
    * issues / PRs: https://github.com/scrapinghub/dateparser/issues/580 https://github.com/scrapinghub/dateparser/issues/609 https://github.com/scrapinghub/dateparser/issues/636 https://github.com/scrapinghub/dateparser/pull/637 

I'm not sure about the solution we could implement, but it's obvious that we can not choose a timezone randomly.

Some ideas:
* We could choose the most used, for example, we could maybe change IST from "Israel Standard Time" (current) to "Indian Standard Time" (hard to decide, not a solution itself).
* We could add a warning when parsing timezones with more than one meaning (self-explanatory, but not a solution and it could be a problem when people already know it, as it could spam with a lot of warning messages).
* We could add a setting like "PREFER_TIMEZONE_FROM" or something similar where users can specify if they prefer timezones from a continent or another negative vs positive timezones (I'm not sure about this, as there are some examples in the list (like CDT or ECT) where all the timezones are from the same continent/region).
[...]

PR #1190, opened 2023-10-09
Update timezones.py

KhawerAbbas7:
Corrected IST


Gallaecio:
The right approach to this needs to be discussed in https://github.com/scrapinghub/dateparser/issues/683 first, and will probably not be as trivial as this. We might need a setting to allow choosing how to interpret some timezones.

PR #1201, opened 2023-11-22
Fixed RecursionError discovered by OSSFuzz

ennamarie19:
This PR introduces a fix to resolve the crashes discovered thus far on ClusterFuzz.

Currently, _split_by_known_words was implemented as a recursive function with two main issues

1) Infinite recursion was possible when the `unknown` variable was always equal to the original `string`
2) Large inputs cause StackOverflow, as Python has a very limited call-stack before raising a RecursionError

This fix converts the function into an iterative one. All tests that didn't fail before still don't fail. 

Thank you for your review!

### DIFF
--- a/dateparser/languages/dictionary.py
+++ b/dateparser/languages/dictionary.py
@@ -181,26 +181,34 @@ class Dictionary:
         ):
             cache.pop(list(cache.keys())[0])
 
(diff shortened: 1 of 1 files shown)


Gallaecio (approved):
:heart:

PR #1203, opened 2023-12-16
OSSFuzz Bug Fix 

ennamarie19:
This pull request fixes a crash encountered during OSSFuzz continuous fuzzing. Previously, an InvalidTimeError exception in pytz's UTC offset when processing 01-01-1970 0400 would cause parsing to exit prematurely. This fix catches the exception and resolves dateobj_time the secondary way. 

Additionally, to detect bugs prior to them being merged into the codebase, the CIFuzz.yml job that comes with OSSFuzz is proposed in this PR. With this, each PR will be fuzz-tested as part of the CI workflow. 

Thank you for your time and your review!

### DIFF
--- a/dateparser/parser.py
+++ b/dateparser/parser.py
@@ -144,10 +144,10 @@ class _no_spaces_parser:
             "%m%d%y": (
                 self._preferred_formats
                 + sorted(
-                    self._all,
-                    key=lambda x: x.lower().startswith("%m%d%y"),
-                    reverse=True,
-                )
+                self._all,
+                key=lambda x: x.lower().startswith("%m%d%y"),
+                reverse=True,
+            )
             ),
             "%m%y%d": sorted(
                 self._all, key=lambda x: x.lower().startswith("%m%y%d"), reverse=True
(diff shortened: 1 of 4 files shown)


Gallaecio:
Could we include a test case for this scenario as well?

capuanob:
@Gallaecio Good evening, we just pushed a test case to go with the change. It utilizes the same settings and args that caused the bug during fuzzing to ensure that this bug is not reintroduced down the line.

Thank you!

dateparser/parser.py
```diff
@@ -564,9 +564,19 @@ def _correct_for_time_frame(self, dateobj, tz):
...
 
+            dateobj_time = None
+            if tz:
+                try:
+                    dateobj_time = (dateobj - tz.utcoffset(dateobj)).time()
+                except pytz.InvalidTimeError:
+                    pass
+
+            if not dateobj_time:
+                dateobj_time = dateobj.time()
```
Gallaecio, line 578: This code looks like a no-op at the moment, it defines a `dateobj_time` variable that is never used.

ennamarie19:
@Gallaecio It looks like someone updated the same function and merged to main since this fix, which caused a merge conflict that was not resolved correctly. This latest commit actually handles the bug

Gallaecio (approved):
Thanks!

PR #1208, opened 2024-01-07
Fixed ClusterFuzz build error by adding dateparserr.data as a binary

capuanob:
Fixes the nightly build of the OSSFuzz continuous testing.


capuanob:
The above CIFuzz job will pass once this is merged in!

Issue #1197, opened 2023-11-10
hijri-converter deprecation


gholam-deblan:
This package has been deprecated and will not receive future updates. Please use the [hijridate](https://pypi.org/project/hijridate/) package instead.

For migration to hijridate, install the package:

pip install hijridate~=2.3.0

Then change the import statement as:

from hijridate import Hijri, Gregorian

PR #1211, opened 2024-01-25
Migrate from hijri-converter to hijridate

pastalian:
hijri-converter has been renamed to hijridate since 2.3.0.
Closes: #1197


Gallaecio:
Thanks!

PR #1222, opened 2024-02-28
Add `en-US` locale

ZelGray:
The code for English (United States) 'en-US'. The prefix, 'en', is a language code following the [ISO 639-1 standard](https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes). The suffix, 'US', is a country code following the [ISO 3166-1 Alpha-2](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) standard.
E.g. Discord uses that locale name for language `English, US`

### DIFF
--- a/dateparser/data/languages_info.py
+++ b/dateparser/data/languages_info.py
@@ -496,6 +496,7 @@ language_locale_dict = {
         "en-TZ",
         "en-UG",
         "en-UM",
+        "en-US",
         "en-VC",
         "en-VG",
         "en-VI",


Gallaecio:
Could you elaborate on why you are proposing this change? I think dateparser already uses “en-US” by default if you don’t specify a different country.

soksanichenko:
Hello.
That's my PR, I really forgot about that. 
I remember dataparser doesn't work properly if I use something like this

```import dateparser
options = {'locales': ['en-US']}
dataparser.parse('12:00', **options
```

I will get error `ValueError: Unknown locale(s): 'en-US'`

Gallaecio:
But *why* would you set that locale? Does `en` not work as expected?

soksanichenko:
That's just an example.
In my code I pass a locale as variable and don't know what's exactly locale it can be. Because of it I referred to Discord in the start post.

AdrianAtZyte:
Thanks!

PR #1223, opened 2024-03-04
Check for derived types for configuration

nadavge:
Currently if we want to use a derived type for one of the configuration fields, we can't do that. This aims to permit it.

We were using a class that overrides datetime for our tests, and needed this change.


Gallaecio (approved):

Gallaecio:
Thank you!

PR #1238, opened 2024-10-02
Avoid repeated loop over timezones

nijel:
By removing repeated iteration over timezones, the import time is reduced by 5-10%.


Gallaecio:
Wow, great catch!

nijel:
Hmm, seems like ordering does matter in the end, looking into the test failures.

nijel:
Hmm, I'm getting the same failures on the master branch, so these are probably unrelated to my changes.

Gallaecio (approved):

PR #1241, opened 2024-10-23
chore: update Python version

serhii73:


dateparser/date_parser.py
```diff
@@ -32,7 +32,7 @@ def parse(self, date_string, parse_method, settings=None):
         else:
             if "local" in _settings_tz:
                 stz = get_localzone()
-                if hasattr(stz, "localize") and sys.version_info < (3, 6):
+                if hasattr(stz, "localize") and sys.version_info < (3, 9):
```
wRAR, line 35: Is this correct?

.github/workflows/main.yml
```diff
@@ -38,10 +32,22 @@ jobs:
...
+            env:
+              TOXENV: py
+          - python-version: "3.13"
+            env:
+              TOXENV: latest
+          - python-version: "3.13"
             env:
               TOXENV: twinecheck
-          - python-version: "3.11"  # Keep in sync with tox.ini
+          - python-version: "3.13"  # Keep in sync with tox.ini
```
wRAR, line 50: Not sure why does this mention tox.ini as we usually need to sync this with `.readthedocs.yml` (which wasn't changed in this PR, but we currently can't use 3.13 to build docs as RTD doesn't support that yet, so the docs Python version should be bumped only to 3.12.

wRAR:
Besides the test failures, it looks like all non-test envs (flake8, twinecheck, docs) run `tox -e py` instead of the correct check.

dateparser/date_parser.py
wRAR: Actually this is always False now, so this whole block can be removed. There may be other such places in the code so I suggest looking for them and cleaning them up, as a separate PR.

PR #1259, opened 2025-03-17
use different timezones to run CI tests

gutsytechster:


Gallaecio (approved):
Nice! Maybe a single TZ job would be enough, but :shrug:

gutsytechster:
@Gallaecio yeah. The only difference between the two is that one observes DST while the other doesn't. So I kept both of them.

Issue #1258, opened 2025-03-14
dateparser doesn't work in FIPS compliant environments


trevor-scale:
Many secure environments disable the `hashlib.md5` function because it is insecure, the causes the line below to fail making dateparser unusable in these environments:

https://github.com/scrapinghub/dateparser/blob/02bd2e5dd4477b4f6db98c5e98149458eb3cc821/dateparser/conf.py#L52

This can easily be remedied by replacing that line with `return hashlib.md5("".join(keys).encode("utf-8"), usedforsecurity=False).hexdigest()` which appropriately bypasses the fips security check because dateparser is not using the md5 hash for security applications.

This is a great project, would appreciate it if this fix could be pushed so it becomes useable in FIPS-compliant environments!

trevor-scale:
Thanks for the ack. I'd be happy to PR this myself if you give me permissions to contribute!

Gallaecio:
No need for permissions, feel free to create a PR from a fork.

ghmj2417:
If trevor-scale is unable to create the PR, I would be more than happy to do it. However, wouldn't this change be a breaking change since the setup.py requires python3.8 or later. This change would force users to 3.9.

ghmj2417:
This should support >= 3.8
`return hashlib.new("md5", "".join(keys).encode("utf-8"), usedforsecurity=True).hexdigest()`

trevor-scale:
PR submitted here: https://github.com/scrapinghub/dateparser/pull/1267

Gallaecio:
Removing support for 3.8 is fine, it is end-of-life.

ghmj2417:
@trevor-scale Looks like @Gallaecio doesn't mind the breaking change since 3.8 is end of support. I will leave it to you guys on how the PR should be adjusted. 

Also, I am not a maintainer of this repo so I cannot approve or merge the PR.

AdrianAtZyte:
Fixed in #1267, released in dateparser 1.4.0 (2026-03-26).

PR #1267, opened 2025-04-15
Allow md5hash to function in fips environments

trevor-scale:
Many secure environments disable the `hashlib.md5` function because it is insecure, this causes the line below to fail making dateparser unusable in these environments:

https://github.com/scrapinghub/dateparser/blob/02bd2e5dd4477b4f6db98c5e98149458eb3cc821/dateparser/conf.py#L52

This is remedied by replacing that line with `return hashlib.new("md5", "".join(keys).encode("utf-8"), usedforsecurity=True).hexdigest()` which appropriately bypasses the fips security check because dateparser is not using the md5 hash for security applications.

### DIFF
--- a/dateparser/conf.py
+++ b/dateparser/conf.py
@@ -49,7 +49,7 @@ class Settings:
             return "default"
 
         keys = sorted(["%s-%s" % (key, str(settings[key])) for key in settings])
-        return hashlib.md5("".join(keys).encode("utf-8")).hexdigest()
+        return hashlib.new("md5", "".join(keys).encode("utf-8"), usedforsecurity=True).hexdigest()
 
     @classmethod
     def _get_settings_from_pyfile(cls):


trevor-scale:
> I posted this in #1258 as well. Change `usedforsecurity=True` to `usedforsecurity=False`. The suggestion/example provided in issue 1258 was to use `hashlib.new` but it also included the full command which had `usedforsecurity=True` to test a failure when using FIPS compliant encryptions/hashes/etc. My apologies for not being clear.

Thanks for catching this wasn't looking closely when I copy-pasted haha. Will update now!

trevor-scale:
> > I posted this in #1258 as well. Change `usedforsecurity=True` to `usedforsecurity=False`. The suggestion/example provided in issue 1258 was to use `hashlib.new` but it also included the full command which had `usedforsecurity=True` to test a failure when using FIPS compliant encryptions/hashes/etc. My apologies for not being clear.
> 
> Thanks for catching this wasn't looking closely when I copy-pasted haha. Will update now!

@ghmj2417 are you able to approve and merge this, would love to get this added in!

trevor-scale:
@Gallaecio can you give this a review and lmk if any changes are needed when you have the chance?

dateparser/conf.py
```diff
@@ -49,7 +49,7 @@ def get_key(cls, settings=None):
             return "default"
 
         keys = sorted(["%s-%s" % (key, str(settings[key])) for key in settings])
-        return hashlib.md5("".join(keys).encode("utf-8")).hexdigest()
+        return hashlib.new("md5", "".join(keys).encode("utf-8"), usedforsecurity=False).hexdigest()
```
Gallaecio, line 52: Why switch to  `new`? 

This seems also possible:

```suggestion
        return hashlib.md5("".join(keys).encode("utf-8"), usedforsecurity=False).hexdigest()
```

ghmj2417:
hashlib.new was to accommodate Python 3.8. If setup.py is going to require >= 3.9 then hashlib.md5 is fine.

PR #1311, opened 2026-02-09
Move project metadata and configuration to 'pyproject.toml'

EpicWink:
Makes most metadata and build configuration declarative

Also:

* Drop `License :: OSI Approved :: BSD License` classifier (license classifiers are deprecated)
* Specify license using [SPDX](https://spdx.org/licenses/) identifier
* Specify `setuptools` explicitly as a build requirement using PEP 518 (and as the backend using PEP 517)
* Fix keywords declaration (from string to list)

I kept your custom `long_description` (pyproject.toml: `readme`) construction: there's no harm in it. Setuptools also has native functionality to read the version from a Python module attribute.


setup.py
```diff
@@ -16,47 +12,5 @@
 )
```
Gallaecio: I think we need to figure out how to get rid of `setup.py` entirely. Maybe we could use the README.rst instead for the long description.

pyproject.toml
```diff
@@ -1,3 +1,62 @@
+[build-system]
+build-backend = "setuptools.build_meta"
+requires = [
+    "setuptools >= 77.0",
+]
```
Gallaecio: Could we use hatchling instead, as [we do in Scrapy](https://github.com/scrapy/scrapy/blob/06fb87f7bb0cd860536fc8d15415cb3d78314f15/pyproject.toml#L1-L3)?

pyproject.toml
```diff
@@ -1,3 +1,62 @@
...
+[tool.setuptools]
+include-package-data = true
+zip-safe = false
+
+[tool.setuptools.packages.find]
+namespaces = false
+exclude = ["tests", "tests.*"]
+
+[tool.setuptools.dynamic]
+version = { attr = "dateparser.__version__" }
```
Gallaecio: We use `bumpversion`. So it probably makes sense to hard-code the version instead of making it dynamic, bumpversion will take care of updating it in both places. In fact, I think it can update it in pyproject.toml without specific configuration for it, and it may be possible to remove the `current_version` definition under `[tool.bumpversion]` once the project version is defined through standard pyproject.toml syntax.

EpicWink: I have no familiarity with `bumpversion`, and its docs (and its fork's docs) don't mention pyproject.toml. This pull request currently keeps the current release process unchanged, but I'm open to suggestions.

Gallaecio: https://callowayproject.github.io/bump-my-version/reference/configuration/global/#current_version

> ‡ If pyproject.toml exists, then current_version falls back to project.version in pyproject.toml. This only works if project.version is statically set.

serhii73 (approved):

pyproject.toml
```diff
@@ -1,8 +1,68 @@
+[build-system]
+build-backend = "setuptools.build_meta"
+requires = [
+    "setuptools >= 77.0",
+]
```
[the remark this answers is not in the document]
EpicWink: No, `wheel` is required transitively via `setuptools` when necessary

pyproject.toml
```diff
@@ -1,8 +1,68 @@
...
+
 [tool.bumpversion]
-current_version = "1.4.0"
 commit = true
 tag = true
 
+[[tool.bumpversion.files]]
+filename = "pyproject.toml"
+search = 'version = "{current_version}"'
+replace = 'version = "{new_version}"'
```
AdrianAtZyte, line 64: Hmm… I may be wrong here.

AdrianAtZyte (approved):

PR #1321, opened 2026-04-01
BST is utc+1 instead of +11

Sphynx-HenryAY:


AdrianAtZyte:
[That‘s arguable](https://en.wikipedia.org/wiki/BST#Time_zones).

AdrianAtZyte:
But it also sounds like the most reasonable interpretation. The 11 value seems to be based in an old Alaska timezone, and the British one is the in the tz database.

AdrianAtZyte:
Fixed in #1366, released in dateparser 1.4.3 (2026-09-03).

PR #1324, opened 2026-04-09
fix: preserve whitespace when removing translation tokens (fix #1302)

### LINKED ISSUE #1302: Issue: Extra space in translation
When translating dates in certain languages, skip tokens (like Finnish "klo" meaning "o'clock") that are removed during translation leave behind extra whitespace, resulting in double spaces in the output.

Example:

Input: "28 maalis klo 9:37" (Finnish)
Current output: "28 march  9:37" (with double space)
Expected output: "28 march 9:37" (with single space)

https://github.com/scrapinghub/dateparser/pull/1301#discussion_r2712041608

serhii73:
Close #1302


tests/test_languages.py
```diff
@@ -109,49 +109,57 @@ def setUp(self):
             param("it", "Giovedi Maggio 29 2013", "thursday may 29 2013"),
             param("it", "19 Luglio 2013", "19 july 2013"),
             # Portuguese
-            param("pt", "22 de dezembro de 2014 às 02:38", "22  december  2014  02:38"),
+            param("pt", "22 de dezembro de 2014 às 02:38", "22 december 2014 02:38"),
```
AdrianAtZyte, line 112: Is this change needed?

Ideally, these tests should continue to pass _as is_.

serhii73: My idea here was to make the library preserve the original spacing from the input, so the output would contain the same number of spaces as provided in the input string.

What approach would you suggest here?

AdrianAtZyte: Oh! My bad, I did not realize the modified strings were automatic dateparser translations of the inputs, I thought they were raw inputs!

AdrianAtZyte (approved):

PR #1325, opened 2026-04-19
Better cs translation

honzaHlavnicka:
This pull request improves Czech language parsing in dateparser.
I updated cs.yaml by adding additional word forms, grammatical cases, and synonyms.
It also introduces support for Czech cardinal and ordinal numbers (including inflected forms), adds Czech variants for AM/PM, and enables parsing of expressions like "půl druhé" (1:30).

Notes:
- Regenerated cs.py from the updated YAML
- Added tests covering the new functionality
- All tests are passing


dateparser_data/supplementary_language_data/date_translation_data/cs.yaml
```diff
@@ -1,24 +1,35 @@
-skip: ["přibližně", "v"]
+skip: ["přibližně", "v", "ve", "přibližně", "asi", "zhruba", "cca", "kolem", "okolo", "někdy"]
+
 
 sentence_splitter_group : 1
 
 monday:
     - Pon
+    - po
```
AdrianAtZyte, line 8: The YAML file only needs to add extra stuff not available in the CLDR. If you look at the Python code, you may notice that it already had `po`, and now it has it twice.

honzaHlavnicka: I have found them and deleted them form YAML file. With this change I noticed, that in relative-type exist 0 seconds ago for "now", so I moved all of "now" words from simplifications. Becouse of that, I had to change one test.

dateparser_data/supplementary_language_data/date_translation_data/cs.yaml
```diff
@@ -24,50 +28,199 @@ february:
     - únr
 june:
     - Čer
-july:
-    - Črc
-    - črv
```
AdrianAtZyte, line 29: Was this accidental? I wonder if we should keep `črc`.

honzaHlavnicka: Yes it was, although I have never saw anyone using this variant. Should I put it back?

AdrianAtZyte: Yes, please.

AdrianAtZyte: Now `črv` seems to be gone instead.

AdrianAtZyte (approved):

AdrianAtZyte:
Could you check the 2 test issues?

serhii73 (approved):

AdrianAtZyte:
Thanks!

PR #1330, opened 2026-05-07
fix: Preserve explicit signs for years when combined with decades

serhii73:
When both decades and years are present (e.g. '-1 decade +2 years'), the code in get_kwargs() unconditionally overwrote explicit_signs['years'] with the decades sign value:

    if 'decades' in explicit_signs:
        explicit_signs['years'] = explicit_signs['decades']

This destroyed the user-supplied +/- that was recorded for years during the PATTERN match loop, causing _parse_date() to mishandle the sign of the merged years value.

Fix: only propagate the decades sign to years when years did not carry its own explicit sign (explicit_signs.get('years') is False/absent).

Regression tests added for:
  - '-1 decade' (only decades, negative)
  - '-1 decade 2 years' (decades negative, years implicit)
  - '-1 decade +2 years' (decades negative, years explicitly positive)
  - '+2 years' (only years, explicitly positive future)
  - '1 decade +2 years' (decades implicit, years explicitly positive)
  - '+1 decade -2 years' (decades explicitly positive, years explicitly negative)

Fixes #1304

### DIFF
--- a/tests/test_freshness_date_parser.py
+++ b/tests/test_freshness_date_parser.py
@@ -70,7 +70,7 @@ class TestFreshnessDateDataParser(BaseTestCase):
             param("100 decades", ago={"years": 1000}, period="year"),
             # Regression tests for #1304: explicit signs preserved with decades + years
             param("-1 decade", ago={"years": 10}, period="year"),
-            param("-1 decade 2 years", ago={"years": 8}, period="year"),
+            param("-1 decade 2 years", ago={"years": 12}, period="year"),
             param("-1 decade +2 years", ago={"years": 8}, period="year"),
             param("yesterday", ago={"days": 1}, period="day"),
             param("the day before yesterday", ago={"days": 2}, period="day"),
@@ -1744,7 +1744,7 @@ class TestFreshnessDateDataParser(BaseTestCase):
             param("in a decade", in_future={"years": 10}, period="year"),
             # Regression tests for #1304: explicit signs preserved with decades + years
             param("+2 years", in_future={"years": 2}, period="year"),
-            param("1 decade +2 years", in_future={"years": 12}, period="year"),
+            param("in 1 decade +2 years", in_future={"years": 12}, period="year"),
             param("+1 decade -2 years", in_future={"years": 8}, period="year"),
             param("tomorrow", in_future={"days": 1}, period="day"),
             param("day after tomorrow", in_future={"days": 2}, period="day"),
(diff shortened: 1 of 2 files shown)


wRAR (approved):

tests/test_freshness_date_parser.py
```diff
@@ -68,6 +68,10 @@ def test_relative_past_dates_with_time_as_period(self, date_string, ago, period)
             param("last decade", ago={"years": 10}, period="year"),
             param("a decade ago", ago={"years": 10}, period="year"),
             param("100 decades", ago={"years": 1000}, period="year"),
+            # Regression tests for #1304: explicit signs preserved with decades + years
+            param("-1 decade", ago={"years": 10}, period="year"),
+            param("-1 decade 2 years", ago={"years": 8}, period="year"),
```
AdrianAtZyte: Shouldn’t the expectation be 12 here?

wRAR: I also wasn't sure about this.

tests/test_freshness_date_parser.py
```diff
@@ -68,10 +68,15 @@ def test_relative_past_dates_with_time_as_period(self, date_string, ago, period)
...
-            # Regression tests for #1304: explicit signs preserved with decades + years
+            # Regression tests for #1304: explicit signs preserved with decades + years.
+            # Each component's sign is resolved independently (explicit -> literal,
+            # unsigned -> ago/future context) before ``decades`` is folded into ``years``.
             param("-1 decade", ago={"years": 10}, period="year"),
             param("-1 decade 2 years", ago={"years": 12}, period="year"),
             param("-1 decade +2 years", ago={"years": 8}, period="year"),
+            param("-1 decade -2 years", ago={"years": 12}, period="year"),
+            param("1 decade +2 years", ago={"years": 8}, period="year"),
+            param("1 decade -2 years", ago={"years": 12}, period="year"),
```
AdrianAtZyte: I find some of these confusing.

I assume they are parsed the same if they end in "ago", so I will use "ago" at the end for clarify:

- "-1 decade -2 years ago" sounds like 12 years in the **future** to me.
- "1 decade +2 years ago" sounds like 12 years ago to me, not 8.
- "1 decade -2 years" sounds like 8 years ago to me, not 12.

I think the first and the last example here may be rather uncommon ways to display a date, so it may be OK to keep them off the tests and ignore how they end up being parsed.

"1 decade +2 years ago" sounds somewhat realistic to me, even if uncommon, and I would expect it to be interpreted as 12 years ago.

wRAR: > it may be OK to keep them off the tests and ignore how they end up being parsed.

I agree.

AdrianAtZyte (approved):

wRAR (approved):

PR #1331, opened 2026-05-09
Fix search_dates language fallback

### LINKED ISSUE #1326: search_dates() silently drops a French date when STRICT_PARSING=True and multiple languages are passed
"""
Bug: search_dates() silently drops a French date when STRICT_PARSING=True and multiple languages are passed,
     if the string ends with a bare uppercase token that is a valid ISO 639-1 language code (e.g. "FR").

dateparser version: 1.4.0
Python version: 3.13
"""

```python
import dateparser
import dateparser.search

print("dateparser version:", dateparser.__version__)

TEXT = "Date de facture 23 juillet 2020 Condition Redevable livraison FR"

# Works — single language, no ambiguity for the language detector
r1 = dateparser.search.search_dates(TEXT, languages=["fr"], settings={"STRICT_PARSING": True})

# Fails — same string, same settings, but 7 languages passed
r2 = dateparser.search.search_dates(TEXT, languages=["en", "fr", "es", "pt", "de", "it", "ar"], settings={"STRICT_PARSING": True})

print(f"languages=['fr'],    STRICT_PARSING=True  → {r1}")   # Expected: [('23 juillet 2020', datetime(...))]
print(f"languages=[7 langs], STRICT_PARSING=True  → {r2}")   # Bug:      None

assert r1 is not None, "Baseline broken"
assert r2 is not None, f"BUG: date not found when multiple languages passed — got {r2}"
```

Note that it works well when we remove `FR` in this input TEXT

It works also if we replace `FR` with `EN` or `US` or `Frn`

Kill1ngPig:
## Summary

Fix `search_dates()` so it can fall back to other explicitly provided languages when the initially selected language does not find any dates.

Fixes #1326.

---

## Problem

When `search_dates()` is called with multiple languages and `STRICT_PARSING=True`, the language selection step may choose a language that cannot parse the date expression. In that case, the current implementation stops after the first selected language and returns no dates, even if another explicitly provided language can parse the same text.

For example, this French date is parsed correctly with `languages=["fr"]`, but was previously dropped when several languages were provided:

```python
"Date de facture 23 juillet 2020 Condition Redevable livraison FR"
```

---

## Changes

- Keep the existing detected/selected language as the first parsing attempt.
- When multiple languages are explicitly provided, try the remaining candidate languages if the first one returns no dates.
- Return the first language that successfully finds dates.
- Add a regression test for the French `STRICT_PARSING=True` case from #1326.

This is not specific to French or to the `FR` suffix. It handles the broader case where the first selected language fails but another user-provided language can parse the date.

---

## Tests

```bash
python -m pytest -p no:cacheprovider tests/test_search.py tests/test_clean_api.py -q
```

Result:

```text
262 passed
```

### DIFF
--- a/dateparser/search/search.py
+++ b/dateparser/search/search.py
@@ -320,7 +320,9 @@ class DateSearchWithDetection:
             detect_languages_function=detect_languages_function,
         )
 
-        candidate_languages = self._get_candidate_languages(language_shortname, languages)
+        candidate_languages = self._get_candidate_languages(
+            language_shortname, languages
+        )
         if not candidate_languages:
             return {"Language": None, "Dates": None}
 
--- a/dateparser/search/search.py
+++ b/dateparser/search/search.py
@@ -218,6 +218,21 @@ class DateSearchWithDetection:
         self.available_language_map = self.loader.get_locale_map()
         self.search = _ExactLanguageSearch(self.loader)
 
+    def _get_candidate_languages(self, detected_language, languages):
+        candidates = []
+        if detected_language:
+            candidates.append(detected_language)
+
+        if isinstance(languages, (list, tuple, Set)) and len(languages) > 1:
+            candidates.extend(languages)
+
+        seen = set()
+        return [
+            language
+            for language in candidates
+            if not (language in seen or seen.add(language))
+        ]
+
     @apply_settings
     def detect_language(
         self, text, languages, settings=None, detect_languages_function=None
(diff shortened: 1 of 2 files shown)


AdrianAtZyte:
Please, run `pre-commit`.

Kill1ngPig:
Thanks for the reminder. I ran `pre-commit run --all-files` locally and pushed the formatting changes in e329bef.

AdrianAtZyte (approved):

serhii73 (approved):

serhii73:
Thanks!

PR #1332, opened 2026-05-31
Document region usage and stabilize current-month parsing

### LINKED ISSUE #465: No apparent way to get a list of accepted regions for use with the region argument for parse.  If there is it should be in the docs.  Also I went through the code and still not clear what exactly a region even is.  IE US works but ID does not.

tine1117:
## Summary

- Document how the `region` argument maps to locale subtags listed on the supported locales page, with `en-IN` and `fr-CA` examples.
- Reuse the parser's current month when applying `PREFER_MONTH_OF_YEAR="current"` to incomplete dates, so current month and current day come from the same relative base.
- Add a regression test for the month-boundary case where UTC is still May 31 but the local CI timezone is already June 1.

Fixes #465.

## Test

- `python -m sphinx -b html docs docs\_build\html -q`
- `python -m pytest tests/test_search.py -q`
- `python -m pytest tests/test_date_parser.py -k "dates_with_no_day_or_month" -q`
- `python -m ruff check dateparser/parser.py tests/test_date_parser.py`
- `python -m ruff format --check dateparser/parser.py tests/test_date_parser.py`

### DIFF
--- a/dateparser/parser.py
+++ b/dateparser/parser.py
@@ -598,16 +598,11 @@ class _parser:
         return dateobj
 
     def _correct_for_month(self, dateobj):
-        relative_base = getattr(self.settings, "RELATIVE_BASE", None)
-        relative_base_month = (
-            relative_base.month if hasattr(relative_base, "month") else relative_base
-        )
-
         if getattr(self, "_token_month", None):
             return dateobj
 
         dateobj = set_correct_month_from_settings(
-            dateobj, self.settings, relative_base_month
+            dateobj, self.settings, current_month=self.now.month
         )
         return dateobj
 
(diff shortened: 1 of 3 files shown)


tine1117:
I looked into the failed Build run. The docs job had passed; the failure was limited to the Python 3.13 jobs running under Pacific/Auckland and Pacific/Fiji. Those jobs crossed a month boundary relative to UTC, so incomplete dates such as `2014` were using the parser's UTC day but the local CI month. That could produce an invalid intermediate date like June 31 and fall back to December 31.

I pushed a small follow-up commit to keep the current month and day on the same parser-relative base and added a regression test for that case. The new Actions runs are currently in `action_required`, so they need maintainer approval before they can execute.

serhii73 (approved):

serhii73:
Thank you!

PR #1335, opened 2026-06-06
make relative-date regex digit quantifiers possessive

alhudz:
The relative-date matchers are built from `\d+[.,]?\d*` and run across the whole input: `PATTERN` in `freshness_date_parser` via `findall`/`sub`, and the `split_relative` regex in `dictionary.py` via `split`. On a long run of digits the two adjacent digit quantifiers backtrack without ever changing the match, so cost grows superlinearly.

Repro: `dateparser.parse('9' * 3200)`, or any long digit run reaching the relative parser.
Cause: the expression is matched at every position; splitting `\d+` against `\d*` over a digit run backtracks quadratically in `split`, near-cubically for the freshness `findall`.
Fix: make the digit quantifiers possessive (`\d++`/`\d*+`). A digit run can only be followed by `[.,]` or the literal unit text, so giving digits back can never complete a match.

`PATTERN.findall('9' * 3200)` goes from ~23s to ~0.02s; match results are unchanged and the existing suite passes.


serhii73 (approved):

serhii73:
Thanks!

PR #1339, opened 2026-06-20
Reject non-positive date_range steps

CodingFeng101:
## Summary

Reject `date_range()` steps that do not advance the generated date. Passing values such as `months=0` or `days=-1` currently allows the generator to yield the start date and then continue without making forward progress, which can hang callers that consume the full iterator with `list(...)`.

## Changes

- Add a positive-step validation in `dateparser.date.date_range()`.
- Add a focused regression test for zero-month and negative-day steps.

## Validation

- `python -m pytest tests/test_date_range.py -q` -> `2 passed`
- Manual smoke check confirmed normal daily and monthly ranges still produce the same forward dates.


AdrianAtZyte:
Does Python’s `range` implement any kind of validation like this, or does it leave it up to the user to pass the right value?

CodingFeng101:
Thanks for checking. Python's `range()` validates the one case that cannot make progress: `range(..., step=0)` raises `ValueError: range() arg 3 must not be zero`. It allows negative steps because `range()` has explicit backward-iteration semantics when the caller provides matching `start`/`stop` values.

My thinking here was that `date_range()` currently advances from `begin` toward `end`, and non-positive deltas such as `months=0` or `days=-1` can make the generator fail to progress toward that end in common calls like `list(date_range(begin, end, months=0))`. So the validation is meant to fail early for non-progressing steps instead of leaving callers with an accidental infinite iterator.

That said, if you prefer to leave this to callers or want an issue first before changing this behavior, I understand and am happy to close the PR.

tests/test_date_range.py
AdrianAtZyte, line 1: Could you move this to the existing TestDateRangeFunction?

CodingFeng101: Moved it into the existing TestDateRangeFunction in 4cbea3a and removed the separate test file.

AdrianAtZyte (approved):

AdrianAtZyte:
Please, run `pre-commit run --all-files`.

CodingFeng101:
Ran `pre-commit run --all-files`. `ruff-format` reformatted the new date range test in `1150f2f`, and a second run now passes:

```text
ruff.....................................................................Passed
ruff-format..............................................................Passed
```

I also ran `git diff --check`, which passed. I tried rerunning `python -m pytest tests/test_date.py::TestDateRangeFunction -q` locally, but on Windows collection fails before reaching the test because `tests/test_date.py` imports `time.tzset`, which is not available on Windows.

PR #1341, opened 2026-06-20
Reject negative cache size limits

CodingFeng101:
## Summary

Reject negative `CACHE_SIZE_LIMIT` values during settings validation. The setting is used as a cache-size limit, but negative integers are currently accepted and make the cache eviction condition true immediately after entries are added.

## Changes

- Add a non-negative validation check for `CACHE_SIZE_LIMIT`.
- Add a focused regression test for negative values.
- Keep the existing behavior for `0` and positive limits unchanged.

## Validation

- `python -m pytest tests/test_cache_size_limit.py -q` -> `1 passed`
- Manual smoke check confirmed `CACHE_SIZE_LIMIT` values `0` and `1` are still accepted.

### DIFF
--- a/dateparser/conf.py
+++ b/dateparser/conf.py
@@ -175,6 +175,11 @@ def _check_between_0_and_1(setting_name, setting_value):
         )
 
 
+def _check_non_negative(setting_name, setting_value):
+    if setting_value < 0:
+        raise SettingValidationError("{} cannot be negative".format(setting_name))
+
+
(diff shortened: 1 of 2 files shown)


AdrianAtZyte:
I don’t see how a user could do this accidentally, and it does not sound like a big issue if they do, so I am not sure this is worth it.

Also, please don’t spam PRs, specially PRs without issues confirming what they address needs to be addressed.

PR #1343, opened 2026-06-24
Update CLDR data to 44.1.0 and add hour/duration support

serhii73:
## Summary

- Upgrades CLDR source data from **31.0.1 → 44.1.0**: 288 new locales added, all existing locale JSONs updated
- Adds **hour/duration unit** to `relative-type` for all locales (e.g. "2 hours ago" now parses correctly in more languages)
- Regenerates all `dateparser/data/date_translation_data/*.py` files from the new source data
- Modernizes `dateparser_scripts/` to use `pathlib.Path` and the unified `cldr-json` repository layout
- Restores Ukrainian words removed by the CLDR upgrade (`uk.yaml`)
- Updates `docs/supported_locales.rst` and `languages_info.py`
- Updates 34 test inputs to reflect CLDR 44.1.0 abbreviation/name changes
- Adds `/cldr-json/` to `.gitignore`

Finishes #1216 (by @Gallaecio). The original PR was a draft with merge conflicts; this re-implements the same work cleanly on top of current master, preserving the possessive-quantifier optimization from #1335.

## Test plan

- [x] All 1376 existing tests pass (`python -m pytest tests/test_languages.py`)
- [x] Test inputs updated where CLDR 44.1.0 changed abbreviations (bs-Latn, ce, kl, qu, so, sr, sw, zu, am, as, brx, hy, ig, kok, mr, nn, de, eu, gu, mk, chr, bs-Cyrl)
- [x] Basic smoke test: `dateparser.parse("2 hours ago")` and `dateparser.parse("vor 2 Stunden")` both return valid dates

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1349, opened 2026-07-03
Add 'w' as an English abbreviation for 'week' (#1257)

### LINKED ISSUE #1257: "1w ago" not supported
When I tried to parse the datetime object from '1w ago,' dateparser returned None. However, it worked fine with '1week ago'

Image

apoorvdarshan:
Fixes #1257.

## Problem

`"1w ago"` returns `None`, even though the full word and every other single-letter unit abbreviation work:

```python
>>> dateparser.parse("1w ago")     # None
>>> dateparser.parse("1week ago")  # works
>>> dateparser.parse("1d ago")     # works (day)
>>> dateparser.parse("1h ago")     # works (hour)
>>> dateparser.parse("1m ago")     # works (minute)
>>> dateparser.parse("1s ago")     # works (second)
>>> dateparser.parse("1y ago")     # works (year)
```

Every relative-time unit had a single-letter abbreviation **except week**: in the English translation data, `day` has `d`, `hour` has `h`, `minute` has `m`, `second` has `s`, `year` has `y`, but `week` only had `week`/`wk`/`weeks`.

## Fix

Add `w` to the English `week` translation data in `dateparser_data/supplementary_language_data/date_translation_data/en.yaml` and regenerate `dateparser/data/date_translation_data/en.py` with `dateparser_scripts/write_complete_data.py`.

`month` intentionally keeps no single-letter form (since `m` already means minute), but `w` is unambiguous, so this fills a genuine gap rather than introducing a conflict.

```python
>>> dateparser.parse("1w ago")   # 7 days ago
>>> dateparser.parse("2w ago")   # 14 days ago
>>> dateparser.parse("in 2w")    # 14 days in the future
```

## Testing

- Added `"1w ago"` / `"2w ago"` cases to `tests/test_freshness_date_parser.py` (next to the existing `"2y ago"` abbreviation case). They fail on `main` and pass with this change.
- Full `tests/test_freshness_date_parser.py` (1057) and `tests/test_search.py` (244) pass — no regressions.
- `ruff check` / `ruff format --check` clean on the changed files.

---

*Disclosure: this change was prepared with the assistance of an AI tool (Claude Code). I reproduced the issue, made the data change and regenerated `en.py` with the project's own script, verified the fix and the test run, and take responsibility for the contribution and will respond to review feedback personally.*

### DIFF
--- a/dateparser/data/date_translation_data/en.py
+++ b/dateparser/data/date_translation_data/en.py
@@ -105,6 +105,7 @@ info = {
     "week": [
         "week",
         "wk",
+        "w",
         "weeks"
     ],
     "day": [
--- a/dateparser_data/supplementary_language_data/date_translation_data/en.yaml
+++ b/dateparser_data/supplementary_language_data/date_translation_data/en.yaml
@@ -37,6 +37,7 @@ year:
 month:
     - months
 week:
+    - w
     - weeks
 day:
     - d
--- a/tests/test_freshness_date_parser.py
+++ b/tests/test_freshness_date_parser.py
@@ -135,6 +135,9 @@ class TestFreshnessDateDataParser(BaseTestCase):
             param("six days ago", ago={"days": 6}, period="day"),
             param("five years ago", ago={"years": 5}, period="year"),
             param("2y ago", ago={"years": 2}, period="year"),
+            # Single-letter week abbreviation (#1257)
+            param("1w ago", ago={"weeks": 1}, period="week"),
+            param("2w ago", ago={"weeks": 2}, period="week"),
             # Fractional units
             param("2.5 hours", ago={"hours": 2.5}, period="day"),
             param("10.75 minutes", ago={"minutes": 10.75}, period="day"),


AdrianAtZyte (approved):

serhii73:
Thanks for this, @apoorvdarshan, and sorry it sat for a while. The ground moved under the PR: #1257 was fixed independently by the CLDR 44.1.0 data update (#1343), which shipped in 1.4.2 and added the narrow relative-time patterns `(\d+)w ago` and `in (\d+)w` to the English `relative-type-regex` data. On current master:

```pycon
>>> s = {"RELATIVE_BASE": datetime(2026, 1, 15, 12, 0)}
>>> dateparser.parse("1w ago", languages=["en"], settings=s)
datetime.datetime(2026, 1, 8, 12, 0)
>>> dateparser.parse("2w ago", languages=["en"], settings=s)
datetime.datetime(2026, 1, 1, 12, 0)
>>> dateparser.parse("in 2w", languages=["en"], settings=s)
datetime.datetime(2026, 1, 29, 12, 0)
```

I applied this branch's data change to a copy of master and diffed the output for those three strings across 13 settings combinations, 6 language sets and the `en` / `en-US` / `en-GB` / `en-001` locales: byte-identical to master. The two params the PR adds to the freshness tests also pass on unpatched master, so they don't exercise the change.

What the change still adds is the *word-list* forms — `1 w ago`, bare `1w`, `1w2d ago`, `1w before` — and that's where it runs into trouble. Making a bare `w` an English date token stops `search_dates` from using a lone `W` as a split point:

```pycon
# master
>>> search_dates("50 W 34th St on 3 May 2021", languages=["en"])
[('34th St on 3 May 2021', datetime.datetime(2021, 5, 3, 0, 0))]
# with this PR
>>> search_dates("50 W 34th St on 3 May 2021", languages=["en"])
None
```

A correct date silently disappears on a very ordinary US address. Plain `parse()` gains false positives too — watts, compass bearings, route numbers, win/loss records:

```pycon
>>> dateparser.parse("60 W", languages=["en"], settings=s)                # None on master
datetime.datetime(2024, 11, 21, 12, 0)
>>> dateparser.parse("2021-01-01 12:00 W", languages=["en"], settings=s)  # None on master
datetime.datetime(2026, 1, 15, 12, 0)   # the real date is discarded
```

14 of 20 realistic English sentences I tested changed behaviour, 13 for the worse. The full suite stays green (24186 passed) because nothing covers a standalone `W` in prose — so CI passing isn't evidence of safety here.

Since the reported issue is already fixed and the remaining delta carries that regression, I'm closing this. If someone wants the spaced forms (`1 w ago`, `in 2 w`) specifically, the safer shape is a pair of `relative-type-regex` entries in `en.yaml` instead of a word-list token: that keeps `w` anchored to an adjacent number plus `ago`/`in`, and leaves `search_dates` alone.

For what it's worth, the mechanics of your diff were right — I re-ran `python -m dateparser_scripts.write_complete_data` with your `en.yaml` change on current master and it reproduces your `en.py` hunk character for character. The gap you identified in the word list is real; it's just that closing it that way costs more than it gains. Thanks for the careful write-up.

PR #1350, opened 2026-07-19
Honor PREFER_DATES_FROM for a date_formats entry with no year

chuenchen309:
`parse_with_formats` applied `PREFER_DATES_FROM` only to a two-digit-year format
(through `_apply_century_preference`, added in #1342). A `date_formats` entry
with no year directive at all took the current year unconditionally:

```python
import dateparser
from datetime import datetime
RB = datetime(2026, 6, 15)

dateparser.parse("December 25", date_formats=["%B %d"],
                 settings={"RELATIVE_BASE": RB, "PREFER_DATES_FROM": "past"})
# -> 2026-12-25   (a future date, under PREFER_DATES_FROM="past")
```

The result contradicts the setting, and disagrees with the NLP path, which
parses the same string+settings correctly to `2025-12-25`. `docs/settings.rst`
states `PREFER_DATES_FROM` applies "if date string is missing some part" (its
example omits the year), with no carve-out for `date_formats`. This is the
other half of #1342 — that PR wired the setting into the adjacent two-digit-year
branch but left the year-less branch.

Fix: generalise the `#1342` helper to shift by ±N years and call it with `1`
for the year-less branch (`100` stays for the two-digit branch), so both share
the same direction logic and Feb-29 leap-year handling. Added a regression test.

---
This PR was authored by an AI coding agent (Claude Code) running on this account:
the AI found the bug, ran the repro, wrote the test, and wrote this description.
The human account holder reviews every change and is accountable for it. The
verification is real and re-runnable from the diff. If this isn't the kind of
contribution you want, say so and I'll close it.

### DIFF
--- a/dateparser/date.py
+++ b/dateparser/date.py
@@ -177,22 +177,23 @@ def get_date_from_timestamp(date_string, settings, negative=False):
         return date_obj
 
 
-def _apply_century_preference(date_obj, now, prefer_from):
-    """Shift *date_obj* by ±100 years to satisfy PREFER_DATES_FROM.
+def _apply_year_preference(date_obj, now, prefer_from, years):
+    """Shift *date_obj* by ±*years* to satisfy PREFER_DATES_FROM.
 
-    *now* is normalised to naive so a tz-aware RELATIVE_BASE does not raise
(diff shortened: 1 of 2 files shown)


AdrianAtZyte:
Ask your human to explain the fix to me in their words.

PR #1351, opened 2026-07-20
Add "ngram" search strategy to search_dates

serhii73:
The new strategy tokenizes the text on characters that never occur inside a date expression and tries to parse token n-grams, longest first, so that the most complete date is found. Compared to the default translation-based "split" strategy it produces more predictable results on noisy real-world text.

The strategy is exposed through the new "strategy" parameter of search_dates and defaults to the existing behavior.

### DIFF
--- a/dateparser/search/__init__.py
+++ b/dateparser/search/__init__.py
@@ -9,6 +9,7 @@ def search_dates(
     settings=None,
     add_detected_language=False,
     detect_languages_function=None,
+    strategy="split",
 ):
     """Find all substrings of the given string which represent date and/or time and parse them.
 
@@ -34,6 +35,13 @@ def search_dates(
         and returns a list of detected language codes.
         Note: detect_languages_function is only uses if `languages` are not provided.
     :type detect_languages_function: function
+
+    :param strategy:
+        The search strategy to use: "split" (default) translates the text and splits it
+        into chunks that are likely to contain dates, while "ngram" tries to parse the
+        longest possible sequences of tokens as dates. The "ngram" strategy tends to
+        produce more predictable results, at the cost of more parse attempts.
+    :type strategy: str
 
     :return: Returns list of tuples containing:
         substrings representing date and/or time, corresponding :mod:`datetime.datetime`
@@ -64,6 +72,7 @@ def search_dates(
         languages=languages,
         settings=settings,
         detect_languages_function=detect_languages_function,
+        strategy=strategy,
     )
     dates = result.get("Dates")
     if dates:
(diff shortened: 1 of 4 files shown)


serhii73:
## Implementation summary

This PR adds an opt-in **`"ngram"` strategy** to `search_dates`, adapting a date-search algorithm from a production date extraction pipeline where it has consistently produced more predictable results than the split-based search on noisy real-world text. The default behavior is unchanged (`strategy="split"`).

### How the ngram strategy works

For example, in `"Posted on 4 October 1957 by admin"`:

1. **Tokenize** the text on characters that never occur inside a date expression (whitespace and `,|()@`), keeping each token's position in the original text.
2. **Drop noise words** (`on`, `at`, `of`, `a`) that may appear between date components.
3. Scan tokens left to right, trying **n-grams of up to 7 tokens, longest first** — `4 October 1957 by admin` ✗ → … → `4 October 1957` ✓ — so the *most complete* date phrase wins over fragments like `October 1957`.
4. **Reject blacklisted candidates** that parse but are almost always false positives: bare 1–3 digit numbers, `#123`-style refs, separator runs, single letters.
5. Parse candidates with `DateDataParser(languages=..., use_given_order=True)`; after a match, scanning resumes past it, so all non-overlapping dates are returned.
6. Return `(substring, datetime)` pairs, with substrings taken from the original text (punctuation-stripped), matching the existing `search_dates` contract.

### Changes

- New module `dateparser/search/ngram_search.py` with `_NgramDateSearch` implementing the algorithm above.
- New `strategy` parameter on the public `search_dates()` and `DateSearchWithDetection.search_dates()`: `"split"` (default) or `"ngram"`; invalid values raise `ValueError`.
- Language detection and candidate-language resolution are shared between both strategies; the ngram path passes the full candidate-language list to the parser in given order and falls back to `settings.DEFAULT_LANGUAGES` like the split path.
- `RETURN_TIME_SPAN` post-processing extracted into a shared module-level `_add_time_span_results()` helper, so both strategies have feature parity.
- 12 new tests in `TestNgramSearch`: en/fr/ru extraction, noise-word/punctuation handling, blacklist, `RELATIVE_BASE`, `RETURN_TIME_SPAN`, module-level passthrough, `add_detected_language`, `None` when no dates found, and strategy validation.

### Verification

- `tests/test_search.py`: **256 passed**
- Full suite (`pytest tests`): **24072 passed, 20 skipped, 1 xfailed**
- `ruff check` and `ruff format`: clean

### Usage

```python
from dateparser.search import search_dates

search_dates("Posted on 4 October 1957 by admin", strategy="ngram")
# [('4 October 1957', datetime.datetime(1957, 10, 4, 0, 0))]
```

AdrianAtZyte (approved):

PR #1359, opened 2026-08-03
Fix raw-HTML pattern in README parsing

EpicWink:
Fixes the issue introduced by #1311 where all of the README's contents are not in the [published package's long-description](https://pypi.org/project/dateparser/1.4.1/).

Also explicitly specify README as RST.

Explanation:

* Removing `s` flag - so dots `.` don't match newlines. This made the capturing group too greedy
* Making the group's quantifier `*` not greedy, so the matched text ends as soon as the rest of the pattern can be satisfied
* Specify end delimiter `\n\n` - this ends all RST blocks, so we match (and remove) that delimiter
* Specifying RST as the content-type is good practice, even if PyPI assumes RST

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -41,7 +41,7 @@
 
 
 Key Features
-------------
+============
 
 -  Support for almost every existing date format: absolute dates,
    relative dates (``"two weeks ago"`` or ``"tomorrow"``), timestamps,
@@ -60,7 +60,7 @@ Key Features
 -  Time span detection for expressions like "past month", "last week".
 
 Online demo
------------
+===========
 
 Do you want to try it out without installing any dependency? Now you can test
 it quickly by visiting `this online demo <https://gnyman.github.io/dateparser-demo-app/>`__!
@@ -68,7 +68,7 @@ it quickly by visiting `this online demo <https://gnyman.github.io/dateparser-de
 
 
 How To Use
-----------
+==========
 
 The most straightforward way to parse dates with **dateparser** is to
 use the ``dateparser.parse()`` function, that wraps around most of the
@@ -132,7 +132,7 @@ If you know the possible formats of the dates, you can use the
     datetime.datetime(2010, 12, 22, 0, 0)
 
 Relative Dates
-^^^^^^^^^^^^^^
+--------------
 
 .. code:: python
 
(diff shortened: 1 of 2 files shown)


EpicWink:
Failing build is due to mismatch between README and changelog (aka HISTORY) regarding section title underlines. There's two options to fix this:

1. Add a new top-level section title to the README, with equal-sign `=` underlining [^1]. This would demote the heading levels of all current titles in the README
2. Change the README's section titles' underlines to match HISTORY (ie top-level from hyphen `-` to equal-sign `=`, second-level from carat `^` to hyphen `-`)

[^1]: You could convert the new bold text in #1360 to this heading

AdrianAtZyte:
I am fine with either.

EpicWink:
I chose the second option, to have the formatted README look the same

AdrianAtZyte (approved):

serhii73:
Flagging that this fixes a regression that is live on PyPI right now, in case it helps prioritise the merge.

On master, the `(?ms)` flag in `setup.py:7` makes `.` match newlines, so `^[ \t].*\n` greedily consumes everything from the first `.. raw:: html` (README.rst:3) to the end of the file:

```
README.rst:                    15174 bytes
after the current (?ms) regex:   165 bytes   (only the leading comment survives)
after this PR's (?m) regex:    13787 bytes
```

The result is visible at https://pypi.org/project/dateparser/ — the project description currently contains no README at all, just the changelog. Confirmed against the published metadata for both affected releases:

```console
$ curl -s https://pypi.org/pypi/dateparser/1.4.2/json | jq -r '.info.description' | head -c 120
.. Note that we use raw HTML in the header section because centering images ...
.. :changelog:

History
```

"Key Features", "How To Use", "Installation" and "Supported Calendars" are all absent from 1.4.1 and 1.4.2. 1.4.0 is unaffected because its long description came from `docs/introduction.rst`, before #1311 switched it to README.rst.

I also confirmed the heading rework in this PR is required rather than cosmetic: with only the regex fixed, `twine check` fails with `Inconsistent title style: skip from level 1 to 3`, because README.rst used `-` at its top level while HISTORY.rst uses `=`. With the full PR applied, `twine check --strict` passes and PKG-INFO carries the complete README.

Worth noting the fix only reaches pypi.org on the next upload, so it would be good to land this before the next release rather than after it.

serhii73 (approved):

PR #1337, opened 2026-06-15
1.4.1 release

serhii73:
Release notes and version bump for **dateparser 1.4.1**, following the same two-step approach used for 1.4.0 (#1319 + the subsequent bump commit):

1. **`1.4.1 release notes`** — add the `1.4.1` section to `HISTORY.rst`, summarizing every PR merged since `v1.4.0`.
2. **`Bump version: 1.4.0 → 1.4.1`** — produced by `bump-my-version bump patch`, updating `pyproject.toml`, `dateparser/__init__.py`, and stamping the `HISTORY.rst` release date. `SECURITY.md` is unchanged because it tracks only `major.minor` (still `1.4.x`).

### Changelog

**Breaking changes:**
- Remove fastText language detection support; the `fasttext` extra is dropped and `detect_languages()` now raises `ImportError`. Migrate to the `langdetect` extra, which also unblocks `numpy` 2.x compatibility (#1315)

**Security fixes:**
- Make digit quantifiers possessive in the relative-date regexes to prevent quadratic backtracking (ReDoS) on long digit runs (#1335)

**New features:**
- Add the `USE_GIVEN_LANGUAGE_ORDER` setting to try `languages` and `locales` in the order given rather than by frequency (#789)

**Fixes:**
- Preserve explicit signs on individual components when parsing relative dates that combine decades with years (#1330)
- Fall back to other provided languages in `search_dates` when the detected language yields no dates (#1331)
- Parse relative date expressions with spaces between the sign and number, e.g. `now - 2 hours` (#1327)
- Use the parser-relative `now` for the current month when filling in incomplete dates (#1332)
- Fix Norwegian Bokmål (`nb`) parsing of relative date expressions (#1334)
- Parse abbreviated English month expressions such as `1mon ago` (#1329)
- Preserve surrounding whitespace when removing skip tokens during translation (#1324)

**Improvements:**
- Move project metadata and build configuration to `pyproject.toml` (#1311)
- Add alternative Korean date expressions (#1289)
- Expand Czech date translations (#1325)
- Replace internal `OrderedDict` usage with the built-in `dict` (#1328)

### After merge

The `v1.4.1` git tag is intentionally **not** included in this branch (a tag on a pre-merge branch would point at the wrong commit). After merging, tag the merge commit on `master`:

```
git tag v1.4.1 && git push origin v1.4.1
```

Internal-only changes (#1333 data-generation script cleanup, #1320 Read the Docs CI bump) were deliberately omitted from the changelog.

[...]

PR #1361, opened 2026-08-04
1.4.2 release

serhii73:
Release notes and version bump for **dateparser 1.4.2**, following the same two-step approach used for 1.4.1 (#1337):

1. **`1.4.2 release notes`** — add the `1.4.2` section to `HISTORY.rst`, summarizing every PR merged since `v1.4.1`.
2. **`Bump version: 1.4.1 → 1.4.2`** — produced by `bump-my-version bump patch`, updating `pyproject.toml`, `dateparser/__init__.py`, and stamping the `HISTORY.rst` release date. `SECURITY.md` is unchanged because it tracks only `major.minor` (still `1.4.x`).

### Changelog

**New features:**
- Add an `IGNORE_SURROUNDING_TEXT` setting that, when enabled, retries parsing after ignoring the leading and trailing words the language does not recognize, so a date wrapped in extra text such as `Published on 16/04/2019` is parsed (#1356)
- Add a `strategy` argument to `search_dates()` to choose the search strategy: the default `"split"` keeps the current behavior, while the new `"ngram"` strategy parses the longest sequences of tokens as dates for more predictable results on noisy text (#1351)

**Fixes:**
- Do not read a two-digit number as a year once a later component has been found in year-first date orders, so the day in Japanese dates such as `4月20日` is no longer consumed as the year (#1358)
- Parse ISO 8601 dates (those starting with a four-digit year) in month-day order even when an explicit day-first language such as `it` or `fr` is given, without needing to set `PREFER_LOCALE_DATE_ORDER` to `False` (#1352)
- Honor `PREFER_DATES_FROM` and `RELATIVE_BASE` when parsing with custom `date_formats` that use a two-digit year (`%y`) (#1342)
- Honor the `%j` (day of year) directive in `date_formats` instead of overwriting the parsed month and day with the current date (#1345)

**Improvements:**
- Update the bundled CLDR locale data to 44.1.0, adding many newly recognized date forms across locales and a standalone hour/duration unit (e.g. Catalan `2 hores`), while keeping previously supported forms parseable (#1343)
- Add the Italian expressions `un ora fa` and `un'ora fa`, and skip `alle` so phrases like `oggi alle 11:00` parse (#1049)
- Expand Czech date translations with month locative/dative forms (e.g. `v lednu 2023`), the July abbreviation `črv`, and relative expressions such as `za týden`, `za měsíc` and `za rok` (#1172)
- Specify the README header content as RST rather than raw HTML (#1360)

### Validation

- `bump-my-version bump patch` produced a clean two-commit diff touching only `HISTORY.rst`, `pyproject.toml`, and `dateparser/__init__.py`; version is consistent across all three.
- `python -m build` + `twine check` pass for both the sdist and the wheel.
- Test suite green over the areas changed this cycle (`test_search`, `test_clean_api`, `test_date`, `test_date_parser`, `test_languages`, `test_timezone_parser`).

### After merge

The `v1.4.2` git tag is intentionally **not** included in this branch (a tag on a pre-merge branch would point at the wrong commit). After merging, tag the merge commit on `master`:

```
git tag v1.4.2 && git push origin v1.4.2
```

All PRs merged since `v1.4.1` are included in the changelog above; there were no internal-only changes to omit this cycle.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,47 @@
 History
 =======
 
+1.4.2 (unreleased)
+------------------
+
+New features:
+
+- Add an ``IGNORE_SURROUNDING_TEXT`` setting that, when enabled, retries
+  parsing after ignoring the leading and trailing words the language does
+  not recognize, so a date wrapped in extra text such as "Published on
+  16/04/2019" is parsed (#1356)
+- Add a ``strategy`` argument to ``search_dates()`` to choose the search
+  strategy: the default ``"split"`` keeps the current behavior, while the
+  new ``"ngram"`` strategy parses the longest sequences of tokens as dates
+  for more predictable results on noisy text (#1351)
+
+Fixes:
+
+- Do not read a two-digit number as a year once a later component has been
+  found in year-first date orders, so the day in Japanese dates such as
+  "4月20日" is no longer consumed as the year (#1358)
+- Parse ISO 8601 dates (those starting with a four-digit year) in month-day
+  order even when an explicit day-first language such as ``it`` or ``fr`` is
+  given, without needing to set ``PREFER_LOCALE_DATE_ORDER`` to ``False``
+  (#1352)
+- Honor ``PREFER_DATES_FROM`` and ``RELATIVE_BASE`` when parsing with custom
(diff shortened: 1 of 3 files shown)


AdrianAtZyte (approved):
🚀

PR #1363, opened 2026-08-05
Fix turn-of-year not recognized in search_dates date ranges

yamilmaud:
## Summary
- `search_dates` splits date ranges like `"Closed from 23th December until 8th January."` into separate substrings and parses each independently, so a range spanning the new year had both dates land in the same calendar year (the second date should roll over to the next year).
- Adds `_ExactLanguageSearch._adjust_year_for_rollovers`, which detects consecutive parsed results without an explicit year, joined by a **range connector** (`until`, `to`, `through`, `till`, `-`/`–`/`—`) in the original text, where the later date's month comes chronologically before the earlier date's month — and increments its year by one.
- Explicit years in either substring (e.g. a 4-digit year already present) are left untouched.

### Why a connector check instead of a bare month comparison
An earlier version of this fix used `curr.month < prev.month and (curr.month - prev.month) % 12 < 6` — a month-gap heuristic. That silently failed to roll over legitimate wide-gap ranges (e.g. `"from 1st September until 15th March"`, a 6-month forward gap), and widening or removing the threshold instead caused false positives: unrelated standalone date mentions in running text that merely happen to go backwards in month (e.g. a paragraph mentioning "June ..." then later "May ...", meant to be read within the same year) would get incorrectly bumped to next year.

Checking for an actual range connector between the two date substrings (rather than guessing from the month gap alone) correctly distinguishes a real range from coincidental backwards-in-month mentions.

### Known scope limits
- The connector word list (`until`/`to`/`through`/`till`/dash) is **English-only**. Non-English ranges fall back to no rollover (safe, but not fixed) rather than guessing — extending this per-locale is future work.
- If irregular whitespace prevents a parsed substring from being located verbatim in the original text, the rollover is safely skipped rather than guessing (covered by `test_search_dates_rollover_skipped_when_substring_position_not_found`).
- For chains of 3+ dates, the rollover only propagates pairwise (each date is compared only to its immediate predecessor).

Fixes the reported issue: `"Closed from 23th December until 8th January."` now parses to `(this_year, 12, 23)` and `(this_year + 1, 1, 8)` instead of both falling in the same year.

## Test plan
- [x] Added tests to `tests/test_search.py`:
  - `test_search_dates_with_prepositions_with_turn_of_year` (the originally reported case)
  - `test_search_dates_turn_of_year_wide_cross_year_range` (3-month gap)
  - `test_search_dates_turn_of_year_beyond_six_months` (6-month gap, previously broken)
  - `test_search_dates_no_rollover_without_range_connector` (standalone mentions must not roll over)
  - `test_search_dates_rollover_skipped_when_substring_position_not_found` (safe fallback on irregular whitespace)
- [x] `pytest tests/` — 24023 passed, 6 skipped, 1 xfailed
- [x] `ruff check` / `ruff format --check` clean on changed files
- [x] Patch coverage: 100% of newly added lines covered

### DIFF
--- a/dateparser/search/search.py
+++ b/dateparser/search/search.py
@@ -44,6 +44,38 @@ class _ExactLanguageSearch:
         relative_base = already_parsed[i][0]["date_obj"]
         return substring, relative_base
 
+    @staticmethod
+    def _has_explicit_year(substring):
+        """Return True if the substring contains a 4-digit year number."""
+        return bool(re.search(r"\b\d{4}\b", substring))
+
+    @staticmethod
+    def _adjust_year_for_rollovers(results):
+        """Fix year-spanning date ranges parsed without explicit years.
+
+        When consecutive parsed dates go backwards chronologically (e.g.
+        "from 23th December until 8th January") but neither substring contains
+        an explicit year, the later-mentioned date belongs to the next calendar
+        year and its year is incremented by one.
+        """
+        if len(results) < 2:
+            return results
+        adjusted = list(results)
+        for i in range(1, len(adjusted)):
+            prev_text, prev_date = adjusted[i - 1]
+            curr_text, curr_date = adjusted[i]
+            if prev_date is None or curr_date is None:
+                continue
+            if _ExactLanguageSearch._has_explicit_year(
+                prev_text
+            ) or _ExactLanguageSearch._has_explicit_year(curr_text):
+                continue
+            if curr_date.month < prev_date.month and (
+                curr_date.month - prev_date.month
+            ) % 12 < 6:
+                adjusted[i] = (curr_text, curr_date.replace(year=curr_date.year + 1))
+        return adjusted
+
     def choose_best_split(self, possible_parsed_splits, possible_substrings_splits):
         rating = []
         for i in range(len(possible_parsed_splits)):
@@ -189,6 +221,7 @@ class _ExactLanguageSearch:
         )
 
         results = list(zip(substrings, [i[0]["date_obj"] for i in parsed]))
+        results = self._adjust_year_for_rollovers(results)
 
         if getattr(settings, "RETURN_TIME_SPAN", False):
             span_info = detect_time_span(text)
--- a/dateparser/search/search.py
+++ b/dateparser/search/search.py
@@ -49,18 +49,46 @@ class _ExactLanguageSearch:
         """Return True if the substring contains a 4-digit year number."""
         return bool(re.search(r"\b\d{4}\b", substring))
 
+    # Text joining two dates that reads as one range rather than two
+    # unrelated mentions (e.g. "from 23th December UNTIL 8th January").
+    # English-only: non-English ranges fall back to no rollover instead of
+    # guessing, since translating this word list per-locale is out of scope.
+    _RANGE_CONNECTOR_REG = re.compile(
+        r"^[\s,]*(?:until|to|through|till|-|–|—)[\s,]*$", re.IGNORECASE
+    )
+
     @staticmethod
-    def _adjust_year_for_rollovers(results):
+    def _is_range_connector(connector):
+        return bool(_ExactLanguageSearch._RANGE_CONNECTOR_REG.match(connector))
+
+    @staticmethod
+    def _adjust_year_for_rollovers(results, text):
         """Fix year-spanning date ranges parsed without explicit years.
 
-        When consecutive parsed dates go backwards chronologically (e.g.
-        "from 23th December until 8th January") but neither substring contains
-        an explicit year, the later-mentioned date belongs to the next calendar
-        year and its year is incremented by one.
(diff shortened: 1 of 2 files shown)


AdrianAtZyte:
Could you keep coverage up?

yamilmaud:
> Could you keep coverage up?

yep, done

AdrianAtZyte:
> I do not think this is the right approach. The relative base chaining in `search_parse` already handles this, in every language, when `PREFER_DATES_FROM` is set to `future`:
>
> ```python
> >>> search_dates("Closed from 23th December until 8th January.", languages=["en"], settings={"PREFER_DATES_FROM": "future"})
> [('from 23th December', datetime.datetime(2026, 12, 23, 0, 0)), ('8th January', datetime.datetime(2027, 1, 8, 0, 0))]
> >>> search_dates("Cerrado desde el 23 de diciembre hasta el 8 de enero.", languages=["es"], settings={"PREFER_DATES_FROM": "future"})
> [('23 de diciembre', datetime.datetime(2026, 12, 23, 0, 0)), ('8 de enero', datetime.datetime(2027, 1, 8, 0, 0))]
> ```
>
> So what this PR really does is change the default, and it does so in a way that has a few problems:
>
> - With `PREFER_DATES_FROM` set to `future`, the year gets bumped twice, so the example above now returns 2028 for January. That breaks the configuration that works today.
> - The connector list is English-only, so the Spanish example above still lands both dates in the same year.
> - `test_search_dates_rollover_skipped_when_substring_position_not_found` asserts the wrong result as expected. If irregular whitespace makes the fix silently not apply, that is a bug, not a feature to test for.
> - Comparing only months means "from 30th December until 2nd December" is not rolled over either.
>
> If the requirement is just that ranges spanning the new year parse correctly, I think `PREFER_DATES_FROM` covers it and we should close this. If the default must change, it should be done in the relative base logic itself, language-independently, and without double-rolling under `future` or looking substrings up in the original text.

yamilmaud:
Thanks for the detailed review — all four points check out, confirmed each one directly (double-bump with `PREFER_DATES_FROM=future`, English-only connector list, the whitespace-fallback test locking in a real bug, and the same-month-wrap gap).

I tried reworking the fix into the relative-base chaining logic as you suggested (force `PREFER_DATES_FROM='future'` only when chaining off a date that had no explicit year of its own, so it doesn't fire on cases like "2014. July 13th" where the year is meant to carry forward as context, not as an ordering constraint — and only when the user hasn't already set `PREFER_DATES_FROM`, so it can't double up). That version correctly handles the original issue, the Spanish case, the same-month wrap, and the wide-gap case, without needing any text-position lookups or English-only word list.

However, it surfaces a deeper pre-existing issue: `parser` (a `DateDataParser`) is a single shared, mutable object reused across every substring — and every split-candidate explored for a substring — in `parse_found_objects`/`choose_best_split`. `parser._settings` gets reassigned per item (originally just for `RELATIVE_BASE`), and that reassignment persists across unrelated items and even unrelated candidate splits, since nothing resets it between them.

Concretely: `"2014. July 13th July 14th"` (an existing test, unrelated to this PR) only gets year 2014 for "July 13th"/"July 14th" today because of this leak — not because of the "official" `set_relative_base`/`already_parsed` chaining. When "July 13th July 14th" fails to parse as a whole and falls into the split-recursion path, that recursion builds its own chain from a **fresh, empty list**, so "2014" isn't actually visible to it through the intended mechanism. It only inherits year 2014 because `parser._settings` was still holding `RELATIVE_BASE=2014-...` left over from processing the *previous, unrelated* top-level item. When I made `parse_item` reset `parser._settings` to a clean baseline before each attempt (to stop *our* future-bias override from leaking the same way), that pre-existing test's accidental year-inheritance broke too — it fell back to today's real year.

So there are two separable things here: (1) the double-bump/English-only/whitespace-bug/same-month issues in our current PR, which I can fix; and (2) this shared-mutable-parser-settings leak, which predates our PR and which some existing tests' expectations are quietly depending on.

I'm not sure how to proceed here — do we close this and pursue the two issues separately? Let me know please.

AdrianAtZyte:
> Thanks for checking. I think we should close this one: the turn-of-year default is a separate decision, and `PREFER_DATES_FROM` already covers the reported case.
>
> The settings leak is a real bug though, and worth its own PR if you are up for it. I think the fix is to seed the split path with the outer chain and set `RELATIVE_BASE` explicitly on every parse, restoring the base settings when there is none. The `2014. July 13th July 14th` test should then pass because of the chaining, not because of the leak, so no expectations need to change.

yamilmaud:
Sounds good — closing this one. Agreed the turn-of-year default is a separate design decision, and `PREFER_DATES_FROM` already covers the reported case for anyone who needs it.

I'll open a separate PR for the `parser._settings` leak using the approach you outlined (seed the split path with the outer chain, set `RELATIVE_BASE` explicitly on every parse, restore base settings when there is none).

PR #1364, opened 2026-08-07
fix(dateparser):  Parse Arabic plural "لبق N تاعاس" (hours ago)

Rodma1:
Parse Arabic plural "لبق N تاعاس" (hours ago)

  CLDR only provides the singular form "ةعاس" for the hour relative-time
  regex, so plural phrases like "تاعاس ٥ لبق" (5 hours ago) were not
  parsed. Add the plural form "تاعاس" to the supplementary data and
  regenerate ar.py.
  example: print(dateparser.parse('قبل ٥ ساعات', languages=['ar']))


AdrianAtZyte:
> Right bug, wrong layer. It patches one CLDR regex for one unit when the root cause is that قبل is missing from the ago word list.

PR #1371, opened 2026-08-30
Avoid shared language_detector and locale cache in search_dates

### LINKED ISSUE #1369: Concurrent `search_dates` calls race on the `_search_with_detection` singleton and return wrong dates
`dateparser.search.search_dates` runs through a single process-wide object created at import time:

```python
# dateparser/search/__init__.py:3
_search_with_detection = DateSearchWithDetection()
```

Two code paths on that singleton store a value on `self` and read it back a statement later, with nothing serialising the pair. Concurrent `search_dates()` calls can therefore end up using a value another call has just stored, and the visible effect is a **wrong result rather than an exception** — `None` for text that does contain a date, or a plausible-looking but wrong date.

#1346 ("Make thread-safe", merged as cd9bb13) fixed the `Dictionary`/`Locale`/`Settings` races and stopped `search_dates` writing `RELATIVE_BASE` onto the shared default settings, but it touches neither of these two sites; both are verbatim on current `master`. This is the follow-up flagged in §3 of https://github.com/scrapinghub/dateparser/pull/1346#issuecomment-5396152832, which was never filed.

## Site 1 — the shared `language_detector`

https://github.com/scrapinghub/dateparser/blob/cd9bb13e63f41eb5f44074cfdc73a41aa62293e1/dateparser/search/search.py#L272-L281

```python
if languages:
    self.language_detector = FullTextLanguageDetector(languages=languages)      # STORE
else:
    self.language_detector = FullTextLanguageDetector(
        list(self.available_language_map.values())
    )                                                                           # STORE

detected_language = self.language_detector._best_language(text) or (            # LOAD
    settings.DEFAULT_LANGUAGES[0] if settings.DEFAULT_LANGUAGES else None
)
```

`DateSearchWithDetection.detect_language` is called unconditionally from `search_dates` (`search.py:336`), so this store happens on **every** call — including when `languages=` is passed, i.e. also for callers who never asked for language detection.

What makes this more than a benign overwrite is that `FullTextLanguageDetector` is *single-use*: `_best_language` → `character_check` narrows `self.languages` in place and never restores it. That is precisely why `detect_language` builds a new detector on every call. Reusing one gives the wrong answer:

```python
from dateparser.languages.loader import LocaleDataLoader
from dateparser.search.text_detection import FullTextLanguageDetector

locales = [LocaleDataLoader().get_locale(s) for s in ("en", "ru", "de")]

d = FullTextLanguageDetector(locales)
print([l.shortname for l in d.languages])       # ['en', 'ru', 'de']
print(d._best_language("19 марта 2001"))        # ru
print([l.shortname for l in d.languages])       # ['ru']   <- narrowed in place
print(d._best_language("4 October 1957"))       # ru       <- wrong
print(FullTextLanguageDetector(locales)._best_language("4 October 1957"))  # en
```

So a call that loads another thread's detector does not just detect with a different-but-equivalent object: it detects with an object already narrowed to somebody else's language (and both threads then mutate it concurrently). Downstream, `_get_candidate_languages` either comes back empty — `search_dates` returns `None` — or carries the foreign language, and the text is translated with the wrong locale.

## Site 2 — the shared `language`

https://github.com/scrapinghub/dateparser/blob/cd9bb13e63f41eb5f44074cfdc73a41aa62293e1/dateparser/search/search.py#L34-L46

```python
class _ExactLanguageSearch:
    def __init__(self, loader):
        self.loader = loader
        self.language = None

    def get_current_language(self, shortname):
        if self.language is None or self.language.shortname != shortname:
            self.language = self.loader.get_locale(shortname)                   # STORE

    def search(self, shortname, text, settings):
        self.get_current_language(shortname)
        result = self.language.translate_search(text, settings=settings)        # LOAD
        return result
```

There is exactly one `_ExactLanguageSearch` in the process (`search.py:222`, on the same singleton), so `self.language` is one shared slot used as a 1-entry cache. A store from another thread landing between the check/store and the load makes this call translate its text with the wrong locale.

## Reproduction

Both store→load gaps are only a few bytecodes wide, so under the GIL they are hard to hit by chance. Parking each thread immediately after its store — so both threads store before either loads — shows the interleaving the code permits. The barrier changes only the timing, not the semantics.

**Site 1:**

```python
import threading
from concurrent.futures import ThreadPoolExecutor
from dateparser.search import search_dates
from dateparser.search.search import DateSearchWithDetection

RU = "Договор подписан 19 марта 2001 года в Москве"
EN = "The satellite was launched on 4 October 1957 from Baikonur"
print("sequential ru:", search_dates(RU, languages=["ru"]))
print("sequential en:", search_dates(EN, languages=["en"]))

_slot, barrier = {}, threading.Barrier(2, timeout=30)

def _get(self):
    return _slot["v"]

def _set(self, value):
    _slot["v"] = value              # the store the code performs
    try:
        barrier.wait()              # ... park until the other thread has stored too
    except threading.BrokenBarrierError:
        pass

DateSearchWithDetection.language_detector = property(_get, _set)

with ThreadPoolExecutor(2) as ex:
    ru, en = ex.map(lambda a: search_dates(a[0], languages=[a[1]]), [(RU, "ru"), (EN, "en")])
print("concurrent ru:", ru)
print("concurrent en:", en)
```

```
sequential ru: [('19 марта 2001 года в', datetime.datetime(2001, 3, 19, 0, 0))]
sequential en: [('on 4 October 1957 from', datetime.datetime(1957, 10, 4, 0, 0))]
concurrent ru: None
concurrent en: [('on 4 October 1957 from', datetime.datetime(1957, 10, 4, 0, 0))]
```

**Site 2** — same idea on `get_current_language`; this one returns a wrong date rather than `None`:

```python
import threading
from concurrent.futures import ThreadPoolExecutor
from dateparser.search import search_dates
from dateparser.search.search import _ExactLanguageSearch

RU = "Договор подписан 19 марта 2001 года в Москве"
EN = "The satellite was launched on 4 October 1957 from Baikonur"
print("sequential ru:", search_dates(RU, languages=["ru"]))
print("sequential en:", search_dates(EN, languages=["en"]))

_orig, barrier = _ExactLanguageSearch.get_current_language, threading.Barrier(2, timeout=30)

def patched(self, shortname):
    _orig(self, shortname)
    try:
        barrier.wait()
    except threading.BrokenBarrierError:
        pass

_ExactLanguageSearch.get_current_language = patched

with ThreadPoolExecutor(2) as ex:
    ru, en = ex.map(lambda a: search_dates(a[0], languages=[a[1]]), [(RU, "ru"), (EN, "en")])
print("concurrent ru:", ru)
print("concurrent en:", en)
```

```
sequential ru: [('19 марта 2001 года в', datetime.datetime(2001, 3, 19, 0, 0))]
sequential en: [('on 4 October 1957 from', datetime.datetime(1957, 10, 4, 0, 0))]
concurrent ru: [('2001', datetime.datetime(2001, 8, 25, 0, 0))]
concurrent en: [('on 4 October 1957 from', datetime.datetime(1957, 10, 4, 0, 0))]
```

The Russian text was translated with the English locale, so instead of 2001-03-19 the call reports the bare token `2001` with today's month and day — wrong, and plausible enough to pass unnoticed downstream.

**Expected:** `search_dates` returns the same result from many threads as from one; each call uses the detector and locale for its own input.
**Actual:** with concurrent calls that disagree about the language, a call can return `None`, or a date built from the wrong locale.

## How often does it actually happen

I want to be precise about this rather than alarming.

**Confirmed:** the interleaving is unsynchronised and permitted, and forcing it produces the results above on `cd9bb13`.

**Confirmed, under natural scheduling, with the store/load pair instrumented:** wrapping `language_detector` in a descriptor that records whether the value read back is the value this thread wrote — 16 threads alternating `search_dates(text, languages=["ru"])` / `["en"]`, `sys.setswitchinterval(1e-7)`, 7921 calls in 240 s — gave **19 cross-thread reads and 14 wrong results**. Not only `None`: an English call returned `[('1957', datetime(1957, 8, 25))]` instead of 1957-10-04, which is exactly what a detector already narrowed to `ru` produces (see the snippet above). Caveat: the wrapper adds a handful of bytecodes on both sides of the gap, so the window it exercises is somewhat wider than production's.

**Not reproduced:** with completely unmodified code — same workload, results compared against the single-threaded baseline, no patching of any kind — **0 wrong results in 20 912 calls** (16 threads, 600 s, the long texts above) and **0 in 21 220 calls** (28 threads, 600 s, short inputs `"19 марта 2001"` / `"4 October 1957"`). So the unwidened window is narrow enough that I cannot put a frequency on this for real workloads.

**Not claimed:** that every threaded user is affected. It needs concurrent `search_dates` calls that disagree about the language — a process that only ever searches one language never changes either shared value. Passing more than one entry in `languages=` softens site 1 without removing it: `_get_candidate_languages` appends the caller's own languages after the detected one, so a wrong detection is still tried first and wins if it finds anything (`search_dates` returns on the first candidate that yields dates).

Given the fix is small and removes state rather than adding locks, I'd argue it is worth doing on the strength of the mechanism, independently of the frequency.

## Suggested direction

Neither value needs to be instance state — both are derived per call:

- `detect_language`: build the `FullTextLanguageDetector` into a local and use the local. Since the object is single-use anyway (it narrows `self.languages`), the attribute buys nothing.
- `_ExactLanguageSearch`: have `get_current_language` *return* the locale and let `search` use the returned value instead of stashing it on `self`. `LocaleDataLoader.get_locale` already caches (`loader.py:207-230`, under `_load_lock`), so dropping the one-slot cache costs a dict lookup, not a reload.

## Notes

- Not covered by the tests added in #1346: `tests/test_thread_safety.py` touches `search_dates` only in `test_search_does_not_mutate_shared_relative_base`, which asserts the shared `Settings.RELATIVE_BASE` is not mutated.
- The `ngram` strategy from #1351 avoids site 2 (`_NgramDateSearch` is stateless) but still goes through `detect_language`, so site 1 applies to both strategies.
- Related, not duplicates: #441 and #1291 (both closed by #1346; both about `dateparser.parse`, neither mentions `search_dates`); #834 ("Resolve concurrency issues", still open, does not touch the search module); #853, where threaded `search_dates` users were pointed at the concurrency gap.

Environment: `cd9bb13e63f41eb5f44074cfdc73a41aa62293e1` (master after #1346), CPython 3.11.15, Linux.

Steve0x2a:
Fixes #1369.

`search_dates` runs through a process-wide `DateSearchWithDetection` singleton. Two leftover store-then-load races from #1346 can make concurrent calls return `None` or a date built from the wrong locale:

1. `detect_language` stashed a `FullTextLanguageDetector` on `self` and then loaded it. That detector is single-use: `_best_language` narrows `self.languages` in place. A concurrent call can observe another call's already-narrowed detector.
2. `_ExactLanguageSearch` kept `self.language` as a 1-slot cache on the same singleton. A store from another thread between `get_current_language` and `translate_search` uses the wrong locale.

Neither value needs to be instance state. Build the detector as a local, and have `get_current_language` return the locale (`LocaleDataLoader.get_locale` already caches). Public API is unchanged.

Thanks @serhii73 for the report and the diagnosis.

### DIFF
--- a/dateparser/search/search.py
+++ b/dateparser/search/search.py
@@ -34,15 +34,13 @@ def _add_time_span_results(results, text, settings):
 class _ExactLanguageSearch:
     def __init__(self, loader):
         self.loader = loader
-        self.language = None
 
     def get_current_language(self, shortname):
-        if self.language is None or self.language.shortname != shortname:
-            self.language = self.loader.get_locale(shortname)
+        return self.loader.get_locale(shortname)
 
     def search(self, shortname, text, settings):
-        self.get_current_language(shortname)
-        result = self.language.translate_search(text, settings=settings)
+        language = self.get_current_language(shortname)
+        result = language.translate_search(text, settings=settings)
         return result
 
     @staticmethod
@@ -270,13 +268,13 @@ class DateSearchWithDetection:
             )
 
         if languages:
-            self.language_detector = FullTextLanguageDetector(languages=languages)
+            language_detector = FullTextLanguageDetector(languages=languages)
         else:
-            self.language_detector = FullTextLanguageDetector(
+            language_detector = FullTextLanguageDetector(
                 list(self.available_language_map.values())
             )
 
-        detected_language = self.language_detector._best_language(text) or (
+        detected_language = language_detector._best_language(text) or (
             settings.DEFAULT_LANGUAGES[0] if settings.DEFAULT_LANGUAGES else None
         )
         return detected_language
--- a/tests/test_thread_safety.py
+++ b/tests/test_thread_safety.py
@@ -8,14 +8,16 @@ from dateparser.conf import settings as base_settings
 from dateparser.date import DateDataParser
 from dateparser.languages.dictionary import Dictionary
 from dateparser.search import search_dates
+from dateparser.search.search import DateSearchWithDetection, _ExactLanguageSearch
 from tests import BaseTestCase
 
 
 class TestThreadSafety(BaseTestCase):
     """Regression tests for thread-safety issues.
 
-    See https://github.com/scrapinghub/dateparser/issues/441 and
-    https://github.com/scrapinghub/dateparser/issues/1291.
+    See https://github.com/scrapinghub/dateparser/issues/441,
+    https://github.com/scrapinghub/dateparser/issues/1291 and
+    https://github.com/scrapinghub/dateparser/issues/1369.
     """
 
     def setUp(self):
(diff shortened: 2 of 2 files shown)


serhii73 (approved):

serhii73:
Thanks!

Issue #1277, opened 2025-06-20
Aren't "zh" (Chinese) and "zh-Hans" (Chinese, traditional script) the same language?


Merinorus:
Hi,

While working on [another project](https://github.com/paperless-ngx/paperless-ngx/pull/10181), we saw that parsing Chinese dates rarely works when the language `zh-Hans`, whereas it works fine with `zh.`

Example output:
```
Detecting text: 2024年6月1日
ja: 2024-06-01 00:00:00
zh: 2024-06-01 00:00:00
zh-Hans-HK: None
zh-Hans: None
zh-Hant-HK: None
zh-Hant: None
```

From what I understand about CLDR, Chinese (simplified) is the default locale for Chinese: https://st.unicode.org/cldr-apps/v#/zh_Hans//

In the dateparser source code, `zh-Hans` appears to be a *subset* of `zh`, sharing most translation tables but lacking simplification rules, which seems to cause the parsing failures.

I made a proposed change here: https://github.com/scrapinghub/dateparser/compare/master...Merinorus:dateparser:chinese-dates

Before submitting a PR, I’d like to confirm whether my understanding is correct.

Thank you!

Issue #875, opened 2021-02-05
Issue parsing ZH-Hant locale


croqaz:
Hi.

This doesn't work:
`dateparser.parse('2020年9月1日 下午6:25', languages=['zh-Hant'])`
This works:
`dateparser.parse('2020年9月1日 下午6:25', languages=['zh'])`
This also works:
`dateparser.parse('2020年9月1日 下午6:25', languages=['zh-Hant', 'zh'])`

It's weird because I can see all the info in https://github.com/scrapinghub/dateparser/blob/master/dateparser_data/cldr_language_data/date_translation_data/zh-Hant.json , but it also needs "zh" just to make it work.

Using latest 1.0.0 version, Python 3.8.

Thank you!

PR #1377, opened 2026-09-04
Fix IndexError in search_dates for zh-Hans, zh-Hant, and yue locales

### LINKED ISSUE #1372: search_dates() raises IndexError for the zh-Hans, zh-Hant and yue locales
## Problem

`search_dates()` raises `IndexError: list index out of range` for the `zh-Hans`, `zh-Hant` and `yue` locales. It is reachable through the plainest possible call, with no `languages=` argument at all:

```python
>>> from dateparser.search import search_dates
>>> search_dates("聽日")          # Cantonese for "tomorrow"; autodetection selects yue
IndexError: list index out of range
>>> search_dates("舊年")          # "last year"
IndexError: list index out of range
```

It also fires whenever one of the three locales is requested explicitly, including in a mixed language list, and including for plain ASCII input:

```python
>>> search_dates("今天", languages=["zh-Hans"])                 # IndexError
>>> search_dates("2020年3月4日", languages=["zh-Hant"])          # IndexError
>>> search_dates("00Z", languages=["yue"])                      # IndexError
>>> search_dates("2020年3月4日", languages=["zh-Hans", "en"])    # IndexError
>>> search_dates("3 hours ago at 5pm", languages=["zh-Hans"])   # IndexError
```

`zh`, `ja` and every other locale are unaffected, and `dateparser.parse()` does not crash (for these locales it returns `None` instead — that half is #1277).

Traceback:

```
dateparser/search/__init__.py:70  -> search/search.py:357
                                  -> search/search.py:184
                                  -> search/search.py:45
                                  -> languages/locale.py:310
      [original_tokens[i], original_tokens[i + 1]],
IndexError: list index out of range
```

## Root cause

Two things combine in `Locale.translate_search()` (`dateparser/languages/locale.py`):

1. `locale.py:277` — `word_joint_unsupported_languages = ["zh", "ja"]`. Exactly `zh` and `ja` carry a `simplifications` block that rewrites `年`/`月`/`日` into separators before this point; `zh-Hans`, `zh-Hant` and `yue` do not, so they reach the word-joining branch at `locale.py:302-312` with raw CJK tokens, which the other two never do.
2. `locale.py:291` computes `next_word` defensively — `simplified_tokens[i + 1] if i < last_token_index else ""` — but the branch it guards reads `original_tokens[i + 1]` at `locale.py:310` with **no** equivalent bounds check. On the final token, `current_and_next_joined` can still match a dictionary entry (the joined form equals the bare word, since `next_word` is `""`), the branch is taken, and the index overflows.

Introduced in c5f0fa4 (#953, 2021-08-09); `git show c5f0fa4^:dateparser/languages/locale.py` contains no `i + 1` index at all.

## Proposal

Add the missing bounds check to the condition at `locale.py:302`:

```python
                elif (
                    i < last_token_index
                    and current_and_next_joined in dictionary
                    and word not in dashes
                    and self.shortname not in word_joint_unsupported_languages
                ):
```

I verified this locally:

- every repro above stops raising;
- the full suite is unchanged — `24186 passed, 18 skipped, 1 xfailed`, identical to baseline;
- `search_dates("聽日")` and `search_dates("舊年")` now return **correct** results rather than merely not crashing.

## Scope and considerations

- **Please do not fix this by adding the three locales to `word_joint_unsupported_languages`.** I tried that: it stops the crash, but `search_dates("2020年3月4日", languages=["zh-Hans"])` then returns `datetime(6, 5, 27)` — silent garbage — because `translate_search` yields `"2020yearmarch4day"`. That trades a loud crash for a wrong date.
- Related but distinct: #1277 (`zh-Hans` vs `zh` locale data) and #875 (`zh-Hant`). Those are data gaps that make `parse()` return `None`; this is a code defect that survives them. I applied #1277's proposed change in-memory and `search_dates()` still raises for `今天`, `昨天`, `上个月` and `00Z`. Sensible sequencing is the bounds guard first (crash → `None`, matching `parse()`), the data fix second (`None` → correct date).
- After the guard lands, `2020年3月4日` under `zh-Hans` returns `None` rather than a date. That is #1277's territory, not a regression from this change.

## Acceptance criteria

- None of the calls listed above raises `IndexError`.
- Regression tests cover `zh-Hans`, `zh-Hant` and `yue` in `tests/test_search.py`, which currently contains zero references to any of the three (`grep -c 'zh-Han\|yue' tests/test_search.py` → `0`) — which is how this survived four and a half years.

abhi-0203:
## Summary
Fixes `IndexError: list index out of range` in `search_dates()` for `zh-Hans`, `zh-Hant`, and `yue` locales.

## Root Cause
In `Locale.translate_search()` (`dateparser/languages/locale.py`), the word-joining branch read `original_tokens[i + 1]` without checking bounds (`i < last_token_index`), causing an index overflow on the final token when joined forms matched dictionary entries.

## Fix
- Added `i < last_token_index` bounds check to the word-joining condition.
- Added regression unit tests covering `zh-Hans`, `zh-Hant`, and `yue` locales in `tests/test_search.py`.

Closes #1372

### DIFF
--- a/dateparser/languages/locale.py
+++ b/dateparser/languages/locale.py
@@ -300,7 +300,8 @@ class Locale:
                     translated_chunk.append(word)
                     original_chunk.append(original_tokens[i])
                 elif (
-                    current_and_next_joined in dictionary
+                    i < last_token_index
+                    and current_and_next_joined in dictionary
                     and word not in dashes
                     and self.shortname not in word_joint_unsupported_languages
                 ):
--- a/tests/test_search.py
+++ b/tests/test_search.py
@@ -1453,6 +1453,14 @@ class TestNgramSearch(BaseTestCase):
             )
         )
 
+
+    def test_search_dates_cjk_locales_no_index_error(self):
+        self.assertIsNotNone(search_dates("聽日"))
+        self.assertIsNotNone(search_dates("舊年"))
+        self.assertIsNotNone(search_dates("今天", languages=["zh-Hans"]))
+        self.assertIsNone(search_dates("2020年3月4日", languages=["zh-Hant"]))
+        self.assertIsNotNone(search_dates("00Z", languages=["yue"]))
+
     def test_unknown_strategy_raises_error(self):
         with self.assertRaisesRegex(ValueError, "strategy must be"):
             search_dates("4 October 1957", languages=["en"], strategy="unknown")


AdrianAtZyte:
> The fix looks right to me, thanks. A few things about the test:
>
> - Pre-commit fails: ruff-format complains about the extra blank line before the new test.
> - Please, move the test to `TestTranslateSearch`; the bug is in `translate_search`, not in the ngram strategy.
> - Please, drop the `assertIsNone` line for `2020年3月4日`. Returning `None` there is a known bug (#875, #1277), and this test would break as soon as someone fixes it.
> - I think the other cases should assert the actual dates rather than just `assertIsNotNone`, since your fix makes them correct now and the base test class already freezes time.

PR #1378, opened 2026-09-05
fix: force month-day order for ISO leading year with explicit DATE_ORDER

3xp0rt:
dateparser already forced month-day ordering after a leading 4-digit year, but only when DATE_ORDER wasn't explicitly set. So an explicit DATE_ORDER=DMY/YDM (needed for a source's genuinely ambiguous dates) still broke unambiguous ISO strings:
```
dateparser.parse("2026-08-31", settings={"DATE_ORDER": "DMY"})  # -> None (invalid month=31)
dateparser.parse("2026-05-08", settings={"DATE_ORDER": "DMY"})  # -> 2026-08-05 (swapped, wrong)
```
This drops that guard, so a leading year always fixes the remaining order as month-then-day, regardless of DATE_ORDER. Ambiguous non-ISO strings without a leading year (e.g. "03-12-2026") still honor DATE_ORDER as before.

### DIFF
--- a/dateparser/parser.py
+++ b/dateparser/parser.py
@@ -636,17 +636,9 @@ class _parser:
             type = 0
 
             num_directives = self.ordered_num_directives
-            if (
-                skip_component == "year"
-                and self.day is None
-                and self.month is None
-                and "DATE_ORDER" not in self.settings._mod_settings
-            ):
-                # The date string starts with a four-digit year (e.g. an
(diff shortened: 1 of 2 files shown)


AdrianAtZyte:
Not sure about this one. If we really want to enforce DATE_ORDER with ISO, which may make no sense to begin with, should we not fail on 2026 as day?

PR #1379, opened 2026-09-08
Repair contributor commands and default tox packaging checks

### LINKED ISSUE #1373: The commands documented in CONTRIBUTING.rst do not work as described
## Problem

The commands `CONTRIBUTING.rst` tells a new contributor to run do not do what the document says.

**1. `tox` never runs the packaging check.** `tox.ini:2` lists an env named `twinedeck`; the env defined at `tox.ini:65` is `twinecheck`. tox does not error on an envlist entry with no matching section — it silently synthesizes it from `[testenv]`:

```console
$ tox config -e twinedeck -k commands
[testenv:twinedeck]
commands = pytest --cov=dateparser --cov-report=xml tests
```

So `tox` — the command `CONTRIBUTING.rst:90` tells contributors to run — runs the full test suite a *second* time under a misleading name, never runs `python -m build` + `twine check`, and exits 0. CI is unaffected only because `main.yml:31-32` passes `-e twinecheck` explicitly, which is what has kept this invisible.

This is a recurrence, not a one-off: #1164 ("flake8 tox env runs Python tests", open since 2023-04, no comments) is the identical tox-synthesis bug in the same file. It was never fixed — 36a5634 (#1248) removed the flake8 env and introduced `twinedeck` in the same commit. For the record, that was my own PR.

**2. Bare `pytest` from the repo root cannot collect.** After the documented `pip install -e .`:

```
24203 tests collected, 6 errors
!!!!!!!!!!!!!! Interrupted: 6 errors during collection !!!!!!!!!!!!!!
```

`pytest.ini:3` sets `--doctest-modules` with no `testpaths`, so pytest imports `dateparser/calendars/{hijri,hijri_parser,jalali,jalali_parser}.py` (needs `hijridate`) and `fuzzing/{dateparser_fuzzer,fuzz_helpers}.py` (needs `atheris`). With `.[calendars]` installed, 2 of the 6 remain.

The flag buys nothing in return: `tox.ini:15` scopes pytest to `tests/`, which contains **zero** doctests. The only two doctests in the package both fail today anyway — `dateparser/date.py:581` hardcodes a day-of-month and fails on 30 days out of 31, and `dateparser/search/__init__.py:57` uses `>>>` where doctest requires `...`, raising `SyntaxError: '(' was never closed`.

**3. `pytest.ini:4`'s `--assert=plain` strips assertion diffs.** It disables pytest's assertion rewriting, so a failing bare `assert` prints `E AssertionError` with no comparison. `tests/` has 34 bare `assert` statements (the 341 `self.assert*` calls are unaffected — unittest builds its own message). Both this and `--doctest-modules` were added by the same commit, c5a4b38 ("add pytest flake8", 2020) — `--assert=plain` was the standard pytest-flake8 workaround. flake8 is long gone.

**4. The document is factually wrong about the toolchain.** `CONTRIBUTING.rst:87` still says changes must "pass flake8", and `:92` says tox "checks for code style and maximum line length (119 characters)". flake8 was replaced by ruff in #1248. This repo's `[tool.ruff]` sets no `line-length` and selects no `E501`, so no line limit is enforced at all, while `ruff-format` targets 88.

**5. `CONTRIBUTING.rst:158`'s data-regeneration command does not run.**

```console
$ dateparser_scripts/write_complete_data.py
bash: ...: Permission denied     # exit 126
```

The file is `0644` with no shebang. The visually identical instruction at `:181` works because that file is `0755` with one.

**6. Dead pre-ruff configuration.** `.isort.cfg` (isort is not installed and ruff's `I` rules are not selected) and `conftest.py:1-9`, a `pytest-flake8` shim for a plugin that is no longer a dependency. Both are inert; nothing reads them.

## Why it matters

This is the on-ramp for every new contributor, and each item silently misleads rather than failing loudly. Item 1 in particular means the packaging check a contributor believes they ran did not run.

## Proposal

Fix the `twinedeck` → `twinecheck` typo; decide what to do about `--doctest-modules` (see below); drop `--assert=plain`; update `CONTRIBUTING.rst` steps 3-5 to describe ruff and `tox -e pre-commit`; change the regeneration command to `python -m dateparser_scripts.write_complete_data`, matching the style already used at `:190`; delete `.isort.cfg` and the `conftest.py` shim.

Removing `--assert=plain` is verified safe: `pytest -o addopts='--assert=rewrite' tests` gives the same `24186 passed, 18 skipped, 1 xfailed`.

## Scope and considerations

- **One genuine decision, not a mechanical fix:** `--doctest-modules` can be (a) dropped outright, losing two doctests that already fail and never run in CI, (b) kept with `testpaths = tests`, which fixes bare `pytest` but leaves the doctests dead, or (c) kept with the two doctests repaired and explicit `--ignore` entries restored. Note `[testenv]` declares no `extras` and installs `atheris` only on `python_version < '3.12'`, so simply pointing the flag at the package would re-break the base `py`/`min` envs.
- Deleting `.isort.cfg` should not turn into "should we enable ruff's `I` rules?" — that is a separate decision; this issue only removes config nothing reads.
- Closes #1164, which is independently obsolete: `tox.ini` no longer has a flake8 env.
- Adjacent, out of scope here: `MANIFEST.in:7` includes a `requirements.txt` deleted in 2020 (one setuptools warning per build), `.readthedocs.yml:12` says "Keep in sync with `.github/workflows/checks.yml`", a file that has never existed in this repository, and `docs/installation.rst:11` still recommends `easy_install`. All three are one-line fixes that can ride along.

## Acceptance criteria

- `tox` runs the packaging check.
- `pytest` from the repo root collects and runs without errors after `pip install -e .`.
- A failing bare `assert` in `tests/` shows the comparison.
- Every command in `CONTRIBUTING.rst` runs successfully as written, and it names the tools actually in use.

jackwalkerlabs:
Repair the contributor commands from #1373: correct the default tox packaging environment name, scope bare pytest to `tests/`, restore assertion rewriting, remove unused pytest-flake8/isort configuration, and document the actual ruff/test setup plus the executable translation-regeneration command and its dependencies.

The removed doctest flag did not exercise package doctests in the existing tox test command (`pytest tests`); root collection instead imported optional scripts/fuzzers and failed. Package doctest repairs can be handled separately.

Fixes #1373.

Validation: reproduced the synthesized `twinedeck` test command and six baseline collection errors. After the change, `TZ=UTC pytest -q`: 24,358 passed, 4 skipped, 1 xfailed; `tox -e twinecheck`: wheel/sdist build and both twine checks pass; `tox -e pre-commit`: ruff and formatting pass. The documented translation regeneration command runs successfully with no generated-data changes. On this host’s America/New_York timezone, two pre-existing timezone assertions fail with both old and new pytest options; UTC avoids those environment-dependent failures. The complete multi-interpreter tox matrix was not run locally.

AI disclosure: Implemented, reviewed, and tested by OpenAI Codex under the account owner’s authorization; no human code review is claimed.


serhii73 (approved):

serhii73:
Thanks!

PR #1380, opened 2026-09-08
Reject relative dates that would truncate a grouped number

jackwalkerlabs:
Reject relative-date matches that start partway through a grouped numeric quantity. For example, `1,000,000 days ago` must not silently consume only the trailing numeric fragment. Reject the whole expression when grouping would truncate a quantity, including expressions with another valid duration after it.

Addresses the unambiguous truncation bug in #1367. A comma is read as a decimal mark, and numbers with grouping separators are rejected; locale-aware interpretation of grouping remains outside this change. Decimal commas such as `1,5 hours ago`, spaced signs, compound durations, and adjacent clock times remain covered. The accidental `1,000`/`1,0000` interpretation is no longer pinned by these regression tests.

The prefix check uses a straightforward forward search. `get_kwargs` consistently returns `(kwargs, explicit_signs)`, including the empty result, and the regression class precedes the `__main__` block. The README uses the requested single-sentence explanation after the introductory examples.

This also makes previously meaningless combinations such as `2024-06-01 3 days ago` and `1st 3 days ago` return `None`, as discussed in review.

Validation of current head `55ff4c0d222ba2d863708576140f3015dc3e03ee`: independent Factory source/UX review approved. Fresh offline Python 3.11.2 verification passed the targeted grouping regressions, the full freshness suite, `TZ=UTC pytest -q tests` (**24,219 passed, 21 skipped, 1 xfailed**), and repository-pinned Ruff 0.9.3 lint/format. The full supported-Python and optional-extra matrix remains for upstream CI.

Factory delivery run: `FC-20260914-20A898`, retaining this PR and its original fork branch. Review and verification completed in `FC-20260913-101079`; the delivery retry reused that exact approved and tested commit. The initial run `FC-20260913-DD248D` was stopped before publication after an invalid reviewer response; an operator recovered its actual implementation checkpoint for fresh review and verification.

AI disclosure: Implemented, independently reviewed, and tested by hosted FactoryChief using Codex under the account owner's authorization. Operator assistance handled recovery, delivery with an existing workflow-authorized credential after a token-scope rejection, and this description update; no human code review is claimed.


AdrianAtZyte:
> Thanks, the direction is fine: failing beats returning a truncated number, and the whole-string rejection is right since a lookbehind in `PATTERN` would still let `1,000,000 days 3 hours ago` parse as 3 hours ago. A few things before merging:
>
> - Please, drop the `1,000 hours ago` and `1,0000 hours ago` cases. `1,000` meaning one hour is an accident of #876, not something I want a test to pin down. `1,5 hours ago` is enough.
> - The new test class ended up after the `if __name__ == "__main__":` block, please move it above it.
> - The `REVERSE` + `endpos` + `\Z` search is harder to read than a forward search with `$` on `date_string[:match.start()]`, and date strings are short enough that the copy does not matter. I would go with the simple version. The comment above it can go too, the check reads fine on its own.
> - The README paragraph can be a single sentence: a comma is read as a decimal mark, and numbers with grouping separators are rejected. Please, put it right after the existing intro text of the section rather than before the example.
> - `get_kwargs` still returns either `{}` or a tuple, and `_parse_date` still has the `isinstance` branch for it. Returning `kwargs, explicit_signs` unconditionally lets you drop that branch (the benchmark test already unpacks a tuple).
>
> For the record, this also turns things like `2024-06-01 3 days ago` or `1st 3 days ago` from “3 days ago” into `None`. I think that is fine, those never meant anything, but worth a line in the description.

AdrianAtZyte:
> CodSpeed flags `test_freshness_get_kwargs` as 27% slower. Half of that is `finditer` replacing `findall`, which the fix needs, but the other half is running `_NUMERIC_PREFIX` on every match, and on real input that search can never succeed: the character before each match is a letter or the start of the string, as in `1 day 2 hours ago`. Skipping it in that case gets most of the time back:
>
> ```python
> for match in PATTERN.finditer(date_string):
>     start = match.start()
>     if (
>         start
>         and not date_string[start - 1].isalpha()
>         and _NUMERIC_PREFIX.search(date_string[:start])
>     ):
>         return {}, {}
>     num, unit = match.groups()
> ```
>
> Locally the benchmark loop goes from 22 µs on master to 32 µs on this branch and 26 µs with the guard. I fuzzed 200k strings made of digits, separators, signs and unit words through `get_kwargs` with and without the guard and got identical results, so it is a pure optimization. Please, add it.

itzzdev09 (commented):
I tested this branch against #1367 and it holds up well:

- `1,000,000 days ago` and `1,000,000,000 seconds ago` now return `None` instead of silently reading a truncated fragment. 
- No false positives on the valid relative forms I tried: `1000 days ago`, `5 days ago`, `1 day 2 hours ago`, `2 days and 3 hours ago`, `3 weeks 2 days ago`, `+5 days`, `- 3 days`, `1.5 hours ago` and `1,5 hours ago` all still parse as before.
- `12:00 5 days ago` still works, so the `(?<![\d:])` guard correctly keeps clock times from tripping the prefix check.

One thing worth pinning down with a test or a comment, since it surprised me at first: `12,345 days ago` is *not* rejected — it's still read as `12.345` days (the single-separator, comma-as-decimal case). That's consistent with the issue's scope (only the unambiguous leading-digit truncation is being fixed here), but a reader might reasonably expect it to be rejected too, so an explicit test asserting the intended `12,345` behaviour would make the boundary clear.

Minor: `_NUMERIC_PREFIX` (`\d[\d.,]*[.,\s]*$`) has adjacent greedy classes; anchored at `$` with no alternation the backtracking is bounded, so not a concern, just noting it.

PR #1381, opened 2026-09-09
fix: preserve complete week dates when parsing custom formats

### LINKED ISSUE #1242: %w | %W and %V | %u format string directives are not supported?
Assuming `dateparser` should support all format string directives [supported by python](https://docs.python.org/3/library/time.html#time.strftime), the following code demonstrates discrepancies in how its parsing results deviate from `datetime`'s `strptime`:

```python
import datetime as dt
import dateparser as dp

date = '01.24-1'
date_format = '%W.%y-%w'

date_iso = '01.2024-1'
date_format_iso = '%V.%G-%u'

print('Non-ISO date:')
print('Datetime: %s' % dt.datetime.strptime(date, date_format).date())
print('Dateparser: %s' % dp.parse(date, [date_format]).date())

print('ISO date:')
print('Datetime: %s' % dt.datetime.strptime(date_iso, date_format_iso).date())
print('Dateparser: %s' % dp.parse(date_iso, [date_format_iso]).date())

```
outputs:

```shell
Non-ISO date:
Datetime: 2024-01-01
Dateparser: 2024-10-30
ISO date:
Datetime: 2024-01-01
Dateparser: 2024-10-30
```

ryanchou1994:
## Summary

Custom formats containing a year, week number, and weekday are complete dates, but the missing-field detection treats them as missing their month and day. For example, `01.24-1` with `%W.%y-%w` is overwritten with date preferences instead of returning 2024-01-01. ISO week dates also have their computed year replaced because `%G` is not recognized.

Preserve the complete date returned by `strptime`, including calendar years that differ from the ISO week year. Week 0 can reach the previous year and needs `calendar.isleap`; the private calendar currently loads `_strptime` by mistake. Copy the standard calendar namespace before rebinding the English names, without re-executing calendar or changing its global constants.

The missing-field check uses a flat year/week/weekday condition after `strptime` has validated the format.

The regressions cover week numbering conventions, year boundaries, leap days, weekday names, date preferences, escaped directives, incomplete formats, invalid ISO dates, and global calendar isolation.

Fixes #1242.

## Validation

- The original week-date regressions fail on the base implementation; the additional invalid ISO pairing is rejected before and after simplifying the condition.
- macOS Python 3.13.13 and Linux Python 3.13.15: 24,380 passed, 4 skipped, 1 xfailed, plus 36 subtests on each platform; coverage remains 97%.
- Focused format, calendar-isolation, and strptime tests: 108 passed plus 36 subtests, with the real French locale available on both platforms.
- Ruff 0.9.3 lint and format checks passed with the current upstream configuration, including W605. The full suites ran with the new warnings-as-errors setting enabled.


AdrianAtZyte:
> The new condition in _get_missing_parts re-encodes rules strptime already enforces: %V requires %G and rejects %Y, and %G requires %V plus a weekday. Since we only reach this code after strptime succeeded, a flat check works and reads better.

ryanchou1994:
Thanks, simplified the condition to the three groups: week, weekday, and year. The regex still consumes `%%` pairs so literal directives do not count. I also added the `%G-W%W-%u` case to the invalid ISO tests to cover the validation we rely on from `strptime`.

The focused tests and full suite pass on macOS and Linux with the new warnings-as-errors configuration (24,380 passed, 4 skipped, 1 xfailed, plus 36 subtests on each platform).

AdrianAtZyte (approved):
> Looks good, thanks! Two nits, no need for another round if you would rather not:
>
> - The `if "%U" in fmt or "%W" in fmt or "%V" in fmt:` guard is redundant with the first set intersection, so the block can be flat.
> - I would drop the comment in `strptime.py`. It explains why the alternative was not taken, which belongs in the commit message; the code reads fine on its own.

ryanchou1994:
Thanks for the approval, @AdrianAtZyte!

@serhii73 when you have a moment, could you take a look and approve the workflow runs? They're still waiting on maintainer approval, so the full test suite hasn't run on this PR yet.

Happy to apply Adrian's two nits here or in a follow-up, whichever you prefer.

serhii73 (approved):

serhii73:
Thanks!

PR #1383, opened 2026-09-10
Parse English ordinal dates containing day of

### LINKED ISSUE #1256: Unable to parse "the 1st day of March"
Adding "day" in the middle of a date fails parsing:

```
>>> import dateparser as dp
>>> dp.parse("the 1st day of March, 2025")
>>> dp.parse("the 1st of March, 2025")
datetime.datetime(2025, 3, 1, 0, 0)
```

Version 1.2.1

Shy7777:
Dates like `the 1st day of March, 2025` currently return `None`, even though the same date without `day` parses correctly. This adds a small English rule that treats `1st day of` like `1st of`, so the existing date parser can handle it.

The change updates the English language data and its generated file, with four regression cases covering `st`, `nd`, `rd`, and `th`.

The full suite passes on Python 3.12: 24,363 passed, 3 skipped, and 1 expected failure. Generated-data consistency, lint, and formatting checks pass too.

Fixes #1256.


AdrianAtZyte (approved):
> LGTM, thanks! Two optional nits: I would drop the README example, since the list there is for broad features and this is a fairly niche phrasing, and I think the new tests could be trimmed to three or four cases, since relative days are already covered elsewhere and the whitespace/case variants exercise the simplification machinery rather than this change.

Shy7777:
Thanks, that makes sense. I've removed the README example and trimmed the new tests to four cases, one for each ordinal suffix. The full test suite still passes.

PR #1384, opened 2026-09-10
Fix parser._settings leaking across unrelated items in search_dates

yamilmaud:
## Summary
Follow-up to #1363, extracting the standalone bug identified there.

`DateSearchWithDetection.search_parse` reuses a single `DateDataParser` instance across every substring found in the text, and across every candidate split explored when a substring fails to parse as a whole (`choose_best_split`). `parse_item` mutates `parser._settings` to add `RELATIVE_BASE` when chaining a date off a previously-parsed one, but that mutation persisted on the shared `parser` object for every subsequent, unrelated item and candidate split — nothing ever reset it.

This was mostly invisible because the split-recursion path started each candidate from an **empty** chain (`current_parsed = []`), so relative-base chaining for the first sub-item of any split silently depended on whatever `RELATIVE_BASE` happened to be left over from processing a completely unrelated prior item, rather than genuinely finding the relevant preceding date through `set_relative_base`.

Concretely: `search_dates("2014. July 13th July 14th")` only resolves "July 13th"/"July 14th" to year 2014 today because of this leak, not because of the intended chaining mechanism — `"July 13th July 14th"` fails to parse whole and falls into split-recursion, which starts from an empty chain and can't see "2014" at all through the real mechanism.

## Fix
- `parse_item` now always explicitly sets `parser._settings` from the pristine `settings` for every parse: with `RELATIVE_BASE` when chaining applies, or the base settings otherwise. No mutation carries over between calls.
- The split-recursion path in `parse_found_objects` is seeded with the outer chain of already-parsed dates (sliced back out before merging results), so it can find a genuinely relevant preceding date through `set_relative_base` instead of relying on the leak.

`test_splitting_of_not_parsed_1_en` and `_5_en` (existing, unrelated tests) now pass because of the intended chaining mechanism rather than by accident — no test expectations needed to change.

## Test plan
- [x] `pytest tests/` — 24224 passed, 6 skipped, 1 xfailed (no regressions, no expectation changes)
- [x] `ruff check` / `ruff format --check` clean
- [x] Manually verified the leak is gone with additional ad-hoc cases (multiple independent year/month-day sequences in one text)


AdrianAtZyte:
Would you say the following is true?

> In practice it is a refactor with no observable behavior change.

yamilmaud:
Not quite — there's a real behavior change here, not just a refactor.

The exact part that changes things is in `parse_item`. Before:

```python
if relative_base:
    parser._settings = parser._settings.replace(RELATIVE_BASE=relative_base)
    parsed_item = parser.get_date_data(item)
```

`RELATIVE_BASE` only ever got *set*, never *unset* — and `parser` is the same instance reused for every substring and every candidate split in the text. So once one item set `RELATIVE_BASE`, it just stayed on `parser._settings` for every later, unrelated parse, since there was no `else` branch to reset it back to the base settings.

On top of that, split-recursion (`parse_found_objects`) started each candidate split from `current_parsed = []` — an empty chain — so it had no way to look up the *real* preceding date via `set_relative_base` for that split. Any relative-base chaining it appeared to do was really just inheriting whatever `RELATIVE_BASE` leaked over from a previous, unrelated item.

Example: `search_dates("2014. July 13th July 14th")` resolves to 2014 today, but only by that leak — `"July 13th July 14th"` fails to parse as a whole and falls into split-recursion with an empty chain, so it never actually sees "2014" through the intended mechanism. Swap the order of unrelated date sequences in the text and that same leak could just as easily hand you the wrong year.

The fix makes `parser._settings` always explicit on every parse (reset to base settings when there's no relative base) and seeds split-recursion with the real outer chain instead of an empty one, so chaining works through `set_relative_base` for real.

So nothing *intended* breaks — but inputs that were only resolving correctly by luck now do so deterministically instead 🙂

AdrianAtZyte:
> Thanks, I confirmed the behavior change: on master, `March 3` in your test gets 2002, leaked from a discarded split candidate.
>
> A few things:
>
> - I think the seeding can be simpler: keep `current_parsed = []` and pass `parsed + current_parsed` to `parse_item`, no need for `seed_len` and slicing. Tests pass with that.
> - Please, trim the new comments; they describe the bug rather than the code. With the simpler seeding I think the second one can go entirely.
> - In the test, please, drop the “Before the fix” part of the docstring, and I wonder if asserting only `result[0]` would be better, since the rest pins unrelated odd results like `9 2001` → 2001-09-03.
> - `split_by_relative_expression` (#1391) still calls `parser.get_date_data(rest)` with whatever settings the last parse left behind. It only checks for `None`, so probably harmless, but it may be worth resetting there too.

yamilmaud:
Pushed 63f6d7ce with the simpler seeding, trimmed comments, and narrower test assertion.

On `split_by_relative_expression`: agreed it's harmless — `RELATIVE_BASE` only changes what a relative expression resolves to, never whether it parses, so it can't flip that `is None` check either way.

Is ok to rebase this branch and apply the same fix there, or would you rather handle it as a small follow-up once this merges and the branch is on current master? Thanks!

AdrianAtZyte:
> > Is ok to rebase this branch and apply the same fix there, or would you rather handle it as a small follow-up once this merges and the branch is on current master?
>
> Neither, I think. With this change, `split_by_relative_expression` only runs right after `parse_item` has set the settings for that same chunk, so there is nothing stale to reset. Also, please, do not rebase branches with an open PR; if you ever need master, merge it in.
>
> Two leftovers:
>
> - Please, drop the “rather than mutating leftover state from a prior call” part of the `parse_item` comment, or the comment altogether; it still describes the bug rather than the code.
> - Please, cut the test docstring down to its first sentence.
>
> Please, also update the PR description, it still mentions slicing the seeded chain back out.

